Issue #458, opened 2026-08-27
Authenticated requests 401 with "Failed to obtain authorization token" about ten minutes into a session


abhijithanandan:
Roughly ten minutes into a browser session on `echno.in`, every authenticated request starts failing with `401 Failed to obtain authorization token` and the UI fills with `Server Error` toasts. It is not confined to one module: chat, employee, user and leave-requests all fail together.

Found while doing the browser verification for #453. It is unrelated to that feature, but it made ordinary use of staging unusable partway through, so it is worth its own issue.

## Evidence

The 401 body comes from `RPTExchangeFilter`, not from the web BFF failing to read the session. The backend logs the underlying cause:

```
ERROR RPTExchangeFilter  Failed to exchange token for RPT on request to /api/v1/chat/rooms/web:
                         Failed to obtain RPT token for authorization
ERROR RPTExchangeFilter  Failed to exchange token for RPT on request to /api/v1/employee/web: ...
ERROR RPTExchangeFilter  Failed to exchange token for RPT on request to /api/v1/user/web/employees: ...
```

So the bearer token reached the backend; what failed is the backend exchanging it with Keycloak for an RPT.

The web side records the resulting 401s against every endpoint at once:

```
WARN Backend returned error {"status":401,"targetUrl":".../api/v1/chat/rooms/web/1", "body":"Failed to obtain authorization token"}
WARN Backend returned error {"status":401,"targetUrl":".../api/v1/leave-requests/web/pending-approvals/count", ...}
WARN Backend returned error {"status":401,"targetUrl":".../api/v1/user/web/employees", ...}
```

## Shape of it

- Logged in around 09:58 UTC. Everything worked normally for the first several minutes.
- Failures ran from **10:08:42Z to 10:12:30Z**, 137 of them.
- They stopped when I closed the browser tabs, that is when requests stopped, not because anything recovered. There have been none since, but nothing has exercised it since either.
- `echno-keycloak` reported healthy throughout and still does. The box was idle (load average 0.83 on a 72 core machine), so this is not resource pressure.
- A spontaneous full page reload happened in the same window (`performance.getEntriesByType('navigation')[0].type === 'reload'`), which is consistent with the middleware bouncing on a session error.

## What I have not established

[...]

Issue #261, opened 2026-03-06
Modifications Needed in Existing Relationships Inventory Related Modules


abinkjohny:
### Description

1. The `receivedBy` in `GoodsReceivedNoteDto` should be changed from `User` to `Employee`
2. The `createdBy` in `IntendDto` should be changed from `User` to `Employee`
3. The `createdBy` in `PurchaseOrderDto` should be changed from `User` to `Employee`
4. The `createdBy` in `PayableDto` should be changed from `User` to `Employee`
5. The `createdBy` in `InventoryTransactionDto` should be changed from `User` to `Employee`
6. The `createdBy` in `MaterialConsumptionDto` should be changed from `User` to `Employee`

### Motivation

Change in the business logic

PR #460, opened 2026-08-27
Classify RPT exchange failures so an expired token, a down Keycloak and a misconfig are distinguishable

abhijithanandan:
Closes the diagnosability half of #458. The cure for the underlying cause, the BFF forwarding an expired access token instead of refreshing it, ships on the web side; this PR is the second half the issue asks for: make the failure mode diagnosable and classify it correctly.

## Why

`RPTExchangeFilter` logged the same generic sentence for every failure mode. An expired token, an unreachable Keycloak and a bad client secret all produced `Failed to exchange token for RPT on request to ...: Failed to obtain RPT token for authorization`, and all three answered the caller with 401. Keycloak's own status and error body never reached the log, which is why the recent app-wide 401 took longer to pin down than it should have.

## What changed

`KeycloakAuthorizationService.exchangeForRPT` now classifies the failure and throws `RPTExchangeException`, a typed exception in the same package carrying a `Reason`, Keycloak's HTTP status and its response body:

| Reason | Trigger | Log level | Response |
|---|---|---|---|
| `TOKEN_INVALID` | OAuth body of `invalid_token` / `invalid_grant`, or a 401 with no readable error code | WARN | 401, `{"error":"invalid_token"}` + `WWW-Authenticate` |
| `PERMISSION_DENIED` | `access_denied`: the token is fine but grants nothing | WARN | 401, `{"error":"access_denied"}` |
| `EXCHANGE_MISCONFIGURED` | any other explicit error code (`invalid_client`, unknown audience, authorization services off), or any other 4xx | ERROR | 401, `{"error":"unauthorized"}` |
| `KEYCLOAK_UNAVAILABLE` | `ResourceAccessException` (refused, DNS, connect/read timeout) or a 5xx from Keycloak | ERROR | **503**, `{"error":"service_unavailable"}` + `Retry-After` |
| `PROTOCOL_VIOLATION` | 2xx with no `access_token` | ERROR | 401, `{"error":"unauthorized"}` |

Three things worth calling out:

- **Log level follows ownership.** A caller turning up with a stale token is routine, so it logs at WARN, not ERROR. That is the noise the issue flagged: 137 ERROR lines in four minutes for one user's expired token. Config and dependency failures are genuine ERRORs and now log Keycloak's status and (bounded, truncated at 512 chars) response body so a single line explains the failure.
- **503 for a dependency outage.** A client that answers a 401 by re-authenticating against a Keycloak that is down would be doing exactly the wrong thing, so an unreachable or erroring Keycloak is no longer dressed up as an authentication failure.
- **The caller can now distinguish the cases.** The body is terse and OAuth-shaped, so the web side can branch on `invalid_token` (refresh and retry) versus `access_denied` (do not) versus `unauthorized` (opaque). Misconfiguration detail never reaches the caller; it stays a 401 and lives only in the log.

Nothing is logged that carries bearer material: not the access token, not the RPT, not the client secret. Only Keycloak's OAuth error payload, which carries none.

Unchanged: `shouldNotFilter` on `/actuator`, the no-bearer passthrough, the request wrapper that swaps in the RPT, the `RPTCache` contract (a failed mint is still never cached), and the success path on the wire.

## Tests

Plain JUnit + Mockito, no Spring context anywhere, per the test-JVM heap ceiling.

- New `KeycloakAuthorizationServiceTest` (15 tests): a mocked `RestTemplate` drives every classification branch, including a non-JSON error body from a proxy, a blank `access_token`, and an assertion that the access token never lands on the exception.
- `RPTExchangeFilterTest` extended to 14 tests: each reason mapped to its status, body and headers, plus the misconfiguration case asserting Keycloak's payload does not leak to the caller, the chain never being reached on failure, the no-bearer passthrough and `shouldNotFilter`.

```
$ ./gradlew test --tests '*RPTExchangeFilterTest' --tests '*KeycloakAuthorizationServiceTest'
BUILD SUCCESSFUL in 1m
RPTExchangeFilterTest            tests="15" skipped="0" failures="0" errors="0"
KeycloakAuthorizationServiceTest tests="15" skipped="0" failures="0" errors="0"
```

## Division of labour with the web fix

The backend now answers an expired token with a 401 and `{"error":"invalid_token"}`. The web fix on `echno-web` PR #261 keys its automatic session refresh off the BFF's own `SessionTokenExpired` signal, which fires before the request ever reaches the backend. The two therefore do not overlap: the web side recovers the session upstream, and a caller that somehow arrives here with a lapsed token gets a clean, terse 401 rather than an automatic recovery. That is the right split, and the web client should not be wired to this error vocabulary as a refresh trigger.

## Review follow-up

`rejectClassified` was a switch *statement* with arrow labels and no `default`, so the compiler did not enforce exhaustiveness. A sixth `Reason` would have matched nothing, written no status and no body, and let the request fall out of the filter as an empty 200: an authorization failure answered as success, silently. It is now a switch expression yielding a `Rejection` (status, error code, description), which the compiler does check. Verified by temporarily adding a constant:

```
RPTExchangeFilter.java:95: error: the switch expression does not cover all possible input values
        Rejection rejection = switch (e.getReason()) {
                              ^
1 error
```

`Retry-After` moved into `writeError` next to `WWW-Authenticate`, keyed on the status, so each arm is a single `yield`. A new test iterates `Reason.values()` and asserts every constant produces a 4xx or 5xx and a non-empty machine-readable body, so the gap fails the suite as well as the build.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authorization-token exchange error handling with clearer, machine-readable OAuth responses.
  * Invalid tokens and denied permissions now return appropriate 401 responses.
  * Service outages now return a 503 response with retry guidance.
  * Added standardized authentication headers and JSON error messages.
  * Prevented sensitive access-token details from appearing in error messages or logs.

* **Tests**
  * Added coverage for successful exchanges, failure scenarios, network errors, malformed responses, and request filtering behavior.

## Second review round (CodeRabbit)

**`invalid_client` on a 401 was classified as the caller's problem.** Keycloak answers 401 with `error=invalid_client` when *our* client credentials are wrong. Checking the status before the error code reported a bad client secret as `TOKEN_INVALID`: logged below ERROR where nobody looks, and handed back as `invalid_token`, inviting the client to re-authenticate in a loop against a server whose configuration is broken. That is exactly the confusion the issue describes, so the explicit OAuth error code now decides and the status is only the fallback when the body carries no readable code.

**The body was truncated before it was parsed.** A payload with a long field ahead of `error` parsed as malformed JSON, lost the code, and fell back to the status. Classification now reads `getResponseBodyAsString()` whole and truncates only what is logged and carried on the exception.

**`PERMISSION_DENIED` is now 403.** The token authenticated and only the authorization decision went against the caller, so `WWW-Authenticate` comes off that response as well.

Both classifier fixes have regression tests that were confirmed to fail against the previous ordering:

```
KeycloakAuthorizationServiceTest > invalidClientOn401_classifiesAsMisconfigurationNotAnExpiredToken() FAILED
KeycloakAuthorizationServiceTest > errorCodeBeyondTheLogTruncationLimit_isStillClassified() FAILED
```

### The 401 check before moving to 403

Grepped `echno-web` (`development`, plus PR #261's diff) for anything keying on 401 on this path. Nothing branches on a backend 401 in a way a 403 would bypass:

- Forced sign-out in `components/providers/auth-provider.tsx` fires on the NextAuth session error (`RefreshAccessTokenError`, `SessionRevoked`) and on `sessionExpiresAt`, never on a response status. `proxy.ts` redirects on the same session errors.
- `ApiError.isAuthError` is `status === 401 || status === 403`, so both are treated alike, and `lib/query/retry.ts` suppresses retries for both.
- The 401 cases in `lib/api/api-client.ts` and `lib/utils/api-utils.ts` are message lookup tables; 403 already has the better copy.
- The BFF's own 401s (`Session revoked`, and PR #261's `SessionTokenExpired`) are minted by the BFF before the request reaches the backend. PR #261's `isSessionExpiredResponse` requires both a 401 and that specific body code, and it has a test asserting a 401 without the signal is not recovered from.

### Test output after the fixes

```
$ ./gradlew test --tests '*RPTExchangeFilterTest' --tests '*KeycloakAuthorizationServiceTest' --rerun
32 tests PASSED
RPTExchangeFilterTest            tests="15" skipped="0" failures="0" errors="0"
KeycloakAuthorizationServiceTest tests="17" skipped="0" failures="0" errors="0"
```

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/configuration/KeycloakAuthorizationService.java
+++ b/src/main/java/org/tornotron/echno_backend/common/configuration/KeycloakAuthorizationService.java
@@ -130,18 +130,31 @@ public class KeycloakAuthorizationService {
     }
 
     /**
-     * Splits a 4xx from Keycloak into the caller's problem and ours. A 401, or any OAuth error body of
-     * {@code invalid_token} or {@code invalid_grant}, means the caller arrived with a token Keycloak
-     * will not accept. {@code access_denied} means the token was fine but carries no permission.
-     * Everything else (unknown client, unknown audience, bad client credentials, authorization services
-     * switched off) is a misconfiguration on this side.
+     * Splits a 4xx from Keycloak into the caller's problem and ours.
+     *
+     * <p>The OAuth error code in the body is checked first and the HTTP status is only a fallback,
+     * because the two disagree in a way that matters. Keycloak answers 401 with {@code invalid_client}
+     * when <em>our</em> client credentials are wrong, so trusting the status would report a bad client
+     * secret as the caller's expired token: logged below ERROR where nobody looks, and handed back as
+     * {@code invalid_token}, which invites the client to re-authenticate in a loop against a server
+     * whose configuration is broken. An unambiguous error code always beats a guess from the status.
+     *
+     * <p>So: {@code invalid_token} or {@code invalid_grant} means the caller's token was rejected,
+     * {@code access_denied} means the token was fine but carries no permission, and any other explicit
+     * code (unknown client, unknown audience, bad client credentials, authorization services switched
+     * off) is a misconfiguration on this side. Only when the body carries no readable code does the
+     * status decide, and there a 401 is the caller's token.
      */
     private RPTExchangeException classifyClientError(String tokenEndpoint, HttpClientErrorException e) {
         int status = e.getStatusCode().value();
-        String responseBody = truncate(e.getResponseBodyAsString());
-        String oauthError = oauthErrorCode(responseBody);
+        // Classify on the whole body and truncate only what is logged and carried: truncating first
+        // can cut a long payload mid-JSON, so the error code would be unreadable and misclassified.
+        String fullBody = e.getResponseBodyAsString();
+        String oauthError = oauthErrorCode(fullBody);
+        String responseBody = truncate(fullBody);
 
-        if (status == 401 || (oauthError != null && TOKEN_ERRORS.contains(oauthError))) {
+        boolean tokenRejected = oauthError == null ? status == 401 : TOKEN_ERRORS.contains(oauthError);
+        if (tokenRejected) {
             // Routine: a caller turned up with an expired or revoked token. Logged below ERROR on purpose,
             // one stale token can otherwise produce hundreds of lines a minute.
             log.warn("RPT exchange at {} rejected the caller's access token: HTTP {}, error={}. Body: {}",
@@ -163,7 +176,11 @@ public class KeycloakAuthorizationService {
                 "Keycloak rejected the RPT exchange configuration", e);
     }
 
-    /** Reads the OAuth {@code error} code out of an error payload, or returns {@code null} if there is none. */
+    /**
+     * Reads the OAuth {@code error} code out of an error payload, or returns {@code null} if there is
+     * none. Must be given the untruncated body, or a long payload parses as malformed JSON and the
+     * code is lost.
+     */
     @Nullable
     private static String oauthErrorCode(@Nullable String responseBody) {
         if (responseBody == null || responseBody.isBlank()) {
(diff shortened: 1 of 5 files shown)


abhijithanandan:
@coderabbitai review

Issue #260, opened 2026-03-06
Changes Needed In PurchaseOrder, Intent & GRN


abinkjohny:
### Description

1. One `Intent` can have multiple related `PurchaseOrder` as the materials can be purchased from multiple `Vendors`
2. One `PurchaseOrder` can have multiple `GRN` as the materials from a single purchase can be stored in different `Locations`.

### Motivation

Micro Management

PR #462, opened 2026-08-27
fix(inventory): complete the movement-history entry and fetch-join its associations

abhijithanandan:
## What

Two fixes to the material movement-history endpoint added in #425, both surfaced while
pointing echno-web's material timeline at it (tornotron/echno-web#253).

### 1. The entry was missing the running balance and the creator

`MaterialMovementHistoryDto` carried the location, project, type, direction, quantity,
timestamp and reference, but not `openingStock`, `closingStock` or who booked the movement,
although the `InventoryTransaction` row it is built from has all three. A timeline built on
the endpoint therefore had to drop the per-row `Balance: opening -> closing - by <name>`
line that the full `InventoryTransactionDto` supports, which is most of what a stock
timeline gets read for.

Adds `openingStock`, `closingStock` and `createdByName`, with `@Schema` annotations in the
same style as the existing fields. The creator flattens to a display name through a MapStruct
source path (`createdBy.employeeName`), the same way `storageLocationName` and `projectName`
flatten, rather than embedding an employee object in a timeline payload. `createdBy` is
nullable, and the generated mapper null-checks the path, so an uncredited row yields a null
name rather than an NPE.

### 2. The finder fetched nothing, so the endpoint was already N+1

```java
@Query("SELECT it FROM InventoryTransaction it WHERE it.material.id = :materialId " +
       "ORDER BY it.transactionDate ASC, it.id ASC")
```

`project`, `storageLocation` and `createdBy` are all `@ManyToOne(fetch = LAZY)` on the entity,
and the mapper reads a field off each. At the default page size of 10 that was 1 page query +
1 count + 20 association selects; adding the creator name naively would have made it 30.

All three are now `LEFT JOIN FETCH`-ed. They are to-one associations, so the join neither
multiplies rows nor pushes Hibernate into its in-memory pagination fallback (`HHH000104`);
no collection is fetched here, which is what would break paging. `LEFT` rather than inner
because `storageLocation` and `createdBy` are nullable. The count query is spelled out,
because Spring Data cannot derive one from a query carrying fetch joins.

## Generated SQL (checked, not assumed)

Captured by re-running the test with `hibernate.show_sql` + `format_sql` on the lab build box.
The page is now one statement, with the pagination pushed into SQL:

```sql
select
    it1_0.id, it1_0.closing_stock, it1_0.created_at,
    cb1_0.id, cb1_0.employee_name, ...,
    it1_0.material_id, it1_0.opening_stock, it1_0.organization_id,
    it1_0.project_id, p1_0.id, ...,
    sl1_0.id, ...,
    it1_0.transaction_date, it1_0.transaction_type, it1_0.unit_cost
from
    inventory_transaction it1_0
left join
    storage_location sl1_0
        on sl1_0.id=it1_0.storage_location_id
left join
    project p1_0
        on p1_0.id=it1_0.project_id
left join
    employee cb1_0
        on cb1_0.id=it1_0.created_by_id
where
    it1_0.material_id=?
order by
    it1_0.transaction_date,
    it1_0.id
fetch
    first ? rows only
```

plus the separate `select count(it1_0.id) from inventory_transaction it1_0 where
it1_0.material_id=?`. Note `fetch first ? rows only`: the limit is in the database, so
Hibernate is not paginating in memory. No `HHH000104` warning is emitted.

Two residual selects are visible and neither scales with rows on the page: `material` is an
eager `@ManyToOne` on the entity and loads once for the material being asked about, and
`Employee.orgRoles` is an eager `@ElementCollection` that Hibernate batches into a single
`where or1_0.employee_id = any (?)` for every creator on the page rather than one per
employee.

## Tests

`MaterialMovementHistoryIT` (repository slice against the real CockroachDB Testcontainer)
grows a fixture of four cement movements with a chained balance, alternating over two storage
locations and two creators, plus a steel movement that must not leak in:

- `movementHistory_returnsMaterialRowsOldestFirstWithLocationAndDirection` - the original
  assertions, widened to the new fixture.
- `movementHistory_dtoCarriesBalanceProjectAndCreator` - the mapped DTO carries the chained
  opening/closing figures, the project and location names and the alternating creator names.
- `movementHistory_leavesTheCreatorNameNullWhenTheRowHasNoCreator` - an uncredited row maps
  to a null name.
- `movementHistory_doesNotQueryPerRow` - the N+1 pin. Asserts every association is already
  initialised on each row after the page query (`Hibernate.isInitialized`), and that a
  four-row page issues **exactly** as many statements as a two-row page covering the same
  distinct locations and creators. Query counting runs off Hibernate's `Statistics`, toggled
  on programmatically and restored in a `finally`, so it needs no extra `@TestPropertySource`
  and therefore no second Spring context, which keeps it inside the CI heap cap.

```
MaterialMovementHistoryIT > movementHistory_returnsMaterialRowsOldestFirstWithLocationAndDirection() PASSED
MaterialMovementHistoryIT > movementHistory_doesNotQueryPerRow() PASSED
MaterialMovementHistoryIT > movementHistory_dtoCarriesBalanceProjectAndCreator() PASSED
MaterialMovementHistoryIT > movementHistory_leavesTheCreatorNameNullWhenTheRowHasNoCreator() PASSED

BUILD SUCCESSFUL in 4m 33s
```

No schema change, so no Liquibase changelog: the three fields already exist as columns on
`inventory_transaction` and `employee`.

## Downstream

Head of a three-PR chain. This must merge **and deploy** before the two below, since both
read the new fields:

1. this PR
2. tornotron/echno-core#39 - parses the three fields, then releases `@tornotron/echno-core` 1.14.0
3. tornotron/echno-web#260 - restores the `Balance` row on the repointed timeline

Refs tornotron/echno-web#253, #425

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/inventoryTransaction/InventoryTransactionController.java
+++ b/src/main/java/org/tornotron/echno_backend/inventoryTransaction/InventoryTransactionController.java
@@ -118,8 +118,9 @@ public class InventoryTransactionController {
             summary = "Get a material's movement history (timeline, paginated)",
             description = "Returns the material's stock movements as a timeline, oldest movement first, one "
                     + "page at a time. Each entry carries the storage location, project, movement type and its "
-                    + "stock direction, the quantity changed, the timestamp and the source reference, so a "
-                    + "caller can show where the material has been, when, and how much."
+                    + "stock direction, the quantity changed, the stock level either side of it, the timestamp, "
+                    + "the source reference and the employee who booked it, so a caller can show where the "
+                    + "material has been, when, how much, and on whose hand."
     )
     @ApiResponses({
             @ApiResponse(responseCode = "200", description = "Page of movement-history entries for the material"),
--- a/src/main/java/org/tornotron/echno_backend/inventoryTransaction/InventoryTransactionControllerWeb.java
+++ b/src/main/java/org/tornotron/echno_backend/inventoryTransaction/InventoryTransactionControllerWeb.java
@@ -108,8 +108,9 @@ public class InventoryTransactionControllerWeb {
             summary = "Get a material's movement history (timeline, paged)",
             description = "Returns the material's stock movements as a timeline, oldest movement first, "
                     + "one page at a time. Each entry carries the storage location, project, movement type "
-                    + "and its stock direction, the quantity changed, the timestamp and the source reference "
-                    + "so the caller can show where the material has been, when, and how much."
+                    + "and its stock direction, the quantity changed, the stock level either side of it, the "
+                    + "timestamp, the source reference and the employee who booked it, so the caller can show "
+                    + "where the material has been, when, how much, and on whose hand."
     )
     @ApiResponses({
             @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "200", description = "Page of movement-history entries returned"),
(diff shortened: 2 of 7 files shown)


abhijithanandan:
@coderabbitai review

PR #463, opened 2026-08-27
Retry CockroachDB serializable aborts instead of returning 500

### LINKED ISSUE #459: CockroachDB serializable retry aborts surface as 500s instead of being retried
CockroachDB runs `SERIALIZABLE` by default and expects the client to retry a transaction it aborts with `RETRY_SERIALIZABLE` (SQLSTATE `40001`). We do not retry. Spring's `@Transactional` has no retry of its own, so the abort propagates out of the commit, `GlobalExceptionHandler` logs it as "Unknown error occurred" and the user gets a 500.

Seen live on `.116` staging in the last 24 hours, on light single-user traffic:

```
2026-08-27T10:03:06Z  ChatService.markRead      <- ChatControllerWeb.markRoomRead:116
2026-08-27T10:05:47Z  ChatService.sendMessage   <- ChatControllerWeb.sendMessage:155
2026-08-27T10:20:24Z  ChatService.sendMessage   <- ChatControllerWeb.sendMessage:155
```

with:

```
org.springframework.dao.CannotAcquireLockException: Hibernate transaction: Unable to commit against
JDBC Connection; ERROR: restart transaction: TransactionRetryWithProtoRefreshError:
TransactionRetryError: retry txn (RETRY_SERIALIZABLE) ... iso=Serializable ... lock=true stat=PENDING
```

Three aborts in a day from one person clicking around is a bad sign. These are contended-row writes: `markRead` and `sendMessage` both touch the same room row, so the contention scales with participants in a room, not with users on the platform. Real chat traffic will produce this steadily, and every occurrence is a message the user believes failed.

The user-visible failure is worse than the abort itself. A retryable abort is a normal, expected outcome under `SERIALIZABLE`. Presenting it as `500 An unexpected error occurred` sends the user to retry by hand, and on `sendMessage` they cannot tell whether the message was written or not.

## What to do

The fix belongs in one place rather than in each service. Retry the transaction on `40001` with bounded attempts and a little jitter, applied around transactional service methods, so a retryable abort becomes a slightly slower success instead of a 500. The transaction has to be restarted from the beginning, so the retry has to sit outside the transaction boundary, not inside it.

Worth doing alongside:

- keep the retryable case out of the "unknown error" bucket in `GlobalExceptionHandler`, so an abort that genuinely exhausts its retries returns something honest (`409` or `503` with a retry hint) and is greppable;
- narrow the chat transactions themselves. `markRead` in particular looks like it holds more of the room row than it needs to;
- once retries are in, count them, so contention is visible before it becomes a support ticket.

Not chat-specific. Chat is simply the first module with enough write contention to surface it, and the same abort can hit any hot row (stock, leave, invoice numbering).

abhijithanandan:
Closes #459.

CockroachDB runs `SERIALIZABLE` and resolves a read-write conflict by aborting one side with SQLSTATE `40001`, expecting the client to run the transaction again. We did not, so the abort escaped the commit as `CannotAcquireLockException`, landed in the "Unknown error occurred" branch of `GlobalExceptionHandler`, and reached the user as a 500. On `sendMessage` the user could not even tell whether their message had been written.

## The mechanism, and how the retry is kept outside the transaction

`TransactionRetryTemplate` runs a unit of work in a transaction and starts it over on a `40001` abort. Bounded attempts (4 by default), exponential backoff doubling from 25 ms with full jitter capped at 250 ms, all three tunable through `echno.transaction.retry.*`.

The correctness constraint is that a retry has to restart the transaction from the beginning, so the loop must sit outside the transaction boundary. Two findings shaped how that is guaranteed here.

**An `@Around` aspect cannot do it in this application.** `EchnoBackendApplication` carries `@EnableTransactionManagement(order = Ordered.HIGHEST_PRECEDENCE)`, which pins the transaction advisor at `Integer.MIN_VALUE`. No advice can be ordered ahead of that, so an `@Order`-based retry aspect would have run *inside* the transaction and quietly done nothing, which is the worst possible outcome for this bug. Rather than move a global setting that every other advisor's position depends on, the boundary here is a plain method call: `TransactionalWorkRunner.runInTransaction` opens and commits the transaction entirely within one invocation, and the retry loop wraps that invocation. Each attempt is therefore a new transaction by construction, independent of any advisor ordering.

**The runner has to be declaratively `@Transactional`, not a `TransactionTemplate`.** `HibernateFilterConfig` advises `@Transactional` methods in this package tree and is what enables the `orgFilter` tenant filter on the session. A transaction opened programmatically carries no such annotation, so it would have run with tenant filtering off. Keeping the boundary declarative keeps the filter behaviour byte-for-byte what it is today.

Retryability is decided by reading `getSQLState()` off the `SQLException` and comparing it to `40001`, never by matching the server's message. The detector walks the cause chain (Spring's `DataAccessException` over Hibernate's `JDBCException` over the driver's `SQLException`) and also `SQLException.getNextException()`, which the PostgreSQL driver uses for batch failures. Both walks are depth-capped so a self-referencing chain terminates.

If the work is invoked while a caller already owns a transaction, the template declines to retry and runs once: there is no boundary here to restart, and the caller's transaction is already doomed.

## Where it is applied, and why each is safe to re-run

| Method | Applied | Reason |
|---|---|---|
| `ChatService.markRead` | yes | Reads the room row that every `sendMessage` bumps, which is exactly the read-write pair that aborts. The work is one participant timestamp plus an event released after commit, so a rolled-back attempt leaves nothing behind. |
| `ChatService.sendMessage(roomId, content, replyToId)` | yes | The JSON path from `ChatControllerWeb:155`, the one in the evidence. Bumps the room's `updatedAt`, so two people talking in one room write the same row. Database writes and an `AFTER_COMMIT` event only. |
| `ChatService.sendMessage(..., List<MultipartFile>)` | **no** | Uploads attachments to object storage inside the transaction. A retry would repeat those uploads, from multipart streams that have already been consumed. Left on the existing `@Transactional`. |

`ChatEventListener` is `@TransactionalEventListener(AFTER_COMMIT)`, so an aborted attempt notifies nobody and only the committed attempt fans out. Neither method is called from inside another transaction; both entry points are the controller.

Applying it elsewhere (stock, leave, invoice numbering) is now one injected bean and one lambda, with the same "is this safe to re-run" question answered per method.

## Status mapping

`TransactionRetriesExhaustedException` maps to **409 Conflict** with `retryable: true` and `retryAfterSeconds: 1`, as an RFC 7807 problem alongside the existing handlers.

409 rather than 503 because the service is healthy and every other request is being served normally. 503 asserts unavailability, and clients, proxies and load balancers act on it by backing off from the whole API or failing it out, which would turn one contended row into a much larger outage. 409 is also already this codebase's answer to "someone else got there first" (`DataIntegrityViolationException`, `DuplicateResourceException`, `PeriodClosedException`), and the `retryable` flag tells the client it may simply send the request again.

`CannotAcquireLockException` maps to the same problem, so a `40001` that escapes from code not yet using the template is honest and greppable rather than a 500. That handler is narrow on purpose: it does not cover `ConcurrencyFailureException` generally, so optimistic-locking behaviour is unchanged.

## Observability

Two Micrometer counters, tagged with a stable low-cardinality `operation` label:

- `echno_transaction_retry_attempts_total` — one per restarted attempt
- `echno_transaction_retry_exhausted_total` — one per unit of work that gave up

Contention now shows on the dashboards before it shows in a support ticket.

## Tests

`TransactionRetryTemplateTest` (plain JUnit, no application context) drives the **real** `TransactionalWorkRunner` through a **real** Spring `TransactionInterceptor` over a transaction manager that counts what it is asked to do. The proxy is assembled by hand with `ProxyFactory`, which is what makes the central assertion possible: after one abort the counters read 2 transactions begun, 1 rolled back, 1 committed, rather than two attempts sharing one doomed transaction. Also covers retries exhausted producing the mapped exception with its cause, a non-retryable `DataAccessException` not being retried, the no-result form, and the nested-transaction guard.

`SerializationFailureDetectorTest` covers the SQLSTATE walk: the real `CannotAcquireLockException` → `LockAcquisitionException` → `SQLException` chain from the incident, the driver's next-exception chain, `23505` correctly rejected, no `SQLException` at all, and a self-referencing chain.

`TransactionRetryTemplateIT` proves it against a real CockroachDB v26.2.4, not a stand-in. It raises a genuine abort with `crdb_internal.force_retry` (behind `SET LOCAL allow_unsafe_internals`, which v26 now requires), so no second thread and no timing window are involved, and asserts the transaction restarted, the row committed for real, and the retry counter moved. The server log line it produces is `ERROR: restart transaction: TransactionRetryWithProtoRefreshError: forced by crdb_internal.force_retry()`, which confirms the whole driver → Hibernate → Spring chain really does carry `40001` to the detector.

On the context-cache budget: the only Spring context added is one `@DataJpaTest` slice, the lightest thing that can hold a real transaction boundary, and it carries `@DirtiesContext(AFTER_CLASS)` so it does not sit in the cache for the rest of the run. The behavioural ordering proof needs no context at all.

## Verification

Full `./gradlew check` on the lab build box (Testcontainers CockroachDB v26.2.4):

```
BUILD SUCCESSFUL in 6m 49s
tests 373 failures 0 errors 0 skipped 0
line coverage 34.95%   (JaCoCo floor 0.25)
```

## Deliberately not done

Narrowing `markRead`'s transaction. The only read it could shed is the room-existence check, and `requireParticipant` would then answer an unknown room with 403 instead of the documented 404. That is a public API behaviour change and belongs in its own commit, not smuggled into a reliability fix. The retry covers the contention in the meantime.

## Summary by CodeRabbit

- **New Features**
  - Added automatic recovery for transient database conflicts while marking chats as read or sending text messages.
  - Added configurable retry attempts and backoff settings.
  - Added clear conflict responses when updates cannot be completed after retries.

- **Bug Fixes**
  - Reduced failures caused by concurrent chat updates by retrying eligible transactions safely.

- **Tests**
  - Added unit and integration coverage for retry behavior, conflict detection, limits, and successful recovery.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/chat/ChatService.java
+++ b/src/main/java/org/tornotron/echno_backend/chat/ChatService.java
@@ -20,6 +20,7 @@ import org.tornotron.echno_backend.common.entity.AttachmentDto;
 import org.tornotron.echno_backend.common.exception.ResourceNotFoundException;
 import org.tornotron.echno_backend.common.multitenancy.TenantContext;
 import org.tornotron.echno_backend.common.multitenancy.TenantEntityHelper;
+import org.tornotron.echno_backend.common.retry.TransactionRetryTemplate;
 import org.tornotron.echno_backend.common.service.AttachmentService;
 import org.tornotron.echno_backend.employee.Employee;
 import org.tornotron.echno_backend.employee.EmployeeRepository;
@@ -72,6 +73,7 @@ public class ChatService {
     private final EmployeeRepository employeeRepository;
     private final AttachmentService attachmentService;
     private final ApplicationEventPublisher eventPublisher;
+    private final TransactionRetryTemplate transactionRetryTemplate;
 
     public ChatService(ChatRoomRepository roomRepository,
                        ChatParticipantRepository participantRepository,
@@ -82,7 +84,8 @@ public class ChatService {
                        UserContextService userContextService,
                        EmployeeRepository employeeRepository,
                        AttachmentService attachmentService,
-                       ApplicationEventPublisher eventPublisher) {
+                       ApplicationEventPublisher eventPublisher,
+                       TransactionRetryTemplate transactionRetryTemplate) {
         this.roomRepository = roomRepository;
         this.participantRepository = participantRepository;
         this.messageRepository = messageRepository;
@@ -93,6 +96,7 @@ public class ChatService {
         this.employeeRepository = employeeRepository;
         this.attachmentService = attachmentService;
         this.eventPublisher = eventPublisher;
+        this.transactionRetryTemplate = transactionRetryTemplate;
     }
 
     @Transactional(readOnly = true)
@@ -153,18 +157,27 @@ public class ChatService {
         return toRoomDto(saved, employeeId);
     }
 
-    /** Marks the room read for the caller by advancing their {@code lastReadAt} to now. */
-    @Transactional
+    /**
+     * Marks the room read for the caller by advancing their {@code lastReadAt} to now.
+     *
+     * <p>Run through the retry template rather than under a plain {@code @Transactional}: this
+     * reads the room row that every {@link #sendMessage} concurrently bumps, which is exactly
+     * the read-write pair CockroachDB resolves by aborting one side under {@code SERIALIZABLE}.
+     * The work is a single participant timestamp plus an after-commit event, so running it
+     * again from the start costs nothing and produces the same result.
+     */
     public void markRead(Long roomId) {
-        Long employeeId = resolveCurrentEmployeeId();
-        requireRoom(roomId);
-        ChatParticipant participant = requireParticipant(roomId, employeeId);
-        participant.setLastReadAt(LocalDateTime.now());
-        participantRepository.save(participant);
+        transactionRetryTemplate.executeWithoutResult("ChatService.markRead", () -> {
+            Long employeeId = resolveCurrentEmployeeId();
+            requireRoom(roomId);
+            ChatParticipant participant = requireParticipant(roomId, employeeId);
+            participant.setLastReadAt(LocalDateTime.now());
+            participantRepository.save(participant);
 
-        // Only the reader. An unread count is per viewer, so nobody else's room list changes
-        // when someone catches up on their own.
-        raise(ChatEventType.ROOM_UPDATED, roomId, null, employeeId, List.of(employeeId));
+            // Only the reader. An unread count is per viewer, so nobody else's room list changes
+            // when someone catches up on their own.
+            raise(ChatEventType.ROOM_UPDATED, roomId, null, employeeId, List.of(employeeId));
+        });
     }
 
     @Transactional(readOnly = true)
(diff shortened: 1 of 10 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #465, opened 2026-08-27
Review only, do not merge: replay of #463 (d3d8f73) for a deferred bot review

abhijithanandan:
## Review-only. Do not merge. This PR will be closed.

This carries no new change. It replays the diff that already merged into `development`
so an automated review can run over it, and it will be closed and its branch deleted
once that review has been read.

- Original PR: #463
- Merged commit: `d3d8f73` "Retry serializable aborts instead of returning 500"
- Branch base: `7c688ed`, the first parent of that commit, so the diff here is byte for
  byte what merged.

The original merged inside a window where the review bot's hourly allowance was already
spent. Its status check reported a pass with the summary "Review rate limited", so nothing
was actually reviewed. This recovers that review after the fact.

### What the change does

CockroachDB runs SERIALIZABLE and resolves a read/write conflict by aborting one side with
SQLSTATE 40001, expecting the client to run the transaction again. The abort escaped commit
as `CannotAcquireLockException`, fell into the unknown-error bucket of `GlobalExceptionHandler`
and reached the user as a 500. The change adds a bounded retry with fully jittered exponential
backoff, applies it to the two contended chat writes, and maps an exhausted conflict to 409.

### Two design decisions worth a reviewer's attention

**The retry sits outside the transaction boundary, and it has to.** An `@Around` aspect cannot
achieve that here: `EchnoBackendApplication` carries
`@EnableTransactionManagement(order = Ordered.HIGHEST_PRECEDENCE)`, which pins the transaction
advisor at `Integer.MIN_VALUE`, and no advice can be ordered ahead of that. An annotation-driven
retry would therefore have run inside the transaction and quietly done nothing. The boundary is
a method call instead: `TransactionalWorkRunner.runInTransaction` is invoked through the proxy,
and the retry loop wraps that call, so every attempt is a fresh transaction whatever the advisor
ordering turns out to be.

**`TransactionalWorkRunner` is declaratively `@Transactional` rather than a `TransactionTemplate`.**
`HibernateFilterConfig` advises `@Transactional` methods to enable the `orgFilter` tenant filter.
A programmatic transaction boundary would have run with tenant filtering off, which is a data
leak rather than a style difference.

### Other points to weigh

- The 409 versus 503 choice for a conflict that exhausts its retries.
- Which chat methods are safe to retry. The multipart `sendMessage` overload is deliberately
  excluded: it uploads attachments to object storage inside the transaction, and a retry would
  repeat those uploads from streams that have already been consumed.
- Retryability is decided by reading `getSQLState()` off the `SQLException`, walking both the
  cause chain and the driver's next-exception chain, rather than by matching server wording.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when sending messages and marking messages as read during temporary database transaction conflicts.
  * Automatically retries eligible operations before reporting failure.
  * Returns a clear conflict response with retry guidance when attempts are exhausted.

* **Configuration**
  * Added settings to control retry attempts and backoff timing.

* **Monitoring**
  * Added metrics for transaction retries and exhausted retry attempts.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/chat/ChatService.java
+++ b/src/main/java/org/tornotron/echno_backend/chat/ChatService.java
@@ -20,6 +20,7 @@ import org.tornotron.echno_backend.common.entity.AttachmentDto;
 import org.tornotron.echno_backend.common.exception.ResourceNotFoundException;
 import org.tornotron.echno_backend.common.multitenancy.TenantContext;
 import org.tornotron.echno_backend.common.multitenancy.TenantEntityHelper;
+import org.tornotron.echno_backend.common.retry.TransactionRetryTemplate;
 import org.tornotron.echno_backend.common.service.AttachmentService;
 import org.tornotron.echno_backend.employee.Employee;
 import org.tornotron.echno_backend.employee.EmployeeRepository;
@@ -72,6 +73,7 @@ public class ChatService {
     private final EmployeeRepository employeeRepository;
     private final AttachmentService attachmentService;
     private final ApplicationEventPublisher eventPublisher;
+    private final TransactionRetryTemplate transactionRetryTemplate;
 
     public ChatService(ChatRoomRepository roomRepository,
                        ChatParticipantRepository participantRepository,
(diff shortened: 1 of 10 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Review recovered and read. One finding, tracked as #468. Closing as planned; this PR was never intended to merge.

PR #467, opened 2026-08-27
Review only, do not merge: replay of #462 (7c688ed) for a deferred bot review

abhijithanandan:
## Review-only. Do not merge. This PR will be closed.

This carries no new change. It replays the diff that already merged into `development`
so an automated review can run over it, and it will be closed and its branch deleted
once that review has been read.

- Original PR: #462
- Merged commit: `7c688ed` "fix(inventory): complete the movement-history entry and fetch its associations"
- Branch base: `f8d0e84`, the first parent of that commit, so the diff here is byte for
  byte what merged.

The original merged inside a window where the review bot's hourly allowance was already
spent. Its status check reported a pass with the summary "Review rate limited", so nothing
was actually reviewed. This recovers that review after the fact.

### What the change does

The movement-history entry omitted the running balance and the creator although the ledger
row carries both, so a timeline built on it could not show the per-row balance line. Adds
`openingStock`, `closingStock` and `createdByName`, with the creator flattened to a display
name the same way the storage location and project already flatten to theirs, and left null
for a row with no recorded creator.

The finder also fetched nothing. `project`, `storageLocation` and `createdBy` are all lazy
`@ManyToOne` and the mapper reads a field off each, so a page cost one query plus two per row
before the change and would have cost three per row after it. All three are now
`LEFT JOIN FETCH`-ed, which stays a to-one join, so it neither multiplies rows nor forces
Hibernate to paginate in memory. The count query is written out explicitly because a derived
count cannot be built from a query carrying fetch joins.

### Points to weigh

- Whether the explicit `countQuery` stays in step with the main query, and what happens if
  one is edited without the other.
- Whether the fetch joins interact correctly with pagination and with the `orgFilter` tenant
  filter.
- Null handling on the flattened creator name.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/inventoryTransaction/InventoryTransactionController.java
+++ b/src/main/java/org/tornotron/echno_backend/inventoryTransaction/InventoryTransactionController.java
@@ -118,8 +118,9 @@ public class InventoryTransactionController {
             summary = "Get a material's movement history (timeline, paginated)",
             description = "Returns the material's stock movements as a timeline, oldest movement first, one "
                     + "page at a time. Each entry carries the storage location, project, movement type and its "
-                    + "stock direction, the quantity changed, the timestamp and the source reference, so a "
-                    + "caller can show where the material has been, when, and how much."
+                    + "stock direction, the quantity changed, the stock level either side of it, the timestamp, "
+                    + "the source reference and the employee who booked it, so a caller can show where the "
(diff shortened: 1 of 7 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
The review allowance was already spent, so this replay was rate limited and no review ran. Closing as planned; this PR was never intended to merge. #462 still has no bot review.

Issue #473, opened 2026-08-27
26 unpaginated findAll() calls reach controllers, so those responses grow without bound


abhijithanandan:
26 repository `findAll()` calls reach a controller with no pagination and no limit. The response grows with the table, forever.

This came out of an audit of the DTO conversion layer. It is worth separating from that work, because it is the one axis batch fetching does nothing for. `hibernate.default_batch_fetch_size=100` (`application.yml:247`, added in `06b0f48`) keeps the *query count* flat as a result set grows, and it does that well: a paged list holds at 6 statements whether the page is 2 rows or 4. What it cannot do is keep the *row count* bounded. An unpaginated `findAll` loads every row, maps every one to a DTO, and serialises the lot.

On a demo tenant that is invisible. On a real construction client with a few years of tasks, issues, materials and attendance behind them, it is the first thing that falls over, and it fails as a slow response and a memory spike rather than a clean error.

## What needs doing

Find the 26 call sites and decide per endpoint, because they are not all the same:

- Genuine list endpoints get pagination, matching the `Page`/`Pageable` pattern already used elsewhere in the codebase.
- Small reference lookups that are bounded by their nature (a short enum-like table, a per-org settings row) can stay, but should be stated as deliberate rather than left ambiguous.
- Anything feeding a dropdown probably wants a search-and-limit rather than a full list.

Changing a response shape from a bare array to a page is a breaking change for `echno-core` and `echno-web`, so this needs sequencing per endpoint rather than a single sweep: backend, then a core release, then web. Worth grouping the endpoints by whether the frontend consumes them today.

## Guardrail

Once the genuinely unbounded ones are paginated, an architecture test asserting that no controller returns a bare `List<T>` from a `findAll()` would stop the pattern coming back. Worth considering, since this is the second time a scale problem in this area has been found by audit rather than by anything failing.

abhijithanandan:
## The inventory, recorded so it does not have to be rebuilt

Stage-one output for this issue. Posting it here because the count in the title is right but incomplete, and the next person to run the obvious grep will land on 26 again and miss the same three sites.

### The count

`grep -rn "findAll()"` returns **28** hits in `src/main/java`. Two are not repository calls: `CompanyBankAccountService:30` is that class's own method *declaration* (`public List<CompanyBankAccountDto> findAll()`), and `CompanyBankAccountControllerWeb:46` calls that service method. Discard those and it is **26 repository calls**, so the audit figure holds.

### Three more the grep misses

[...]

PR #484, opened 2026-08-27
Guard against unbounded repository reads, and record the deliberate ones

abhijithanandan:
Part two of #473, stacked on #483. The guardrail, plus the record of which unbounded reads are deliberate.

**Review #483 first**; this one's allowlist assumes its deletions have landed.

## The rule

The issue suggested asserting that no controller returns a bare `List<T>` fed by an unpaginated `findAll()`. I did not build that one. Proving reachability from a repository call, through the service layer, to a controller return type needs transitive call-graph analysis, which is where architecture rules turn fragile: they fail for the wrong reasons and point at the wrong line.

The rule here is simpler and strictly stronger:

> No production class may call a Spring Data `findAll()` that has no limit on the rows it returns.

It has no false positives, it catches the CSV and PDF paths a controller-shaped rule would miss, and it fails at the line that introduced the problem rather than somewhere downstream.

`findAll(Sort)` counts too. Sorting an unbounded read still returns every row, and it is exactly the variant the original audit's grep walked past: that grep found 26 sites and missed three more, in the chart of accounts and the cost-category table. `findAll(Pageable)` is bounded and is not matched.

Modelled on `EndpointAuthorizationTest`, the ratchet already in this package that has been driven to an empty allowlist. Pure ArchUnit over compiled classes, no Spring context.

## A heap regression I caused, and the fix

Worth reading before reviewing the test, because it changed how both architecture tests are written.

My first version imported its own class graph with `ClassFileImporter`, exactly as `EndpointAuthorizationTest` already did. That took the suite from clean to **18 failures**, all of them `Failed to load ApplicationContext` or `Could not open JPA EntityManager` downstream of an `OutOfMemoryError`. Not one was an assertion failure, which is what makes this failure mode easy to misread as flakiness or a slow runner.

I did not assume it was pre-existing. Same box, same command, back to back:

| | classes | tests | failures | OOM | result |
|---|---|---|---|---|---|
| `origin/development` | 97 | 463 | 0 | 0 | SUCCESSFUL |
| this branch, first version | | 467 | 18 | 2 | FAILED |

The cause: ArchUnit holds a full ~1100-class graph in memory, two test classes each held their own as a **strong** static reference, and the test JVM is capped at `maxHeapSize = 1024m` with no `forkEvery`, so a single JVM also accumulates a cached Spring context per distinct test configuration for the whole run.

The fix is not "import it once". `archunit-junit5:1.3.0` builds its own cache with `CacheBuilder.newBuilder().softValues()`, so a graph obtained through `@AnalyzeClasses` is **soft-referenced and reclaimable** under exactly this pressure, and shared across every test with the same import spec. Both architecture tests now use `@AnalyzeClasses` with `@ArchTest` members. Same rules, same allowlist, same assertions; only the import path changed. The architecture run also dropped from 3m39s to 2m01s.

Two follow-on notes for reviewers:

- `@AnalyzeClasses` should be the house style for this package. A third architecture test written the obvious way would hit this again.
- The suite has thin headroom under a 1024m cap with no `forkEvery`. My branch tipped it, but 463 tests accumulating in one JVM is already close to the line. That is worth its own look, and deliberately **not** changed here: the cap is documented against the 2 vCPU / 7 GB CI runner and raising it is a shared decision.

## Evidence the rule works

Run with the allowlist emptied, it reported exactly 24 call sites: the 21 `findAll()` calls remaining after the other PR, plus the 3 `findAll(Sort)` calls. That matches the hand-built inventory line for line, which is the real proof the rule catches what it claims to rather than passing vacuously.

Two further rules keep the allowlist honest in the other direction. One fails if an entry no longer makes the call, so entries cannot outlive the reads they excused and the set can only shrink. The other asserts an allowlisted simple name identifies exactly one class, since simple-name matching is only safe while no two production classes share one (none do today).

## The allowlist carries its own reasons

Every entry has the reason next to it, so a deliberate read is recorded by the allowlist itself rather than by a comment somewhere else. The set is split in two, because the halves are read differently:

**Permanent (6).** Bounded by what the table *is*. A chart of accounts is bounded by the accounting structure a tenant defines, not by what it records against that structure; adding a year of trading adds journal entries, not accounts, and those are paged. Same argument for company bank accounts, cost categories, leave policies and the billing feature catalogue.

**Temporary (15).** Genuinely unbounded, waiting on the client sequencing. These shrink as #473 progresses.

Row counts today justify neither half. On a per-client production model one tenant's small reference table is another's long tail, so an entry earns the permanent half only when the ceiling comes from the nature of the data.

The six permanent entries also carry the argument on the method itself, so it is visible to whoever next reads the code and not only to whoever runs the test.

Refs #473

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/billing/services/FeatureService.java
+++ b/src/main/java/org/tornotron/echno_backend/billing/services/FeatureService.java
@@ -23,6 +23,16 @@ public class FeatureService {
         return featureRepository.findByIsActiveTrueOrderByCategory();
     }
 
+    /**
+     * Returns the whole billing feature catalogue, including features not currently on sale.
+     *
+     * <p>Deliberately unpaginated. The catalogue is a fixed enum-like table describing what the
+     * product can sell; it changes when the product changes, not with tenant activity, and it is
+     * not per-tenant data. Callers need it whole in order to resolve a subscription's feature
+     * codes.
+     *
+     * @return Every feature in the catalogue.
+     */
     public List<Feature> getAllFeatures() {
         return featureRepository.findAll();
     }
--- a/src/main/java/org/tornotron/echno_backend/finance/bank/service/CompanyBankAccountService.java
+++ b/src/main/java/org/tornotron/echno_backend/finance/bank/service/CompanyBankAccountService.java
@@ -26,6 +26,16 @@ public class CompanyBankAccountService {
     private final CompanyBankAccountMapper mapper;
     private final TenantEntityHelper tenantEntityHelper;
 
+    /**
+     * Returns every bank account the tenant has registered, active and closed.
+     *
+     * <p>Deliberately unpaginated. The set is bounded by how many accounts a company operates,
+     * which is a decision it makes, not by how long it has operated or how much it has moved
+     * through them. Payments and receipts against these accounts are separate tables and are
+     * paged.
+     *
+     * @return Every company bank account in the current tenant.
+     */
     @Transactional(readOnly = true)
     public List<CompanyBankAccountDto> findAll() {
         return mapper.toDtos(repository.findAll());
--- a/src/main/java/org/tornotron/echno_backend/finance/budget/service/CostCategoryService.java
+++ b/src/main/java/org/tornotron/echno_backend/finance/budget/service/CostCategoryService.java
@@ -34,6 +34,17 @@ public class CostCategoryService {
     private final AccountRepository accountRepo;
     private final TenantEntityHelper tenantEntityHelper;
 
+    /**
+     * Returns the tenant's cost categories, optionally only the active ones.
+     *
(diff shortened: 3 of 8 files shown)


abhijithanandan:
Note on checks: `pull-request.yml` triggers only on `pull_request` into `development`, so no CI runs on this PR while it is based on `feat/cap-unbounded-list-reads-473`. Absent checks here are a consequence of stacking, not a skipped or failing build.

It is verified locally in the meantime. The full suite run on lab `.116` was against **this branch's head** (which contains both commits), not just #483:

```
BUILD SUCCESSFUL in 7m 26s
classes=100 tests=476 skipped=0 failures=0 errors=0
```

Zero `OutOfMemoryError`, against a back-to-back baseline run of pristine `origin/development` on the same box (97 classes, 463 tests, 0 failures) so the comparison is real.

Once #483 merges, retarget this PR's base to `development` and CI will run on it normally.

PR #487, opened 2026-08-27
Fix an inspection's project at creation instead of letting an edit move it

abhijithanandan:
ClickUp: [86d45ju6b Approval should be mapped to a fixed project](https://app.clickup.com/t/86d45ju6b)

Backend half. The frontend half is tornotron/echno-web#283.

## What it actually was

`InspectionService.update` wrote `req.projectId()` onto the record with no comparison at all:

```java
inspection.setProjectId(req.projectId());   // straight overwrite
```

and `PUT /api/v1/inspections/web/{id}` is a full replacement that the web client always sends a `projectId` on. So any caller with `system-admin` or `project-manager` could repoint an inspection at any other project in the tenant.

Two things ride on that association:

1. A legally mandated pre-construction approval is obtained for one project. If the record can be moved, its value as a compliance record goes with it.
2. AI-generated compliance inspections are **identified** by their project. `ComplianceGenerationService` skips a rule it has already produced using `existsByProjectIdAndComplianceRuleRefAndOrganization_Id(projectId, ruleCode, orgId)`. Move one and the original project looks as though the compliance was never generated, so the next regeneration creates a duplicate, while `complianceRuleRef`, `compliancePhase`, `riskLevel` and `aiRationale` stay pointing at rules for the project it left.

## What changed

`update` compares before it writes:

```java
requireSameProject(inspection, req.projectId());
```

- A different id is refused with `InvalidRequestException`, which the global handler maps to 400. The check runs before the first setter, so a rejected request cannot leave the other fields replaced.
- A matching id passes. The web client sends the whole record back on every save, so the stored id arriving unchanged is the normal edit, not an attempted reassignment.
- An omitted id passes and leaves the stored project alone. That also closes a smaller hole: a payload without `projectId` used to null the project out.

The `@Schema` on `UpdateInspectionRequest.projectId`, the record's own docblock and the controller's `@Operation` now say the project is fixed, matching how `ExpenseUpdateDto` and `UpdateCustomerRequest` document their own immutable fields.

Nothing is removed from the wire: `projectId` stays on the request so existing clients keep working, and the create path is untouched.

## Verification

New `InspectionProjectImmutabilityTest`, plain JUnit and Mockito, no Spring context (the suite caches every distinct context and the test JVM heap is capped, so this rule did not warrant one):

```
InspectionProjectImmutabilityTest > update_rejectsAMoveToAnotherProject() PASSED
InspectionProjectImmutabilityTest > update_acceptsThePayloadThatRepeatsTheStoredProject() PASSED
InspectionProjectImmutabilityTest > update_leavesTheProjectInPlaceWhenThePayloadOmitsIt() PASSED
InspectionProjectImmutabilityTest > update_stillReportsAnInspectionThatDoesNotExist() PASSED
```

Confirmed load bearing by restoring the old `setProjectId` line and dropping the guard: `update_rejectsAMoveToAnotherProject` and `update_leavesTheProjectInPlaceWhenThePayloadOmitsIt` both FAILED, then passed again once restored.

Full inspection package on the lab build box (`10.24.6.116`), where Testcontainers has a Docker daemon:

```
$ ./gradlew test --tests "org.tornotron.echno_backend.inspection.*"
InspectionProjectImmutabilityTest  4 PASSED
InspectionServiceIT                1 PASSED
InspectionTaxonomyEnumsTest        9 PASSED
InspectionTaxonomyMigrationIT      5 PASSED
BUILD SUCCESSFUL in 4m 47s
```

`./gradlew compileJava compileTestJava` clean.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/inspection/dtos/UpdateInspectionRequest.java
+++ b/src/main/java/org/tornotron/echno_backend/inspection/dtos/UpdateInspectionRequest.java
@@ -17,10 +17,11 @@ import java.time.LocalDateTime;
 import java.util.List;
 
 /**
- * Full replacement of an inspection. Status is set directly and the result is
- * optional (set once the inspection is concluded). The category falls back to the
- * one derived from the type when it is omitted. The check items and defects are
- * rebuilt from the payload and the summary counts recomputed from them.
+ * Full replacement of an inspection, apart from the project it is against, which
+ * is fixed at creation. Status is set directly and the result is optional (set
+ * once the inspection is concluded). The category falls back to the one derived
+ * from the type when it is omitted. The check items and defects are rebuilt from
+ * the payload and the summary counts recomputed from them.
  */
 @Schema(description = "Payload to fully replace an existing inspection, including its status, result, checklist items and defects.")
 public record UpdateInspectionRequest(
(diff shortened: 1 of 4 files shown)


abhijithanandan:
@coderabbitai review

PR #488, opened 2026-08-27
Let a project state its city, state and PIN code instead of one short line

abhijithanandan:
ClickUp: [86d45jbvr Location Selection while creating a new Project](https://app.clickup.com/t/86d45jbvr)

Backend half. The shared-library half is tornotron/echno-core#40; the web form change follows once that is released. See the ClickUp comment for the sequencing.

## What the ticket is really about

The ticket reads as a form complaint, and the form is the visible half, but the reason it matters is in `ComplianceGenerationService`:

```java
String state = IndianStateResolver.resolve(project.getProjectAddress());
if (state == null) {
    throw new InvalidRequestException("The project address does not include a recognised state, …");
}
```

Compliance rules are keyed by `(state, projectType)`. There was no state field, so the state was recovered by scanning the free-text address for one of 36 state names. `IndianStateResolver`'s own comment said as much: *"Deliberately simple: a structured state field on the project can replace this later."*

That scan cannot find what the address does not say. The address in the ticket's screenshot is "Chennai", which is a complete and ordinary thing to type and contains no state, so compliance generation for that project refused to run and told the user to go and rewrite their address. And `ProjectCreationDto` capped the address at **50 characters**, which is not much room for a street, a city and a state.

So "for the system to effectively build compliance rules" is the operative clause, and the fix is a state field.

## What changed

**New optional columns on `project`** (changelog `v4.0/056`, all nullable, `MARK_RAN`-guarded like its neighbours): `project_city`, `project_state`, `project_postal_code`, plus an index on state. Existing rows have none of them and keep working off the address line.

**State is stored canonically and validated at entry.** `IndianStateResolver` gains a public `canonicalise(String)` that maps any casing of a state or union territory to its canonical spelling and refuses anything that is not one. `ProjectService` calls it on create and on patch, translating the failure into `InvalidRequestException` (400). Storing the canonical spelling is what makes the rule lookup match; refusing a non-state at entry is what stops a typo surfacing much later, at generation time, with nothing pointing at the cause.

**Generation prefers the stated state**, falling back to the address scan:

```java
String state = project.getProjectState() != null
        ? project.getProjectState()
        : IndianStateResolver.resolve(project.getProjectAddress());
```

The error message when neither yields a state now names the field rather than telling the user to edit their address.

**`projectAddress` allows 255 characters** instead of 50. The column was always `TEXT`; only the bean-validation cap moves.

**Coordinates become optional.** `@NotNull` came off `projectLatitude` / `projectLongitude` on the create DTO. The web form has labelled that whole block "(optional)" since it was written and sends `undefined` when the fields are blank, so a project described by address alone could not be created at all. Nothing in the backend reads the coordinates (`grep getProjectLatitude` finds only the two assignments in `ProjectService`), and the client-side geofence already skips a project that has none. They gain the `[-90, 90]` / `[-180, 180]` range check that the patch path has always had and create never did.

`city` and `postalCode` are display only; nothing branches on them.

## What is deliberately not here

- **No geocoding.** Turning a typed address into coordinates needs a provider, a key and a cost decision. Flagged on the ticket.
- **No change to whether any of this is mandatory.** Everything new is optional. Flagged on the ticket.

## Verification

Run on the lab build box (`10.24.6.116`), so the Testcontainers suites have a real CockroachDB and Liquibase actually applied changelog 056:

```
$ ./gradlew test --tests "…project.*" --tests "…compliance.*" --tests "…inspection.InspectionServiceIT"
ComplianceGenerationServiceIT      1 PASSED
ComplianceGenerationServiceTest    8 PASSED
IndianStateResolverTest            6 PASSED
OpenAiCompatibleComplianceServiceTest 4 PASSED
InspectionServiceIT                1 PASSED
ProjectControllerWebAuthzTest      3 PASSED
BUILD SUCCESSFUL in 5m 18s
```

New tests: `IndianStateResolverTest` (canonical spelling, blank, non-state refusal, and the address scan finding nothing in "Chennai", which is the case the field exists for), plus two on `ComplianceGenerationServiceTest` pinning that a stated state is used and beats a different one named in the address. Confirmed load bearing by reverting the preference to the old address-only line: both new tests FAILED, then passed once restored.

`./gradlew compileJava compileTestJava` clean. Plain JUnit and Mockito throughout, no new Spring context.

## Summary by CodeRabbit

- **New Features**
  - Projects now support separate city, state/union territory, and postal code fields.
  - Project addresses can be up to 255 characters and are documented as single-line street addresses.
  - Site coordinates are optional, while entered values must remain within valid geographic ranges.

- **Bug Fixes**
  - Compliance generation now uses the project’s stated location before analyzing its free-text address.
  - Invalid or non-Indian states are rejected with clearer validation messages.
  - Address fields are cleaned automatically when projects are created or updated.

---

### The create payload's constraints were never running

`ProjectController.createProject` and its `/web` twin both declare `@RequestPart("data") @Valid String data` and then `objectMapper.readValue(data, ProjectCreationDto.class)`. The `@Valid` applies to the `String`, which carries no constraints, so Spring never validated the DTO.

This predates the branch: `@NotBlank` on `projectName` has been unenforced since the endpoint was written, and both endpoints document a 400 for "a field failed validation" that could not occur. But adding the address, state and PIN constraints on top of a bypass is worse than not adding them, because the next reader would reasonably assume the 255-character limit holds.

Fixed in `ProjectService.addProject` rather than in either controller. Those two are the only callers of `addProject` (`grep -rn "addProject(" src/main` confirms it), so one call covers both entry points and any later caller by construction. A `ConstraintViolationException` is not something Spring raises from its own argument binding, so `GlobalExceptionHandler` had no mapping and it would have surfaced through the catch-all as a 500; it now returns the same 400 with the same `errors` map as the bound path.

The limit is validation only. `project_address` has been `TEXT` since `v1.0/010-create-project.xml`, so there is nothing to migrate.

### The same pattern is on six other endpoints

Swept for it rather than assuming. `@RequestPart` bound to a `String` and then deserialized by hand appears on:

| Endpoint | DTO | Constraints unenforced |
|---|---|---|
| `TaskController.createTask` | `TaskCreationDto` | 9 |
| `TaskControllerWeb.createTask` | `TaskCreationDto` | 9 |
| `IssueController.createIssue` | `IssueCreationDto` | 6 |
| `IssueControllerWeb.createIssue` | `IssueCreationDto` | 6 |
| `OrganizationController.createOrganization` | `OrganizationCreationDto` | 10 |
| `OrganizationWebController.createOrganization` | `OrganizationCreationDto` | 10 |
| `ChatControllerWeb` send message | `SendMessageDto` | 1 |

None are touched here. They are filed as #490. The partial-update endpoints that read into a `Map<String, Object>` are unaffected, since there is no bean to validate. The `ConstraintViolationException` handler this PR adds is what those fixes will need, so it is deliberately general rather than project-specific.

### The migration guarded one object for four

`MARK_RAN` records a changeSet as run without executing it, so a single precondition on `project_state` would let a database holding that column and none of the rest be stamped complete and never repaired. Split into four changeSets, one per column and one for the index, each with its own condition.

### Verification of the follow-up

New `ProjectCreationValidationTest`, a real Hibernate Validator with mocked collaborators, no Spring context: a blank name (also asserting the repository is never touched), an over-length address, an impossible latitude, an over-length postal code, and a 255-character address getting through. Confirmed it fails against the bypass by deleting the `requireValid` call: **4 of 5 FAILED**, then passed once restored.

Re-run on the lab build box against a real CockroachDB, so Liquibase applied the reworked migration for real:

```
ProjectCreationValidationTest         5 PASSED
ProjectControllerWebAuthzTest         3 PASSED
IndianStateResolverTest               6 PASSED
ComplianceGenerationServiceTest       8 PASSED
ComplianceGenerationServiceIT         1 PASSED
OpenAiCompatibleComplianceServiceTest 4 PASSED
BUILD SUCCESSFUL in 5m 1s
```

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/exception/GlobalExceptionHandler.java
+++ b/src/main/java/org/tornotron/echno_backend/common/exception/GlobalExceptionHandler.java
@@ -10,6 +10,8 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.ProblemDetail;
 import org.springframework.security.authorization.AuthorizationDeniedException;
 import org.springframework.validation.FieldError;
+import jakarta.validation.ConstraintViolation;
+import jakarta.validation.ConstraintViolationException;
 import org.springframework.web.bind.MethodArgumentNotValidException;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.ResponseStatus;
@@ -137,6 +139,27 @@ public class GlobalExceptionHandler {
             String errorMessage = error.getDefaultMessage();
             errors.put(fieldName, errorMessage);
         });
+
+        ProblemDetail pd = problem(HttpStatus.BAD_REQUEST, "Validation Failed", "Validation Failed", request);
+        pd.setProperty("errors", errors);
+        return pd;
+    }
+
+    /**
+     * The same 400 as above, for a bean validated outside Spring's own argument binding. A
+     * payload that arrives as the JSON string part of a multipart request has to be
+     * deserialized and validated by hand, and that raises this rather than
+     * {@link MethodArgumentNotValidException}. Without this it would fall through to the
+     * catch-all handler and be reported as a 500.
+     */
+    @ExceptionHandler(ConstraintViolationException.class)
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    public ProblemDetail handleConstraintViolationException(ConstraintViolationException exception,
+                                                           WebRequest request) {
+        Map<String, String> errors = new HashMap<>();
+        for (ConstraintViolation<?> violation : exception.getConstraintViolations()) {
+            errors.put(violation.getPropertyPath().toString(), violation.getMessage());
+        }
 
         ProblemDetail pd = problem(HttpStatus.BAD_REQUEST, "Validation Failed", "Validation Failed", request);
         pd.setProperty("errors", errors);
--- a/src/main/java/org/tornotron/echno_backend/compliance/ComplianceGenerationService.java
+++ b/src/main/java/org/tornotron/echno_backend/compliance/ComplianceGenerationService.java
@@ -89,12 +89,17 @@ public class ComplianceGenerationService {
                             + "Commercial) before generating compliance.");
         }
 
-        String state = IndianStateResolver.resolve(project.getProjectAddress());
+        // A project that states its own state is taken at its word; scanning the free-text
+        // address is only the fallback for projects that predate the field, and it cannot find
+        // what the address does not say (an address of "Chennai" names no state at all).
+        String state = project.getProjectState() != null
+                ? project.getProjectState()
+                : IndianStateResolver.resolve(project.getProjectAddress());
         if (state == null) {
             throw new InvalidRequestException(
-                    "The project address does not include a recognised state, so the applicable "
-                            + "regulations cannot be determined. Add the state to the project address "
-                            + "(for example 'Chennai, Tamil Nadu') and try again.");
+                    "This project has no state set, and its address does not name one, so the "
+                            + "applicable regulations cannot be determined. Set the project's state "
+                            + "(for example Tamil Nadu) and try again.");
         }
 
         List<ComplianceRule> candidateRules =
--- a/src/main/java/org/tornotron/echno_backend/compliance/IndianStateResolver.java
+++ b/src/main/java/org/tornotron/echno_backend/compliance/IndianStateResolver.java
@@ -4,13 +4,16 @@ import java.util.List;
 import java.util.Locale;
 
 /**
- * Derives the Indian state a project sits in from its free-text address. Compliance
- * rules are keyed by state, so generation needs a state name; projects only carry an
- * address string. The match is a case-insensitive substring scan against a fixed list
- * of states and union territories, returning the canonical name of the first hit.
- * Deliberately simple: a structured state field on the project can replace this later.
+ * The Indian states and union territories compliance rules are keyed by, and the two ways a
+ * project's state is arrived at.
+ *
+ * <p>{@link #canonicalise(String)} is the preferred one: a project that states its own state
+ * gives an exact answer. {@link #resolve(String)} is the fallback for the projects that predate
+ * the field, scanning the free-text address for a state name and returning the first hit. That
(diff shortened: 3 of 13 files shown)


abhijithanandan:
Both review findings were real and are fixed in e4a19e9.

**Validation was never running on this payload.** Confirmed, and worse than the finding put it. Both controllers declare `@RequestPart("data") @Valid String data` and then `objectMapper.readValue(data, ProjectCreationDto.class)`. The `@Valid` applies to a `String`, which carries no constraints, so Spring never validated the DTO. That was true before this branch too: the `@NotBlank` on `projectName` has been decorative since the endpoint was written, while `@ApiResponses` advertised a 400 for "a field failed validation".

Fixed in `ProjectService.addProject` rather than in each controller. Both controllers converge there, it is the only place that covers both without duplicating the call, and it covers any later caller by construction. A `ConstraintViolationException` is not one Spring raises from its own argument binding, so the global handler had no mapping for it and it would have fallen through the catch-all as a 500; it now returns the same 400 with the same `errors` map as the bound path.

New `ProjectCreationValidationTest`: a real Hibernate Validator with mocked collaborators, five tests covering a blank name (asserting the repository is never touched), an over-length address, an impossible latitude, an over-length postal code, and a 255-character address getting through. Confirmed load bearing by deleting the `requireValid` call: 4 of 5 FAIL, then pass once restored.

**Half-applied migration.** Correct on the mechanics, `MARK_RAN` records the changeSet as run without executing it, so a database holding `project_state` and nothing else would be stamped complete and never repaired. Split into four changeSets, one per column and one for the index, each with its own precondition. The reason is now a comment in the file.

Re-verified on the lab build box against a real CockroachDB, so Liquibase applied the reworked migration for real:

```
ProjectCreationValidationTest        5 PASSED
ProjectControllerWebAuthzTest        3 PASSED
IndianStateResolverTest              6 PASSED
ComplianceGenerationServiceTest      8 PASSED
ComplianceGenerationServiceIT        1 PASSED
OpenAiCompatibleComplianceServiceTest 4 PASSED
BUILD SUCCESSFUL in 5m 1s
```

One note on the first finding's aside: yes, the 255-character limit is bean-validation only. `project_address` has been `TEXT` since `v1.0/010-create-project.xml`, so there is nothing to migrate, and the PR body says as much.

PR #489, opened 2026-08-27
Guard against unbounded repository reads, and record the deliberate ones

abhijithanandan:
Part two of #473, stacked on #483. The guardrail, plus the record of which unbounded reads are deliberate.

**Review #483 first**; this one's allowlist assumes its deletions have landed.

## The rule

The issue suggested asserting that no controller returns a bare `List<T>` fed by an unpaginated `findAll()`. I did not build that one. Proving reachability from a repository call, through the service layer, to a controller return type needs transitive call-graph analysis, which is where architecture rules turn fragile: they fail for the wrong reasons and point at the wrong line.

The rule here is simpler and strictly stronger:

> No production class may call a Spring Data `findAll()` that has no limit on the rows it returns.

It has no false positives, it catches the CSV and PDF paths a controller-shaped rule would miss, and it fails at the line that introduced the problem rather than somewhere downstream.

`findAll(Sort)` counts too. Sorting an unbounded read still returns every row, and it is exactly the variant the original audit's grep walked past: that grep found 26 sites and missed three more, in the chart of accounts and the cost-category table. `findAll(Pageable)` is bounded and is not matched.

Modelled on `EndpointAuthorizationTest`, the ratchet already in this package that has been driven to an empty allowlist. Pure ArchUnit over compiled classes, no Spring context.

## A heap regression I caused, and the fix

Worth reading before reviewing the test, because it changed how both architecture tests are written.

My first version imported its own class graph with `ClassFileImporter`, exactly as `EndpointAuthorizationTest` already did. That took the suite from clean to **18 failures**, all of them `Failed to load ApplicationContext` or `Could not open JPA EntityManager` downstream of an `OutOfMemoryError`. Not one was an assertion failure, which is what makes this failure mode easy to misread as flakiness or a slow runner.

I did not assume it was pre-existing. Same box, same command, back to back:

| | classes | tests | failures | OOM | result |
|---|---|---|---|---|---|
| `origin/development` | 97 | 463 | 0 | 0 | SUCCESSFUL |
| this branch, first version | | 467 | 18 | 2 | FAILED |

The cause: ArchUnit holds a full ~1100-class graph in memory, two test classes each held their own as a **strong** static reference, and the test JVM is capped at `maxHeapSize = 1024m` with no `forkEvery`, so a single JVM also accumulates a cached Spring context per distinct test configuration for the whole run.

The fix is not "import it once". `archunit-junit5:1.3.0` builds its own cache with `CacheBuilder.newBuilder().softValues()`, so a graph obtained through `@AnalyzeClasses` is **soft-referenced and reclaimable** under exactly this pressure, and shared across every test with the same import spec. Both architecture tests now use `@AnalyzeClasses` with `@ArchTest` members. Same rules, same allowlist, same assertions; only the import path changed. The architecture run also dropped from 3m39s to 2m01s.

Two follow-on notes for reviewers:

- `@AnalyzeClasses` should be the house style for this package. A third architecture test written the obvious way would hit this again.
- The suite has thin headroom under a 1024m cap with no `forkEvery`. My branch tipped it, but 463 tests accumulating in one JVM is already close to the line. That is worth its own look, and deliberately **not** changed here: the cap is documented against the 2 vCPU / 7 GB CI runner and raising it is a shared decision.

## Evidence the rule works

Run with the allowlist emptied, it reported exactly 24 call sites: the 21 `findAll()` calls remaining after the other PR, plus the 3 `findAll(Sort)` calls. That matches the hand-built inventory line for line, which is the real proof the rule catches what it claims to rather than passing vacuously.

Two further rules keep the allowlist honest in the other direction. One fails if an entry no longer makes the call, so entries cannot

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/billing/services/FeatureService.java
+++ b/src/main/java/org/tornotron/echno_backend/billing/services/FeatureService.java
@@ -23,6 +23,16 @@ public class FeatureService {
         return featureRepository.findByIsActiveTrueOrderByCategory();
     }
 
+    /**
+     * Returns the whole billing feature catalogue, including features not currently on sale.
+     *
+     * <p>Deliberately unpaginated. The catalogue is a fixed enum-like table describing what the
+     * product can sell; it changes when the product changes, not with tenant activity, and it is
+     * not per-tenant data. Callers need it whole in order to resolve a subscription's feature
+     * codes.
+     *
+     * @return Every feature in the catalogue.
+     */
     public List<Feature> getAllFeatures() {
         return featureRepository.findAll();
     }
--- a/src/main/java/org/tornotron/echno_backend/finance/bank/service/CompanyBankAccountService.java
+++ b/src/main/java/org/tornotron/echno_backend/finance/bank/service/CompanyBankAccountService.java
@@ -26,6 +26,16 @@ public class CompanyBankAccountService {
     private final CompanyBankAccountMapper mapper;
     private final TenantEntityHelper tenantEntityHelper;
 
+    /**
+     * Returns every bank account the tenant has registered, active and closed.
+     *
+     * <p>Deliberately unpaginated. The set is bounded by how many accounts a company operates,
+     * which is a decision it makes, not by how long it has operated or how much it has moved
+     * through them. Payments and receipts against these accounts are separate tables and are
+     * paged.
+     *
+     * @return Every company bank account in the current tenant.
+     */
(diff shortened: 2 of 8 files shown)


abhijithanandan:
Recreated from #484, which GitHub auto-closed when its stacked base branch was deleted on merge of #483. Same commit, cherry-picked onto `development` so it now gets CI and a review, which it could not while stacked.

Issue #490, opened 2026-08-27
Multipart create endpoints never validate their payload


abhijithanandan:
Found while reviewing #488, which fixes this for `project` only. The rest are listed here rather than fixed there, to keep that PR to the area it owns.

## The bug

Every multipart create endpoint takes its payload as a JSON string part and deserializes it by hand:

```java
public ResponseEntity<TaskSimpleDto> createTask(@RequestPart @Valid String data, …) {
    TaskCreationDto dto = objectMapper.readValue(data, TaskCreationDto.class);
    return ResponseEntity.status(HttpStatus.CREATED).body(service.addTask(dto, attachments));
}
```

The `@Valid` is on the `String`. A `String` carries no constraints, so it validates nothing, and the DTO that comes out of `readValue` is never validated at all. Every bean-validation annotation on these creation DTOs is decorative, and each endpoint's `@ApiResponses` advertises a 400 for a field that failed validation, which cannot happen.

## Affected

| Endpoint | DTO | Constraints currently unenforced |
|---|---|---|
| `TaskController.createTask` | `TaskCreationDto` | 9 |
| `TaskControllerWeb.createTask` | `TaskCreationDto` | 9 |
| `IssueController.createIssue` | `IssueCreationDto` | 6 |
| `IssueControllerWeb.createIssue` | `IssueCreationDto` | 6 |
| `OrganizationController.createOrganization` | `OrganizationCreationDto` | 10 |
| `OrganizationWebController.createOrganization` | `OrganizationCreationDto` | 10 |
| `ChatControllerWeb` send message | `SendMessageDto` | 1 |

Already fixed in #488: `ProjectController.createProject` and `ProjectControllerWeb.createProject`.

The partial-update endpoints that read into a `Map<String, Object>` via `TypeReference` are not affected. There is no bean and nothing to validate.

## Suggested fix

The one used in #488: validate in the service rather than in each controller. Task, issue and organization each have a base controller and a `/web` twin that both call the same service method, so validating there covers both at once and covers any later caller by construction:

```java
private void requireValid(TaskCreationDto dto) {
    Set<ConstraintViolation<TaskCreationDto>> violations = validator.validate(dto);
    if (!violations.isEmpty()) {
        throw new ConstraintViolationException(violations);
    }
}
```

[...]

PR #492, opened 2026-08-27
Run the constraints on the task and issue create payloads

abhijithanandan:
Part of #490. Task and issue only; organization and chat follow in their own PRs so a problem in one area does not hold up the others.

## The bug

`TaskController`, `TaskControllerWeb`, `IssueController` and `IssueControllerWeb` all declare their create endpoint the same way:

```java
public ResponseEntity<TaskSimpleDto> createTask(@RequestPart @Valid String data, …) {
    TaskCreationDto dto = objectMapper.readValue(data, TaskCreationDto.class);
```

The `@Valid` is on the `String`. A `String` carries no constraints, so it validates nothing, and the DTO that comes out of `readValue` is never validated at all. Nine constraints on `TaskCreationDto` and six on `IssueCreationDto` have never once run, while each endpoint's `@ApiResponses` advertises a 400 for a field that failed validation, which could not happen.

## The fix

The shape #488 used for projects: validate in the service, at the single shared caller, so the `/web` twin and the base controller are covered by one change rather than two that can drift.

```java
private void requireValid(TaskCreationDto taskCreationDto) {
    Set<ConstraintViolation<TaskCreationDto>> violations = validator.validate(taskCreationDto);
    if (!violations.isEmpty()) {
        throw new ConstraintViolationException(violations);
    }
}
```

The `ConstraintViolationException` to 400 mapping in `GlobalExceptionHandler` already exists (#488 added it), so nothing is added there.

## Constraints that were themselves wrong

A rule that has never run has never been tested against a real request, and three of these are simply incorrect. Enforcing them as written would have started refusing payloads the product means to accept.

**Title, capped at 50 on both task and issue.** Both columns have always been `VARCHAR(255)`, and neither form caps the field: `task-form.tsx` and `issue-form.tsx` check a *minimum* of five characters and put no `maxLength` on the input. Titles over 50 have therefore been accepted for as long as the endpoints have existed. "Install and commission the rooftop HVAC units on Block B" is 56 characters. Both caps move to the column width.

**Issue description, capped at 500 with a message claiming a minimum of ten.** The column is `TEXT`. The form asks for at least twenty characters, offers no ceiling, and its own copy pushes towards prose. 500 characters is about one paragraph, so any issue report carrying reproduction steps would start failing. The cap moves to 2000 and the message stops claiming a minimum the constraint never asked for.

**`assigneeIds` and `tags`, required on a task.** This contradicts the service, which has always guarded both with `if (… != null && !…isEmpty())` and skipped them when absent. A task nobody has been assigned to yet is a state the code supports and the patch path can produce. Both become optional.

## One thing that is documentation, not validation

`TaskCreationDto.progress` is documented as "a fraction from 0 to 1". It is a percentage. `task-form.tsx` clamps the input with `Math.min(100, Math.max(0, …))` and every view renders it as `` `${task.progress}%` ``. The `@Schema` description says so now. No constraint was added: the patch path has never range-checked progress either, and pinning a range here on a value whose units the OpenAPI document has been describing wrongly is worth doing deliberately rather than as a side effect.

## What newly returns 400

This is the point of the change, but it is worth being exact about who is affected.

**Nothing on the ordinary create path.** `validateForm` in both web forms already refuses a blank title, a missing category, a missing employee and a description under twenty characters, and both send `status` and `progress` unconditionally. Every payload the normal flow produces still passes.

**The "Save as Draft" button on the issue form will start failing.** `handleSaveDraft` skips `validateForm` entirely and checks only the employee id and the title:

```js
function handleSaveDraft() {
  if (!currentEmployee?.id) { …; return; }
  if (!form.title.trim()) { toast.error('Please fill in the title before saving draft'); return; }
  props.onSubmit({ fields: form, attachments, isDraft: true });
}
```

`description` starts as `''` and the page sends it unconditionally, so the wire carries `"description": ""`, which `@NotBlank` refuses. Saving an issue draft with no description written yet currently succeeds and will return a 400. The form itself demands at least twenty characters on the ordinary path, so the constraint is right and the draft button is the thing that is wrong; it needs either its own endpoint or a pass through `validateForm`. Filed on the web repo, and worth landing there before this reaches a user-facing environment.

**The task form's draft button is unaffected**, despite skipping validation the same way: `addTask` already throws `InvalidRequestException` for a null `creatorId`, `projectId` or `categoryId`, so a draft missing a category is already a 400 today. The only change there is which 400 it is.

A title of one or two characters, reachable through either draft button, also starts failing `@Size(min = 3)`. Both forms ask for five on the ordinary path, so the constraint is looser than the product's own rule.

## Verification

```
$ ./gradlew test --tests "*TaskCreationValidationTest*" --tests "*IssueCreationValidationTest*"
IssueCreationValidationTest    8 PASSED
TaskCreationValidationTest     9 PASSED
BUILD SUCCESSFUL in 2m 47s
```

Confirmed load bearing by commenting out the two `requireValid` calls and rerunning: all 13 rejection tests FAILED, the two widening guards still passed (they are there to catch the caps being tightened again, not to prove enforcement). Restored, all 17 pass.

Plain JUnit, Mockito and a real Hibernate `Validator` built in the test, no Spring context, following `ProjectCreationValidationTest`.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/issue/IssueService.java
+++ b/src/main/java/org/tornotron/echno_backend/issue/IssueService.java
@@ -1,5 +1,8 @@
 package org.tornotron.echno_backend.issue;
 
+import jakarta.validation.ConstraintViolation;
+import jakarta.validation.ConstraintViolationException;
+import jakarta.validation.Validator;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
@@ -28,6 +31,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 @Service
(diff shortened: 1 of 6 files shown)


abhijithanandan:
The client-side half is up: tornotron/echno-web#291, green.

It removes the "Save as Draft" buttons rather than routing them through `validateForm()`. The short version of why: the button does not save a draft. Neither issue nor task has a draft state anywhere in the product, `isDraft` never leaves the browser, and the record created is an ordinary issue or task that lands in the list looking like every other one. The issue path forces `status` to `open` (which is already the form's default, so the branch can only ever discard a status the user deliberately picked) and the task path changes the toast to "Draft saved". Leave requests and purchase orders have real drafts, as server-side statuses with their own submit flow; these are not that.

Also worth knowing: the issue button was already half broken. `addIssue` dereferences `getTaskId()` unguarded, so a draft raised before picking a related task has always come back a 404 reading "Task with ID null was not found in this organization".

The genuine need the button was standing in for now has a genuine answer: the local form draft in tornotron/echno-web#288, merged yesterday, covers both these forms and keeps half-typed work through a lost session without putting anything on the server.

Rebased onto current `development` after #495 and #500 landed. No conflicts, no changes to the diff, and both suites re-run green on the rebased head:

```
IssueCreationValidationTest    8 PASSED
TaskCreationValidationTest     9 PASSED
BUILD SUCCESSFUL in 53s
```

PR #494, opened 2026-08-27
Stop the web task listing at ten rows without saying so

### LINKED ISSUE #478: GET /tasks/web silently returns only the first 10 tasks
`GET /api/v1/tasks/web` paginates by default and returns a bare array, so every caller silently gets the first 10 tasks and nothing says there are more.

Found while scoping #473. It is a separate defect and a worse one: #473 is a scale problem that shows up as a slow response, this one is a correctness problem that shows up as wrong data on a screen that looks fine.

## The call chain

`TaskControllerWeb.java:89-105`:

```java
@GetMapping
public ResponseEntity<List<TaskDto>> readAllTasks(@RequestParam(defaultValue = "0") int pageNo,
                                                  @RequestParam(defaultValue = "10") int pageSize) {
    Page<TaskDto> tasks = service.getAllTasks(pageNo, pageSize);
    return new ResponseEntity<>(tasks.getContent(), HttpStatus.OK);
}
```

The endpoint is paginated, defaults to `pageSize=10`, and then throws the `Page` wrapper away with `.getContent()`. A caller receives a plain JSON array with no `totalElements`, no `totalPages`, no way to tell a complete result from a truncated one. The OpenAPI description says so ("only the page content is returned, without paging metadata") but nothing in the response does.

`echno-core` calls it with no query parameters at all, so it always takes the defaults:

`src/services/task-service.ts:72`

```ts
async getAll(): Promise<Task[]> {
  const data = await api.get<ApiResponse[]>('/tasks/web');
  return safeParseTasks(data);
}
```

The doc comment above that method reads "Fetches every task visible to the current user". It fetches ten.

That feeds `useTasks()` (`src/hooks/task/use-tasks.ts:25`), which `echno-web` renders in the application shell (`features/common/components/app-layout.tsx:26`), so it is on every route.

It also feeds `useTasksByProject()` (`src/hooks/task/use-tasks.ts:82`), which fetches the "full" list and filters client-side by `projectId`. That filter runs over 10 rows drawn from the whole tenant with no ordering guarantee beyond `id ASC`, so a project whose tasks are not among the ten lowest ids renders as having no tasks at all.

## Impact

On any tenant with more than ten tasks, which is every real one, the task list is wrong and the per-project task view is wrong in a way that looks like missing data rather than a bug. `IndentControllerWeb` `/all` has the same `.getContent()` shape, though its bare `@GetMapping` is separately unpaginated and is covered by #473.

## Suggested fix

Return `Page<TaskDto>` rather than the unwrapped content, so a truncated result is self-describing, then have `echno-core` pass paging parameters and expose a paged result. That is a response shape change, so it sequences the same way as #473: backend, then a core release, then web. `useTasksByProject` should stop filtering client-side and use a real per-project endpoint (`GET /tasks/web/projectId/{projectId}` already exists at `TaskControllerWeb.java:107`).

abhijithanandan:
Fixes #478. **Step 1 of 3** (backend → core release → web).

## The defect

`GET /api/v1/tasks/web` took `pageNo`/`pageSize` defaulting to ten, then answered with `page.getContent()`. The envelope went in the bin, so a caller got a bare array with no total and no way to tell a complete result from a truncated one. No caller passed the parameters, so on any tenant with more than ten tasks every screen fed by this endpoint showed ten and looked finished: the all-tasks page, the dashboard's active-task count, the material-consumption task picker, and every per-project task view, since `useTasksByProject` filters that same truncated list client-side.

## What changed

The paging parameters are gone from the bare listing rather than re-tuned. Keeping them on an endpoint that discards the envelope is what made the truncation silent, and a larger default would only have moved the cliff. The listing is now bounded by `UnpagedResultCap`, the mechanism already used for the other bare-array list endpoints: `X-Total-Count` on every response, `X-Result-Capped` when rows were left out. A capped result says so.

Honest paging gets its own endpoint. `GET /tasks/web/paginated` returns a real `Page<TaskDto>`, optionally filtered by project or a free-text search on title and description, mirroring `/issues/web/paginated`. Page size is clamped to the same cap so one request cannot re-create the unbounded read, and search terms have their LIKE wildcards escaped.

No new unbounded read: the added repository query takes a `Pageable`, and the ArchUnit ratchet is untouched.

## Verification

`TaskControllerWebListingTest` stubs the service so it honours the page it was asked for. A stub that ignores the requested size passes against the truncating controller too, which is the blind spot that let this ship. Against the pre-fix controller body, five of the seven fail:

```
theBareListingDoesNotStopAtTheFirstTenTasks()          FAILED
theBareListingReachesTasksWellPastTheOldTenRowPage()   FAILED
theBareListingAsksForTheResultCapRatherThanAPageOfTen() FAILED
theBareListingReportsTheTrueTotalOnEveryResponse()     FAILED
theBareListingSaysSoWhenRowsWereLeftOut()              FAILED
```

The two `/paginated` tests pass in that run only because the endpoint they exercise is still present; without this change it does not exist.

All pass with the fix, alongside `TaskServicePaginationTest` (clamping, ordering, wildcard escaping) and the architecture ratchets.

## What this does not fix

`useTasksByProject` still fetches the full list and filters in the browser. Moving it onto `GET /tasks/web/projectId/{projectId}`, which already exists, is a pure echno-core change and lands in step 2.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/task/TaskControllerWeb.java
+++ b/src/main/java/org/tornotron/echno_backend/task/TaskControllerWeb.java
@@ -17,6 +17,7 @@ import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.multipart.MultipartFile;
+import org.tornotron.echno_backend.common.pagination.UnpagedResultCap;
 import org.tornotron.echno_backend.common.response.ApiResponse;
 import org.tornotron.echno_backend.common.service.AttachmentService;
 import org.tornotron.echno_backend.task.dto.TaskCreationDto;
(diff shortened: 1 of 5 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #496, opened 2026-08-27
Serve the navigation palette from a search endpoint

abhijithanandan:
Backend half of tornotron/echno-web#269. **Step 1 of 3** (backend → core release → web).

## Why a search endpoint rather than pagination

The web command palette is fed by whole-collection reads of projects, tasks and issues taken in the application shell, so they run on every route, and then filtered in the browser down to a few tens of rows.

I verified the consumer before designing against it. `command-palette.tsx` is not a count badge and not a paged list: it flattens all three collections plus the static nav routes into one `searchEntries` array, filters it by substring on a `keywords` field, ranks prefix matches above contains matches, and slices the ranked result to 50. It is a search box.

That settles the question the issue raised. Pagination does not serve it, because a palette wants whatever matches the query rather than an arbitrary first page. Two further findings from reading the consumer reinforce it:

- the shell pre-slices to 30 projects and 40 tasks/issues **before** the palette searches, so today it cannot find a project past the 30th or an issue past the 40th even when the data is loaded;
- tasks are capped at ten regardless, by echno-backend#478.

So the palette's search is already wrong, not merely expensive. Matching in the database fixes both.

## Design

**One endpoint, not one per kind.** `GET /api/v1/search/web?q=&limit=`. The consumer is a single search box showing a single ranked list, so per-kind endpoints would mean three requests per keystroke-debounce and three loading states to reconcile in the browser, and would push cross-kind ranking back into the client. The three kinds already share one read-authorization expression (`isMemberOfCurrentTenant() or hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')`, verified identical on `TaskControllerWeb`, `ProjectControllerWeb` and `IssueControllerWeb`), so serving them together gives up no authorization fidelity. The per-entity `/vendors/web/search` precedent stands where the consumer is a per-entity picker; this consumer is cross-entity.

**What it returns.** A flat array of `{type, id, title, projectId}`. That is exactly what the palette needs to pick an icon, show a label and build a link, and nothing more. Full DTOs would put back the transfer cost this exists to remove. `projectId` saves the client a second lookup to build a task's href.

**Bounding.** `limit` is per kind, defaults to 10, clamped to 25. A term under two characters returns empty without touching the database. LIKE wildcards in the term are escaped, so a bare `%` matches a per-cent sign rather than every row. No new unbounded read; the ArchUnit ratchet is untouched.

**Its own repository.** The three queries live in `SearchRepository` rather than being spread across the project, task and issue repositories. Search has one consumer and one shape, and keeping its queries together is what keeps them consistent. Every method carries an explicit `@Query`, so none is derived from the declared domain type.

## Verification

`SearchServiceTest` (plain Mockito): short-term short-circuit, wildcard and underscore escaping, limit default/clamp/floor, all three kinds returned together.

`SearchRepositoryIT` against a real CockroachDB, run on lab `.116` since these are constructor-expression JPQL strings that only fail when executed:

```
aProjectIsFoundByPartOfItsNameWhateverTheCase()        PASSED
aTaskIsFoundByItsTitleAndCarriesTheProjectItBelongsTo() PASSED
anIssueIsFoundByItsTitleAndCarriesTheProjectOfItsTask() PASSED
anIssueWithNoTaskIsStillFound()                        PASSED
theRowLimitBoundsTheResultRatherThanTheMatchCount()    PASSED
anEscapedWildcardMatchesItselfRatherThanEverything()   PASSED
BUILD SUCCESSFUL
```

`anIssueWithNoTaskIsStillFound` earns its place: an implicit `i.task.project.id` path join is an inner join and drops such rows silently, which is what the existing issue search does today.

## Not in this PR

The breadcrumb half of #269, and the web-side palette rewrite. The breadcrumb fix needs no backend work at all and ships now as a pure web change.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/search/SearchController.java
+++ b/src/main/java/org/tornotron/echno_backend/search/SearchController.java
@@ -1,0 +1,73 @@
+package org.tornotron.echno_backend.search;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.responses.ApiResponses;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.validation.annotation.Validated;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.List;
+
+/**
+ * Quick search for the web client's navigation palette.
+ *
+ * <p>One endpoint across projects, tasks and issues rather than one per kind, because the consumer
+ * is a single search box that shows a single ranked list. Three endpoints would mean three requests
+ * per keystroke and three loading states to reconcile in the browser, and would push ranking across
+ * kinds back into the client. The three kinds already share one read-authorization rule, so nothing
+ * is given up by serving them together.
+ */
+@RestController
+@RequestMapping("/api/v1/search/web")
+@Validated
+@Tag(
+        name = "Search",
+        description = "Quick search across the records the web client's navigation palette can jump "
+                + "to. Returns a small ranked set of matches rather than a page of a collection, "
+                + "because a search box wants what matches rather than what comes first."
+)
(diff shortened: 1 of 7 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

Issue #481, opened 2026-08-27
The test suite runs in one JVM with no forkEvery, so it is close to its heap cap and fails misleadingly


abhijithanandan:
The test suite has very little headroom, and the way it fails hides the cause. This has now cost three separate pieces of work time, so it is worth fixing the shape rather than continuing to warn people about it.

## The mechanism

`build.gradle` sets `maxHeapSize = 1024m` on the test JVM. The cap is deliberate and documented: it sizes the run to the 2 vCPU / 7 GB CI runner alongside the shared CockroachDB container. That part is correct and is not what needs changing.

The problem is that there is **no `forkEvery`**, so the entire suite runs in **one** JVM, and Spring caches a context per distinct test configuration for the life of that JVM. Every new context is permanent for the rest of the run. At 463 tests the suite now sits close enough to the ceiling that a single addition can tip it.

## Why it is worse than an ordinary limit

It fails somewhere else. When the heap runs out, the failures land as `Failed to load ApplicationContext` and `Could not open JPA EntityManager` in whatever test happened to run next, with **zero assertion failures** anywhere. So the symptom points at unrelated code, usually code the author never touched, which makes "this must be pre-existing" the natural and wrong conclusion.

Measured on 2026-08-27, same box, same command:

| | tests | failures | `OutOfMemoryError` |
|---|---|---|---|
| `origin/development` | 463 | 0 | 0 |
| A branch adding two ArchUnit imports | 467 | 18 | 2 |

All 18 were context-loading failures in six classes the branch did not touch.

Worth stating plainly: the cap is per-JVM, so machine memory is irrelevant. That run OOMed on a box with 103 GB of 125 GB free.

## Options

**`forkEvery`** is the obvious lever: a fresh JVM every N test classes discards accumulated contexts and bounds the peak. It costs JVM startup and re-warming per fork, so N wants measuring rather than guessing, and the Testcontainers-backed tests want thinking about since a fork boundary in the wrong place could churn containers.

**Reduce context proliferation.** Every distinct test configuration is another cached context. Some of the existing slices likely differ in ways that do not need to, and consolidating them buys headroom without touching the build.

**Raise the cap**, only if the runner genuinely has room. It is currently sized against 7 GB with a database container alongside, so this may be a smaller lever than it looks.

These are not exclusive, and the first two are the interesting ones.

## Also worth doing

[...]

PR #497, opened 2026-08-27
Give each trade a checklist an org owns, and copy it onto the inspection

abhijithanandan:
Phase 1 of the inspection module, part 1 of 3. Implements section 3.1 of
`docs/specs/2026-08-26-inspection-module.md`. Phase 0 (PR #469) is the base.

**Merge order: this PR, then the NCR workflow, then the RBAC split.** They stack in
that order; the second and third are branched off this one.

## What this adds

**A checklist library keyed by trade, in two populations that must not be confused.**

- `StarterChecklistTemplate` / `StarterChecklistTemplateItem` are global reference
  data shipped with the product, seeded by Liquibase for all sixteen
  `InspectionTrade` members (16 checklists, 95 check points), following the
  `ComplianceRule` precedent exactly: no `organization_id`, no org filter, read-only
  to the application. They give a new client a working system on day one.
- `ChecklistTemplate` / `ChecklistTemplateItem` are the org's own: `TenantScopedEntity`
  with the `orgFilter` and a `@ManyToOne Organization`, editable, one per trade per org.
- `POST /starters/{trade}/adopt` is the only bridge. It copies a starter into the
  tenant and cuts the link, so a later revision of a shipped checklist never rewrites
  criteria a client has tailored and signed work off against.

**Instantiation onto the inspection.** Creating an inspection for a trade starts it
from a copy of that trade's active template. A copy rather than a reference: an
inspection carried out in March records the criteria that were in force in March, and
a QA engineer tightening a tolerance in April must not rewrite what was signed off.

**Check-item extension.** `acceptanceCriterion`, `tolerance`, a server-computed
`deviation`, and a nullable `bimElementGuid` (unused, indexed, so the BIM phase finds
its link column already in place rather than migrating a live table).

## Decisions where the spec was silent

- **Template ownership** is tenant-owned with a global starter seed, as you decided.
  The seed could not live in `checklist_templates` itself: a row with a null
  `organization_id` would be invisible through the `orgFilter`, which is fail-closed by
  design. Hence the separate global pair of tables.
- **`version` is a revision counter, not a key.** The spec says "one template per trade
  per org, versioned by an integer", which are only consistent if the integer counts
  revisions of the single row. Unique constraint on (`organization_id`, `trade`);
  `update` bumps the counter. `active` then means "instantiated into new inspections",
  so deactivating retires a checklist without deleting a definition history references.
- **The trade is fixed at creation**, mirroring the existing project-immutability rule
  on `Inspection`. It is the key the template is found by, so moving one would silently
  repoint every future inspection of two trades at once.
- **No DELETE endpoint.** `active = false` retires a template; deleting one that
  historical inspections were created from destroys the definition without destroying
  the references to it.
- **`priority` on the template item**, which the spec does not list. Without it a
  template cannot fully instantiate an `InspectionCheckItem`, which already has the
  field. Same `'high' | 'medium' | 'low'` free text, same `medium` default.
- **Deviation is computed only where the subtraction means something.** Both values are
  free text an inspector typed, so the rule is: both must parse as a leading decimal,
  and the trailing unit must match case-insensitively. `15 cm` against `150 mm` yields
  null, not `-135`, because the answer would be in no unit at all and a reader would
  take it for millimetres. Null everywhere else, which is most of a finishing or
  documentation checklist. Rounded to the four decimals the column holds so a
  high-precision input is rounded here rather than rejected on insert.
- **An explicit check-item list on the create request always wins.** A template is a
  default, not an override; appending template rows to a hand-built list would double
  up every check point on the re-inspection of a failed item.
- **`bimElementGuid` on the check item only**, not on `InspectionDefect`. Section 6.3
  wants both eventually; only the check item is in this phase's scope.

## Two test-infrastructure fixes this needed

**Cleanup moved to `@AfterTransaction`.** `@AfterEach` runs inside the test
transaction, so deleting rows that transaction still holds write intents on made
CockroachDB wait for a transaction that could not commit until the delete returned. The
30s statement timeout then failed the test on cleanup with a query timeout and no failed
assertion, which is the shape of failure that gets blamed on an unrelated test. Seen on
CI and on `.116` before the fix; green on both after.

**The Spring test-context cache is now capped at 8** (`build.gradle`). This addresses
**#481** and it does so without touching `maxHeapSize`, which is deliberately sized
against the 2 vCPU / 7 GB CI runner and should stay where it is.

#481 names the mechanism exactly: no `forkEvery`, so one JVM serves the whole suite and
Spring keeps a context per distinct test configuration, each with its own Hibernate
`SessionFactory`, alive to the last test. The default cache of 32 means every one of
them is permanent. Four more entities was enough to tip it: the run finished all 646 of
its tests and then died with `OutOfMemoryError` inside ArchUnit's
`ClassFileImportRecord`, which needs the whole application's class graph in memory at
once and runs late, when there is least of it left. That is #481's "it fails somewhere
else" symptom with a new face.

Capping the cache makes Spring close the least recently used context rather than holding
all of them, so the ceiling stops moving with the schema. A context is rebuilt if a
later class needs it again, which is the only cost, and it does not show: measured on
`.116`, 7m18s with the cap against a 7m31s pristine baseline.

## Verification

Full suite on `lab-node-116-f`, against a pristine `origin/development` baseline run on
the same box:

| | result | time |
|---|---|---|
| pristine `origin/development` | BUILD SUCCESSFUL, 104 test classes | 7m31s |
| this branch | BUILD SUCCESSFUL, 538 passed, 0 failures, 0 OOM | 7m18s |

CI on this branch is green as well, including the JaCoCo floor.

- `MeasurementDeviationTest`: plain JUnit, no context.
- `ChecklistTemplateServiceIT`: real CockroachDB. Also proves the Liquibase starter seed
  loaded, which no unit test can.
- **No new Spring context.** `ChecklistTemplateServiceIT`, `InspectionServiceIT` and
  `InspectionTaxonomyMigrationIT` carry the same annotations and the same `@Import` list
  to the letter, so the cache serves all three from one.

## Liquibase

`057-create-and-seed-checklist-templates.xml` (four tables plus the seed) and
`058-extend-inspection-check-items.xml` (four columns plus an index). Both registered in
`db.changelog-master.xml`. 056 went to the project-location work while this was in
flight, hence 057. Every precondition names **every** object its changeset creates: the
four-column `addColumn` is guarded on all four, not just the first.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,6 +107,16 @@ tasks.named('test') {
     // and the OOM killer then takes the database, failing the rest of the suite.
     maxHeapSize = '1024m'
     jvmArgs '-XX:MaxMetaspaceSize=512m'
+    // Bound how many Spring test contexts the run keeps alive at once. There is no
+    // forkEvery, so one JVM serves the whole suite and the default cache of 32 means
+    // every distinct test configuration's context, each with its own Hibernate
+    // SessionFactory, is held to the last test. That total only grows as the schema
+    // does, and it is what leaves nothing for the ArchUnit import that runs late in
+    // the run and needs the class graph of the whole application in memory at once.
+    // Capping the cache makes Spring close the least recently used context instead of
+    // holding all of them; a context is rebuilt if a later class needs it again,
+    // which costs seconds where the alternative costs the build.
+    systemProperty 'spring.test.context.cache.maxSize', '8'
     // Print each test as it starts and finishes so a CI log always shows which test
     // is running; without this, a hung test leaves no trace of which one stalled.
     testLogging {
--- a/src/main/java/org/tornotron/echno_backend/inspection/domain/ChecklistTemplate.java
+++ b/src/main/java/org/tornotron/echno_backend/inspection/domain/ChecklistTemplate.java
@@ -1,0 +1,94 @@
+package org.tornotron.echno_backend.inspection.domain;
+
+import jakarta.persistence.*;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.Filter;
+import org.hibernate.annotations.UpdateTimestamp;
+import org.tornotron.echno_backend.common.multitenancy.TenantScopedEntity;
+import org.tornotron.echno_backend.inspection.InspectionTrade;
+import org.tornotron.echno_backend.organization.Organization;
+
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * A reusable checklist an organization's QA engineer defines once for a trade and
+ * an inspection of that trade is then created from. Tenant-scoped through the
+ * Hibernate {@code orgFilter} exactly as {@link Inspection} is: the criteria a
+ * client accepts work against are their own, so a template belongs to one org and
+ * is invisible to every other.
+ *
+ * <p>One template per trade per org, enforced by the unique constraint on
+ * ({@code organization_id}, {@code trade}). {@code version} is therefore a
+ * revision counter on that one row rather than a key: it starts at 1 and the
+ * service bumps it on every edit, so an inspection's copied items can be traced
+ * back to the revision they came from. {@code active} decides whether the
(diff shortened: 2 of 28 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #499, opened 2026-08-27
Split the inspection module's authority along the jobs that do the work

abhijithanandan:
Phase 1 of the inspection module, part 3 of 3. Implements section 3.5 of
`docs/specs/2026-08-26-inspection-module.md`.

**Merges last, after #498.**

## What this changes

Every endpoint in the module admitted `system-admin` and `project-manager` and nobody
else. On the NCR that made part 2's closure trail meaningless: whoever could raise a
non-conformance could also close it.

The policy now lives in one `inspectionSecurity` bean, following the existing
`attendanceSecurity` precedent, with each role set overridable per environment:

| Act | Default roles |
|---|---|
| Read inspections, checklists, NCRs | `system-admin, project-manager, qa-engineer, safety-officer, site-engineer` |
| Schedule and record an inspection | `system-admin, project-manager, qa-engineer, safety-officer` |
| Define a checklist template | `system-admin, qa-engineer` |
| Raise and assign an NCR | `system-admin, qa-engineer, safety-officer` |
| Report the corrective action done | `system-admin, site-engineer` |
| Verify / reject / reopen / close a **quality** NCR | `system-admin, qa-engineer` |
| Verify / reject / reopen / close a **safety** NCR | `system-admin, safety-officer` |

## Decisions where the spec was silent

- **The three new roles are org-scoped `OrgRole` constants, not realm occupation roles.**
  The realm already carries `site-engineer` and `safety-officer` as occupation roles, but
  as `docs/specs/2026-08-25-rbac-composite-job-roles.md` records, occupation roles gate
  nothing in application code: `@PreAuthorize` checks the `/org-{id}/{role}` subgroup
  layer. Putting the FRS roles anywhere else would not enforce anything. There is no
  `qa-engineer` in either layer today, so it is new.
- **Existing organizations need no backfill.** Subgroups are created from `OrgRole` when
  an organization is created, so every existing tenant lacks the three new ones.
  `assignOrgRole` now creates a missing subgroup instead of refusing, so the first
  assignment provisions it. The name can only come from the enum, so this cannot invent
  an authority. Without it, adding a role would need a Keycloak migration over every
  tenant before the module could be used.
- **They are not manager roles.** `OrgRole.getManagerRoles()` decides who may be named
  the manager on a project invite code; a QA engineer signs off quality, not headcount.
- **The project manager reads NCRs but does not raise or assign them.** This is a
  deliberate reading of the spec's "full read and approval visibility" line, not an
  omission, and it is the reversible direction. If a PM turns out to need to raise or
  assign, that is one config token; the opposite mistake leaves someone with broad
  authority quietly closing quality reports the spec means a QA engineer to own.
- **The safety officer is not on the checklist-write list.** A template is keyed by
  `InspectionTrade`, which is the QA/QC axis; safety criteria genuinely have no home
  until Phase 2, and granting authority over something that does not exist is not a
  grant. Confirmed as the intended reading rather than left open.
- **The site engineer cannot schedule or record an inspection.** The spec gives them
  NCRs; they act on what an inspection raises rather than carrying it out.
- **The sign-off check is per report, not per endpoint**, because the discipline is a
  property of the report rather than of the request. A report that is not in the tenant is
  allowed through the guard so the caller gets the service's 404 rather than a 403 that
  would read as "it exists and you may not touch it".

## Activation

Nothing blocks the merge. The roles are inert until someone holds them and `system-admin`
covers every act meanwhile, so no tenant is locked out the moment this deploys. To put
the split live on `echno.in`, assign `qa-engineer` / `safety-officer` / `site-engineer`
to the relevant employees through the existing org-role endpoint; the subgroups appear on
first assignment.

## No bot review ran on this one

CodeRabbit's check is green, but its description reads `Review rate limited`, which means
no review happened rather than that it found nothing. I re-requested once and got the
same answer. Worth knowing before reading the green tick as a second opinion: #497 was
never reviewed either, and #498 is the only one of the three that got a real pass.

In its place, the guard rewiring was audited by pairing every request mapping in the
three controllers against the expression that guards it, since that edit was applied
mechanically and a read guard landing on a write endpoint is exactly the mistake it could
have made. All 19 endpoints carry the intended guard: reads on `canRead()`, each write on
the act it belongs to, and the four sign-off actions on the per-report check. Two things
deliberately left alone and confirmed still alone: `removeOrgRole` looks the subgroup up
and fails rather than creating one, because removing a role from a group that does not
exist is a real error, and `OrgRole.MANAGER_ROLES` is unchanged, so project invite codes
and the employee hierarchy do not start seeing the three new roles as manager roles.

## Verification

Full suite on `lab-node-116-f` at the tip of the stack, against a pristine
`origin/development` baseline on the same box:

| | result | time |
|---|---|---|
| pristine `origin/development` | BUILD SUCCESSFUL, 104 test classes | 7m31s |
| this branch, rebased on the merged #497 | BUILD SUCCESSFUL, 673 passed, 0 failures, 0 OOM | 7m41s |

- `InspectionSecurityServiceTest`: plain Mockito, no context. Covers both discipline
  directions and the missing-report case.
- The `EndpointAuthorizationTest` ratchet still passes: every endpoint carries a guard.
- The Keycloak subgroup path has no automated cover: it drives the Keycloak admin client
  and there is no Keycloak in the suite. Read it as read, and it only ever creates a name
  that comes from the `OrgRole` enum.

> Based on `feat/inspection-ncr-workflow` (#498), rebased onto the merged #497. Retarget
> to `development` once #498 lands and CI and CodeRabbit will run on it. Reports
> (spec 3.6) are the remaining Phase 1 item and are filed as #502.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/enums/OrgRole.java
+++ b/src/main/java/org/tornotron/echno_backend/common/enums/OrgRole.java
@@ -15,12 +15,27 @@ import java.util.Set;
  *   @PreAuthorize("@orgSecurity.hasOrgRole(#orgId, 'system-admin')")
  *
  * The groupName field is the actual Keycloak subgroup name (kebab-case).
+ *
+ * <p>Adding a constant here adds the subgroup to every organization created from
+ * then on. Organizations that already exist get theirs the first time the role is
+ * assigned: {@code KeycloakGroupService.assignOrgRole} creates a missing subgroup
+ * rather than refusing, so no backfill over existing tenants is needed.
  */
 public enum OrgRole {
     SYSTEM_ADMIN("system-admin"),
     ORG_MANAGER("org-manager"),
     HR_ADMIN("hr-admin"),
-    PROJECT_MANAGER("project-manager");
+    PROJECT_MANAGER("project-manager"),
+    // The three inspection roles. They are org-scoped rather than realm occupation
+    // roles because this is the layer @PreAuthorize actually checks: the realm has a
+    // 'site-engineer' and a 'safety-officer' occupation role already, but occupation
+    // roles gate nothing in application code and are a catalogue for assignment, not
+    // an authority. They are deliberately not manager roles: a QA engineer signs off
+    // quality, not headcount, and getManagerRoles decides who may be named the
+    // manager on a project invite code.
+    QA_ENGINEER("qa-engineer"),
+    SAFETY_OFFICER("safety-officer"),
+    SITE_ENGINEER("site-engineer");
 
     private final String groupName;
 
--- a/src/main/java/org/tornotron/echno_backend/common/service/KeycloakGroupService.java
+++ b/src/main/java/org/tornotron/echno_backend/common/service/KeycloakGroupService.java
@@ -176,8 +176,15 @@ public class KeycloakGroupService {
      * groups claim, which JwtAuthConverter converts to authority "ORG_5_ROLE_system-admin".
      *
      * NOTE: The user must ALSO be a member of the org (via addUserToOrganization).
-     * Role assignment and membership are separate — a user can be a member without
+     * Role assignment and membership are separate: a user can be a member without
      * any role, but should not have a role without being a member.
+     *
+     * A subgroup that does not exist yet is created here rather than refused. The
+     * subgroups are created from OrgRole when an organization is created, so an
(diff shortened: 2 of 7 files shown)


abhijithanandan:
@coderabbitai review

PR #504, opened 2026-08-27
Stop the web project listing at ten rows without saying so

### LINKED ISSUE #503: GET /project/web silently returns only the first 10 projects
`GET /api/v1/project/web` paginates by default and returns a bare array, so every caller silently gets the first 10 projects and nothing says there are more.

Found while fixing #478, and it shares the cause exactly. That issue was the same defect on `/tasks/web`; this is the same code shape on projects, and the two want fixing the same way so the endpoints stay consistent with each other.

## The call chain

`ProjectControllerWeb.java:100-105`:

```java
public ResponseEntity<List<ProjectDto>> readAllProjects(@RequestParam(defaultValue = "0") int pageNo,
                                                        @RequestParam(defaultValue = "10") int pageSize) {
    Page<ProjectDto> projects = service.getAllProjects(pageNo,pageSize);
    return new ResponseEntity<>(projects.getContent(),HttpStatus.OK);
}
```

Paginated, defaulting to `pageSize=10`, then the `Page` wrapper is thrown away with `.getContent()`. The caller receives a plain JSON array with no `totalElements`, no `totalPages`, and no way to tell a complete result from a truncated one. `ProjectService.getAllProjects` sorts `id ASC`, so it is the ten lowest ids in the tenant.

`echno-core` calls it with no query parameters at all, so it always takes the defaults:

`src/services/project-service.ts:94`

```ts
async getAll(): Promise<Project[]> {
  const data = await api.get<ApiResponse[]>('/project/web');
  return safeParseProjects(data);
}
```

That feeds `useProjects()`.

## Impact, which is wider than the task version

`useProjects()` has **20+ consumers in echno-web**, and most of them are project pickers in forms rather than list pages:

- site transfer, purchase order, material consumption, GRN, indent, material, storage location, labour, inspection and mark-attendance forms
- purchase order and indent info cards, storage location stock tab
- chat room creation, team attendance history, the finance payments page, the dashboard, the command palette

So on any tenant with more than ten projects, **a user cannot select their project in any of those forms** unless it happens to be among the ten lowest ids. The dropdown looks complete and simply does not contain the project. That is worse than a truncated list page, because the failure looks like the project does not exist.

It also compounds with tornotron/echno-web#269: a truncated project list caps what the command palette can find by name, on top of the palette's own client-side slice.

## Suggested fix

The same shape #478 used on tasks, so the two endpoints match:

- drop the paging parameters from the bare listing and bound it with `UnpagedResultCap`, so the response stays a bare array but carries `X-Total-Count` always and `X-Result-Capped` when rows were left out;
- add `GET /project/web/paginated` returning a real `Page<ProjectDto>` for the honest-paging path.

No core or web change is needed for the pickers: they get the fix through the response, since the array shape does not change.

abhijithanandan:
Closes #503. Backend-only, no core or web change needed.

## The defect

`GET /api/v1/project/web` took `pageNo`/`pageSize` defaulting to ten, then answered with `page.getContent()`. Bare array, no total, no way to tell a complete result from a truncated one. echno-core's `projectService.getAll()` passes no parameters, so it always took the defaults.

Same defect as #478 on `/tasks/web`, found while fixing that one, and it lands worse. Most of the 20-odd `useProjects()` callers are **project pickers in forms**, not list pages: site transfers, purchase orders, material consumption, goods receipts, indents, materials, storage locations, labour, inspections, marking attendance, chat room creation, finance payments. A user whose project was not among the ten lowest ids could not select it, and the dropdown looked complete while doing it, so the failure read as the project not existing rather than as a bug.

It also compounds with tornotron/echno-web#269: a truncated project list caps what the command palette can find by name, on top of the palette's own client-side slice.

## What changed

Deliberately **the same shape #494 used on tasks**, because consistency between the two listings matters more than either choice would in isolation:

- the paging parameters are gone from the bare listing rather than re-tuned, since keeping them on an endpoint that discards the envelope is what made the truncation silent;
- the read is bounded by `UnpagedResultCap`, so the response stays a bare array and carries `X-Total-Count` always plus `X-Result-Capped` when rows were left out;
- `GET /project/web/paginated` is new and returns a real `Page<ProjectDto>`, optionally filtered by a free-text search on the name, with page size clamped to the same cap and LIKE wildcards escaped.

Because the bare array shape is unchanged, **every picker is fixed by this PR alone** with no core release and no web change.

No new unbounded read: the added repository query takes a `Pageable`, and the ArchUnit ratchet is untouched.

## Verification

Run on lab `.116` (Testcontainers needs Docker):

```
ProjectControllerWebListingTest    7 PASSED
ProjectServicePaginationTest       8 PASSED
ProjectRepositoryIT                3 PASSED
ProjectControllerWebAuthzTest      3 PASSED  (existing, unaffected)
ProjectCreationValidationTest      5 PASSED  (existing, unaffected)
EndpointAuthorizationTest          PASSED
UnboundedRepositoryReadTest        3 PASSED
BUILD SUCCESSFUL
```

Against the pre-fix controller body, five of the seven listing tests fail:

```
theBareListingDoesNotStopAtTheFirstTenProjects()            FAILED
theBareListingReachesTheProjectsAPickerWouldOtherwiseNeverOffer() FAILED
theBareListingAsksForTheResultCapRatherThanAPageOfTen()     FAILED
theBareListingReportsTheTrueTotalOnEveryResponse()          FAILED
theBareListingSaysSoWhenRowsWereLeftOut()                   FAILED
```

The two `/paginated` tests pass in that run only because the endpoint they exercise is still present; without this change it does not exist.

The stub honours the requested page size on purpose. A stub that ignores it passes against the truncating controller too, which is precisely the blind spot that let the original defect ship with a green suite.

`ProjectRepositoryIT` is new; there was none before. The paginated query is JPQL, so a wrong field name or a mishandled `ESCAPE` clause is invisible until it runs against a real database, which the mocked service test cannot catch by construction.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/project/ProjectControllerWeb.java
+++ b/src/main/java/org/tornotron/echno_backend/project/ProjectControllerWeb.java
@@ -17,6 +17,7 @@ import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.multipart.MultipartFile;
+import org.tornotron.echno_backend.common.pagination.UnpagedResultCap;
 import org.tornotron.echno_backend.common.response.ApiResponse;
 import org.tornotron.echno_backend.employee.dto.EmployeeDto;
 import org.tornotron.echno_backend.project.dto.ProjectCreationDto;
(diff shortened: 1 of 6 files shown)


abhijithanandan:
@coderabbitai review

Issue #510, opened 2026-08-28
Concurrent compliance generation creates duplicate inspections: the idempotency check is an unlocked read with no unique constraint behind it


abhijithanandan:
Compliance generation is idempotent by intent: a `(projectId, complianceRuleRef)` pair that already has an inspection is skipped, so a re-run only adds what is missing. The check is an unlocked read, so it does not hold under concurrency and two overlapping runs create duplicate compliance inspections for the same rule.

## The race

In `ComplianceGenerationService.generateForProject`:

```java
if (inspectionRepository.existsByProjectIdAndComplianceRuleRefAndOrganization_Id(
        projectId, rule.getCode(), orgId)) {
    log.debug("Compliance {} already exists for project {}; skipping", rule.getCode(), projectId);
    continue;
}
...
Inspection saved = inspectionRepository.save(inspection);
```

Nothing between the `exists` and the `save` prevents another transaction doing the same thing. There is no unique constraint on `(organization_id, project_id, compliance_rule_ref)` and no lock, so the check is advisory only.

Two runs overlap easily:

- **Two clicks on the regenerate button.** Each takes 34 to 47 s today (see #482), so the window is not a narrow one, it is most of a minute. A user who thinks the first click did not register gets duplicates.
- **Approval plus a manual regenerate.** `ComplianceGenerationListener` fires generation asynchronously on project approval. A project manager who opens the project and clicks regenerate while that is still running gets the same overlap.
- **Two users.** The endpoint is open to `system-admin` and `project-manager`, so two people looking at the same project can each start a run.

CockroachDB runs `SERIALIZABLE`, which is worth being precise about: it aborts one side of a genuine read-write conflict, but these two transactions do not conflict. They both read "no such row" and then insert *different* rows, because `Inspection` ids and `inspectionNumber` values come from a sequence. There is nothing for the database to detect. Serializable does not save this one.

The result is two `SUGGESTED`, `AI_GENERATED` compliance inspections with the same `complianceRuleRef` on the same project, each with its own `INSP` document number, both shown to the user, both needing to be resolved separately.

## Why it is being filed now

The #482 design proposes a partial unique index on the new job table:

```
UNIQUE (organization_id, project_id) WHERE status IN ('pending','running')
```

so a second generation request joins the running job instead of starting a second one. That closes this race as a side effect, by making concurrent runs impossible rather than by making them safe.

[...]

Issue #509, opened 2026-08-28
Compliance generation reports success with zero results once the rule catalogue truncates the model response


abhijithanandan:
Compliance generation asks the model for one JSON element per candidate rule. Past a certain rule count the response is truncated by the token cap, the parser throws on the truncated JSON, the exception is swallowed, and generation reports **success with zero results**. The user is told nothing is wrong and no compliance inspections are created.

## The chain

**1. The cap.** `compliance.ai.max-tokens` is 4096 (`ComplianceAiProperties`, `application.yml`). The prompt in `OpenAiCompatibleComplianceService.buildUserPrompt` closes with "Include one element for every candidate rule", and each element carries `ruleCode`, `applies`, `riskLevel`, `resolutionOptions` (an array of short strings), `rationale` and `phase`. At a realistic 80 to 150 output tokens per element that caps out somewhere around 30 to 35 rules.

**2. The truncation is not detected.** `extractText` reads `choices[0].message.content` and never looks at `choices[0].finish_reason`, so a `length` stop is indistinguishable from a `stop`.

**3. The parser turns it into a throw.** `parseSuggestions` slices from the first `[` to the **last** `]`:

```java
int start = responseText.indexOf('[');
int end = responseText.lastIndexOf(']');
```

In a truncated response the last `]` is not the array's closing bracket, it is the closing bracket of some element's inner `resolutionOptions` array. The slice is therefore a fragment ending mid-object, and Jackson throws.

**4. The throw is swallowed.** `suggestCompliances` wraps the whole call in `catch (Exception e)`, logs at error, and returns `List.of()`.

**5. Empty is read as "nothing applies".** `ComplianceGenerationService.generateForProject` treats an empty list as a legitimate outcome and returns an empty result, logging "Compliance AI returned no suggestions". The endpoint answers 200 with `[]`.

So the fail-soft design, which is correct for "the AI key is not set" and "the endpoint is down while a project is being approved", also silently absorbs a structural failure that will happen on every single run once the catalogue crosses the threshold.

## The part worth stating plainly

There are **two independent ceilings and they converge on almost the same rule count**.

- The 60 s edge timeout (`proxy_read_timeout` in the deployment edge role) is crossed at roughly 20 to 25 rules, at the measured 2.5 s per rule. That is #482.
- This token ceiling is crossed at roughly 30 to 35 rules.

[...]

Issue #507, opened 2026-08-28
Tenant isolation fails open on a null TenantContext: both the orgFilter aspect and the load listener give up silently


abhijithanandan:
Tenant isolation is enforced by two mechanisms, and both of them **fail open** when `TenantContext` holds no organization id. Any code path that runs without a populated tenant context therefore reads every organization's rows, silently, with no exception and no warning log.

This is not specific to compliance. It is a property of the isolation model, and any background worker, scheduled task, event listener, message consumer or test harness can trip it. It was found while designing #482 but it belongs on its own.

## The two mechanisms

**`HibernateFilterConfig`** is an `@Around` aspect on `@annotation(Transactional)` that enables the Hibernate `orgFilter` on the session:

```java
Long orgId = TenantContext.getCurrentOrgId();
if (orgId != null && !TenantContext.isBypassed()) {
    session.enableFilter("orgFilter").setParameter("organizationId", orgId);
} else {
    log.debug("[HibernateFilter] orgFilter NOT enabled (orgId={}, bypassed={})", orgId, TenantContext.isBypassed());
}
```

A null org id takes the `else` branch. The transaction proceeds with the filter off, and the only trace is a `debug` line that is not enabled in any deployed environment.

**`TenantIsolationLoadListener`** is the fail-closed load-boundary check added in #330. It returns early on the same condition:

```java
Long contextOrgId = TenantContext.getCurrentOrgId();
if (contextOrgId == null || TenantContext.isBypassed()) {
    return;
}
```

So the listener that exists specifically to catch what the filter misses gives up on **exactly the same condition** as the filter does. There is no third line of defence.

That is worth stating plainly, because the two mechanisms read as defence in depth and are not. Defence in depth means two checks that fail independently. These two fail together, on the same input, for the same reason. It is one check written twice.

## Why this is worse than it looks

The two mechanisms were designed to cover each other, and on their intended axis they do: the filter scopes queries but does not apply to primary-key loads, and the load listener covers primary-key loads but not query shape. They are complementary in *what* they check and identical in *when* they give up. So a null tenant context does not degrade isolation, it removes it.

The failure is also invisible. There is no thrown exception, no WARN, no metric. A background job that forgets to set the context returns plausible-looking data drawn from every tenant, and nothing in the logs distinguishes that from a correct run.

[...]

Issue #482, opened 2026-08-27
Compliance generation runs the model inside the request and inside a transaction, so it cannot scale past ~20 rules


abhijithanandan:
Compliance generation calls an external model synchronously inside the request, and the call time grows with the number of curated rules for the jurisdiction. The timeout work done for ClickUp `86d45jr56` bought headroom; it did not change that shape. This issue is for the shape.

## Measured, on the staging model over the same network path

| Rules | Elapsed |
|---|---|
| 6 | 16 s |
| 12 | 27 s |
| 20 | 49 s |

Roughly linear, about 2.5 s per rule. Tamil Nadu residential currently has 6 curated rules, 21 across all jurisdictions, which is why a 47 s worst case was survivable once the client budgets were raised.

nginx at the edge stops waiting at **60 s**, and no request-scoped budget survives that. At roughly 20 to 25 rules for a single jurisdiction the endpoint crosses it and starts failing again, this time with nothing left to tune.

**So this reopens the moment Anand's rule catalogue lands.** That work is expected and is the whole point of the compliance module, so the ceiling is not hypothetical.

## The second defect, which belongs with it

`ComplianceGenerationService.generateForProject` is `@Transactional` across the entire model call, so every click pins a database connection for 34 to 47 seconds. A handful of concurrent users would exhaust the pool while doing nothing but waiting on an external service.

It was deliberately left alone rather than patched, for a good reason: `HibernateFilterConfig` keys the tenant-isolation filter off that same `@Transactional` annotation, so moving the boundary needs care to avoid running the work with tenant filtering off. That is the same trap the serializable-retry work hit. Splitting the transaction and moving to a background job are the same change, and doing them separately means doing them twice.

## What this needs

Generation moves off the request: a job is accepted, the caller polls or is notified, and the transaction wraps only the persistence of the result rather than the model call. The existing compliance flow already materialises `SUGGESTED` inspections asynchronously in spirit, so the UI already has a shape for "results appeared later".

Worth deciding at the same time: whether generation should be triggered per project on demand at all, or scheduled, since a 50 second wait is a poor interaction even when it succeeds.

## What is already done

Client budgets raised so the browser gives up before the BFF and can explain itself, a timeout now shows an honest "still running" message and refreshes the list rather than a red error, and the timeout policy moved somewhere testable. Those stay useful regardless, since they cover the ordinary case. They do not survive the 60 s ceiling.

abhijithanandan:
[...]

PR #514, opened 2026-08-28
Take the compliance model call out of the transaction

### LINKED ISSUE #508: ComplianceGenerationListener enters its transaction with a null org id, so orgFilter is enabled only by ordering accident
`ComplianceGenerationListener.onProjectApproved` runs the compliance generation that fires when a project is approved. It is annotated `@Transactional(propagation = REQUIRES_NEW)` and it sets the tenant context **inside the method body**:

```java
@Async
@Transactional(propagation = Propagation.REQUIRES_NEW)
@TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)
public void onProjectApproved(ProjectApprovedEvent event) {
    Long orgId = event.getOrganizationId();
    ...
    try {
        TenantContext.setCurrentOrgId(orgId);
        complianceGenerationService.generateForProject(projectId, orgId);
```

`EchnoBackendApplication` pins `@EnableTransactionManagement(order = Ordered.HIGHEST_PRECEDENCE)`, and `HibernateFilterConfig` is `@Order(Ordered.LOWEST_PRECEDENCE)`. So the order on entry to this method is:

1. Spring's transaction interceptor opens the `REQUIRES_NEW` transaction.
2. The `HibernateFilterConfig` aspect runs and reads `TenantContext.getCurrentOrgId()`.
3. The method body runs, and only now sets the org id.

The aspect reads the context at step 2, before step 3 has set it. It sees `null`, takes its else branch, and **never enables `orgFilter` for this transaction**.

## Why it appears to work

`generateForProject` is itself `@Transactional`, with the default `REQUIRED` propagation, so it joins the transaction the listener already opened rather than starting its own. Its own `HibernateFilterConfig` advice then fires, and by that point the body has set the org id, so the aspect enables `orgFilter` on the session that is already open.

The filter is therefore switched on as a side effect of an inner call, on a transaction that began without it. It works by ordering accident, not by design.

The accident stops holding if any of these change, none of which look dangerous from the outside:

- `generateForProject` loses its `@Transactional`, which is exactly what the #482 redesign does to it,
- it moves to `REQUIRES_NEW` or is called through a path that does not go via the proxy,
- the advice ordering changes,
- another statement is added to the listener body before `setCurrentOrgId`.

## Why it matters now

Read with #507: both isolation mechanisms fail open on a null tenant context, so a transaction that begins without `orgFilter` is not caught by anything else. Between step 1 and step 3 this transaction has no tenant scoping at all, and if the inner call ever stops enabling it, generation runs entirely unscoped and reads every organization's projects and rules with no error and no log line.

Nothing in the current code exercises that gap, so this is a latent defect rather than an active leak. It is filed as a bug because the thing standing between it and a cross-tenant read is a call-ordering coincidence.

## It also pins a connection for the full model call

Separately, the `@Transactional` on this listener holds a Hikari connection for the entire duration of the method, which includes the synchronous external model call. Measured at 34 to 47 s today, growing at roughly 2.5 s per curated rule (see #482). The pool is 20. The approval path has exactly the same connection-pinning defect as the request path, which is easy to miss because the listener is `@Async` and therefore looks like it is already off the hot path. Being off the request thread does not release the connection.

## Fix

The listener should not own a transaction. It should set the tenant context, delegate, and let the service own its own boundaries.

Concretely, and this is Stage 1 of the #482 plan:

- Drop `@Transactional` from `onProjectApproved` entirely.
- Set and clear the context through a single `TenantScopedJobRunner` bean that refuses a null org id loudly and clears in a `finally`, so the id cannot leak to the next task on the pooled `@Async` thread.
- Split `generateForProject` into load, model call and persist, with each transaction opened through `TransactionalWorkRunner.runInTransaction`. That keeps the boundary declarative, which is what makes `HibernateFilterConfig` fire on it, and it means the connection is released across the model call.

`TransactionalWorkRunner`'s Javadoc already documents why a programmatic `TransactionTemplate` is the wrong instrument here: it carries no annotation, so the aspect never runs and the transaction would execute with tenant filtering off. Same trap, one level up.

## Related

- #507, the isolation model failing open, which is what makes this dangerous rather than merely untidy
- #482, the background-job design where this was found and where the fix is Stage 1

abhijithanandan:
Stage 1 of the #482 plan. Fixes #508, and adds the local containment for #507.

## The problem

`ComplianceGenerationService.generateForProject` was `@Transactional` across the whole external model call. That call is 34 to 47 seconds today and grows at roughly 2.5 s per curated rule, so every click pinned one of twenty pool connections for its duration while doing nothing but waiting on a third party. A handful of concurrent users would exhaust the pool.

Moving that boundary is not a free refactor. `HibernateFilterConfig` advises `@Transactional` methods in this package tree to enable the `orgFilter` tenant filter on the session, so the transaction annotation and the tenant filter are the same annotation. Splitting the boundary carelessly runs the work with tenant filtering off, and because that filter fails open rather than closed, it does so silently.

## What changed

**The service is three phases.** Read the inputs in one transaction, call the model with no transaction and no connection held, persist in a second. Both transactions are opened by calling `TransactionalWorkRunner`, which is declarative and inside the advised package, so the aspect fires on each phase. A programmatic `TransactionTemplate` would have split the boundary correctly and run both phases unfiltered.

The read phase hands the later phases detached entities. Every field they touch is a basic column loaded with the row, and the record carrying them says so explicitly, because that is the invariant an edit could break without a test noticing.

**The listener owns no transaction.** `ComplianceGenerationListener` had the same connection pin plus one of its own. It was `@Transactional(REQUIRES_NEW)` and set the tenant context inside the method body, but the transaction interceptor is pinned at `HIGHEST_PRECEDENCE` and `HibernateFilterConfig` sits at `LOWEST_PRECEDENCE`, so the filter aspect read the context one step before the body set it, saw null, and never enabled `orgFilter` for that transaction. The filter came on later only as a side effect of the inner service call firing the same aspect on the already-open session. It worked by ordering accident, and the phase split above is exactly the kind of change that would have stopped the accident holding.

Being `@Async` was not protection: it moved the wait off the request thread and did nothing about the connection.

**New `TenantScopedJobRunner`.** Both isolation mechanisms fail open on a missing tenant context: `HibernateFilterConfig` logs at debug and skips, and `TenantIsolationLoadListener` returns early on the same condition. So background work that forgets to establish the tenant reads every organization's rows and looks healthy doing it. The runner refuses a null organization id instead, forces the bypass flag off for the duration, and restores the previous context rather than blind-clearing, so it is safe on a request thread as well as a pooled one. On a pooled thread restoring is clearing, which is what stops an org id leaking into the executor's next task.

That is the containment for the work we own. The wider property is #507 and is not addressed here.

## What this does not do

**The endpoint still blocks, and still fails at the 60 second edge timeout.** Nothing here changes what the caller experiences. "Compliance transaction boundary fixed" could reasonably be read as the timeout going with it, and it did not: the model call is still synchronous inside the request, and it still crosses `proxy_read_timeout` at roughly 20 to 25 rules for one jurisdiction. That ceiling is #482, and the job table, dispatcher and polling endpoint that answer it are deliberately not in this PR.

No API change, no schema change, no `echno-core` or `echno-web` change, so this ships on its own.

## Tests

- `TenantScopedJobRunnerTest`, 7 cases: null refusal, tenant visible during the work, context cleared afterwards, cleared on throw, a pre-existing context restored rather than discarded, bypass forced off and restored.
- `ComplianceGenerationListenerTest`, 4 cases, including a reflection guard asserting `onProjectApproved` carries no `@Transactional` with the reason in the failure message. Same guard on `generateForProject`. Reintroducing either annotation now fails a test rather than quietly restoring the accident.
- `ComplianceGenerationServiceIT` is unchanged and passes against a real CockroachDB, which is the check that mattered: the phase split still produces the same rows in the same lifecycle order and is still idempotent on a second run.

Full affected scope green on the lab build box after rebasing onto current `development`: compliance, multitenancy, events and retry, including `TenantIsolationIT` and `TransactionRetryTemplateIT`.

## Related

- #482 the design this is Stage 1 of, and where the remaining stages are set out
- #507 tenant isolation failing open, which this contains locally but does not fix
- #508 the ordering accident, fixed here
- #509 the token ceiling, which asynchrony does not fix and this does not touch
- #510 the duplicate-inspection race, untouched here

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compliance generation reliability during project approval.
  * Prevented tenant context from leaking between background tasks.
  * Invalid compliance-generation requests no longer interrupt the approval process.
  * Improved handling of projects missing required tenant information.

* **Performance**
  * Compliance generation now avoids holding database resources during AI processing, improving scalability and responsiveness.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/events/listeners/ComplianceGenerationListener.java
+++ b/src/main/java/org/tornotron/echno_backend/common/events/listeners/ComplianceGenerationListener.java
@@ -4,22 +4,41 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Component;
-import org.springframework.transaction.annotation.Propagation;
-import org.springframework.transaction.annotation.Transactional;
 import org.springframework.transaction.event.TransactionPhase;
 import org.springframework.transaction.event.TransactionalEventListener;
 import org.tornotron.echno_backend.common.events.ProjectApprovedEvent;
 import org.tornotron.echno_backend.common.exception.InvalidRequestException;
 import org.tornotron.echno_backend.common.exception.ResourceNotFoundException;
-import org.tornotron.echno_backend.common.multitenancy.TenantContext;
+import org.tornotron.echno_backend.common.multitenancy.TenantScopedJobRunner;
 import org.tornotron.echno_backend.compliance.ComplianceGenerationService;
 
 /**
  * Runs AI compliance generation when a project is approved. It fires only AFTER the
  * approving transaction commits (so the project row is durable) and on a separate
  * async thread (so the slow AI call never blocks the approving request). That async
- * thread has no {@link TenantContext}, so the organization id is read from the event
- * and set on the thread for the duration of the tenant-scoped work, then cleared.
+ * thread has no tenant context, so the organization id is read from the event and
+ * established for the duration of the work by {@link TenantScopedJobRunner}.
+ *
+ * <h2>Why this listener owns no transaction</h2>
+ *
+ * <p>It used to be {@code @Transactional(REQUIRES_NEW)} and set the tenant context inside
+ * the method body. That was wrong in two ways.
+ *
+ * <p>The transaction interceptor is pinned at {@code HIGHEST_PRECEDENCE} and
+ * {@code HibernateFilterConfig} sits at {@code LOWEST_PRECEDENCE}, so the order on entry
+ * was: open the transaction, run the filter aspect, run the body. The aspect read the
+ * tenant context one step before the body set it, saw null, and never enabled
+ * {@code orgFilter} for that transaction. The filter came on later only as a side effect
+ * of the inner service call firing the same aspect on the already-open session, which is
+ * an ordering accident rather than a design, and one that both isolation mechanisms would
+ * have failed open on had it stopped holding.
+ *
+ * <p>The transaction also spanned the external model call, so the approval path pinned a
+ * pool connection for its full 34 to 47 seconds. Being {@code @Async} moved the wait off
+ * the request thread; it did nothing about the connection.
+ *
+ * <p>So the listener now does what a listener should: establish the tenant, delegate, and
+ * let {@link ComplianceGenerationService} own its own transaction boundaries.
  */
 @Slf4j
 @Component
@@ -27,9 +46,9 @@ import org.tornotron.echno_backend.compliance.ComplianceGenerationService;
 public class ComplianceGenerationListener {
 
     private final ComplianceGenerationService complianceGenerationService;
+    private final TenantScopedJobRunner tenantScopedJobRunner;
 
     @Async
-    @Transactional(propagation = Propagation.REQUIRES_NEW)
     @TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)
     public void onProjectApproved(ProjectApprovedEvent event) {
         Long projectId = event.getProjectId();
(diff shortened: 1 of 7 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

Issue #485, opened 2026-08-27
SubscriptionCache holds entities, and caching DTOs instead would freeze the computed entitlement fields


abhijithanandan:
`SubscriptionCache` holds `Subscription` **entities** rather than DTOs. That was raised in #476 as a lazy-initialization hazard, since a cached entity outlives its persistence context by definition and is safe today only because the query behind it fetch-joins the whole graph. Converting it to cache DTOs was left undone, and the reasons found while looking are more interesting than the original concern.

Three things ripple, all verified in the code rather than assumed.

## 1. It would freeze fields that are computed fresh today

`SubscriptionDto.active`, `inTrial` and `expired` are derived from `Instant.now()` inside `toSubscriptionDto`, so they are recomputed on every call while the cache holds only the entity. Cached as a DTO they would be pinned for the full 5 minute TTL.

A subscription that lapses inside that window would keep reporting `expired = false`. Those are precisely the entitlement fields the paywall and payment work will read, so this is the reason to deal with it **before** that work rather than after.

## 2. It needs a new field on a public response DTO

Quota checks key on `UsageRecord.featureId`, a plain `Long` column rather than a relation, and no DTO currently carries a feature id. Caching DTOs means adding one to `PlanFeatureDto`, which then surfaces on every plan and subscription response. That is a public contract change, small but not free, and it wants a core release to reach the frontend.

## 3. Writes mutate the cached instance

`changeSubscription` and `cancelSubscription` currently modify the cached entity and save it. With DTOs they would have to load fresh first. This one is a latent hazard on its own terms, independent of the caching question: a write path that mutates a shared cached object is a bug waiting for a second caller.

## Suggested order

Point 3 is worth fixing on its own merits whatever is decided about the caching. Points 1 and 2 are the actual design question, and the answer probably wants settling alongside the payment gateway work rather than in isolation, since that work defines what the entitlement reads need to be correct about.

Until then the current arrangement is safe, and the loader's Javadoc says why, so nobody weakens the fetch-join query without noticing.

abhijithanandan:
Point 3 is done and merged (#515, `83fd134`). The write paths no longer touch the cached instance, and the eviction now also runs on transaction completion, because evicting at the point of the write left a window in which a concurrent read re-cached the pre-change row for the rest of the TTL. Leaving points 1 and 2 open here, with a recommendation rather than a change.

## Recommendation: leave the cache holding entities

[...]

PR #515, opened 2026-08-28
Keep the subscription writes off the instance the cache is holding

abhijithanandan:
Closes the part of #485 that is a defect rather than a design question, and answers the design question rather than acting on it unilaterally.

## What was wrong

`changeSubscription` and `cancelSubscription` resolved their subscription through `loadActiveSubscription`, which returns the very instance the cache is holding, and then mutated it. That instance is detached and shared by every concurrent reader of the user, so:

- the change was visible to every reader the moment it was made, before the transaction committed;
- if the write then failed, the eviction after it never ran and the cache went on serving state that had never reached the database.

`createSubscription` decided whether a user already had a subscription from the same shared instance, so a stale entry could refuse a legitimate subscribe.

All three now read the row for themselves through `loadActiveSubscriptionForWrite`. Each gets a managed entity of its own, the change stays inside its transaction, and it reaches other callers only through the eviction.

## The eviction had the same shape of hole

Evicting at the point of the write leaves a window: the entry is gone but the change has not committed, so a read landing in that window misses the cache, re-reads the pre-change row and caches it for the rest of the five minute TTL. A user who had just paid to change plan would keep the old entitlement for five minutes. `evictOnWrite` evicts immediately and again on transaction completion, which also clears whatever a read parked there if the write rolled back.

## The fetch-join dependency, and a correction to the issue

The issue and the loader's Javadoc both said the entity cache is safe today because `findActiveSubscriptionByUserId` fetch-joins the whole graph. That is only half of it, and the half that is not load-bearing on its own. `theQueryBehindTheCacheReturnsAGraphThatOutlivesItsSession` was written to fail if the fetch joins were weakened; with them stripped out, the rest of the suite went on passing, because every current caller of the loader also maps or walks the whole graph inside its own transaction. So the query alone can be weakened today without anything breaking.

Both halves are now pinned, and the Javadoc says which is which rather than crediting the fetch joins with all of it.

## Recommendation on caching DTOs

Leave the cache holding entities.

- `SubscriptionDto.expired` and `inTrial` derive from `Instant.now()` over immutable timestamps, so the entity cache evaluates them correctly at read time and a DTO cache would pin them for the TTL. Those are the entitlement fields the paywall reads. Caching the public DTO makes them wrong.
- The issue lists `active` with those two. It is not time-derived at all: `Subscription.isActive()` reads only `status`, which the entity cache already freezes. So `active` is equally stale either way, bounded by eviction on write and the TTL. Nothing in the codebase changes a subscription's status outside `SubscriptionService`, so the bound holds (`findExpiredSubscriptions` has no callers).
- The feature-id problem is real but not the constraint the issue takes it for. It only forces a change to `PlanFeatureDto` if the cache has to hold the public DTO, which is the response contract and not the cache's business. An internal projection carrying the feature id would need no contract change and no core release.
- That projection is the honest alternative, and its only payoff is removing a hazard the new tests now catch. The payment gateway work will change what the cache must hold anyway (external subscription ids, provider status, grace periods), so building it now means building it twice.

Cost of leaving it: the query must keep its fetch joins, which is now enforced by a test rather than a comment.

## Verification

Every new test confirmed failing without its fix, on `lab-node-116-f`.

With `SubscriptionService` and `SubscriptionCache` reverted to `development`:

```
changeSubscription_leavesTheCachedInstanceAlone()            FAILED
cancelSubscription_leavesTheCachedInstanceAlone()            FAILED
changeSubscription_evictsAgainOnceItsTransactionHasCompleted() FAILED
```

With the fetch joins stripped from `findActiveSubscriptionByUserId`:

```
theQueryBehindTheCacheReturnsAGraphThatOutlivesItsSession()  FAILED
    org.hibernate.LazyInitializationException
```

Full `./gradlew build`: BUILD SUCCESSFUL.

No new Spring context: the four tests reuse `BillingMappingIT`, which already runs `@Transactional(propagation = NOT_SUPPORTED)` so the mapping happens with the persistence context closed, which is the point. No entities added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subscription cache invalidation during subscription creation, plan changes, and cancellations.
  * Prevented stale subscription data from being retained after transactions complete.
  * Ensured subscription updates and cancellations do not unexpectedly mutate cached data.

* **Tests**
  * Added coverage for cache consistency, session-independent subscription mapping, and post-transaction eviction behavior.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/billing/components/SubscriptionCache.java
+++ b/src/main/java/org/tornotron/echno_backend/billing/components/SubscriptionCache.java
@@ -7,6 +7,8 @@ import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.tornotron.echno_backend.billing.Subscription;
 
 import java.time.Duration;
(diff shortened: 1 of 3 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #517, opened 2026-08-28
Refuse to approve a project whose state nothing can determine

### LINKED ISSUE #493: Require a project's state at approval, not at creation
Follow-up from #488, which added `project_state` as an optional field. Raised rather than built, because it changes the approval path rather than the form and deserves its own review.

## The gap

Compliance generation is keyed by `(state, projectType)`. #488 made the state an explicit field, but left it optional so nothing existing breaks. That means a project can still reach approval with no state set, and generation then fails at exactly the moment it is supposed to run:

> This project has no state set, and its address does not name one, so the applicable regulations cannot be determined.

The user finds out about a missing field at the point the system needed it, not at the point they could have supplied it.

## Proposal

**Optional at creation, required at approval.** Approval is what publishes `ProjectApprovedEvent`, which is what triggers `ComplianceGenerationService`, so approval is the moment the state genuinely becomes necessary. Requiring it earlier would block the ordinary case of a project being drafted before its site paperwork is settled; requiring it at approval catches every project that will actually need it, and catches it while the user is still on the screen that can fix it.

Concretely, in `ProjectService.partialUpdateAProject`, on the transition into `approved` (the same guarded transition that fires the event today):

```java
if (newStatus == ProjectCreationStatus.approved && previousStatus != ProjectCreationStatus.approved) {
    // reject when neither the state field nor the address yields one
}
```

The address-scan fallback should still count, so the projects that predate the field and have a state in their address line keep approving unchanged. Only a project where *both* routes come up empty is refused, with a message naming the field.

Worth checking before building: whether any existing project would be blocked by this. A count of projects where `project_state IS NULL` and `IndianStateResolver.resolve(project_address)` returns null tells you the size of the backfill, if any.

## Also open, and not a developer decision

**Should typing an address auto-fill the coordinates?** This came out of the same ticket (ClickUp 86d45jbvr). It is feasible, but geocoding needs an external provider, an account and a per-lookup cost, so it is Abhijith's call rather than something to pick up here. Noting it so it is not lost. It is independent of the above: the address, the state and the coordinates are three separate inputs today, and nothing is blocked on it.

## Context

- Field added: #488
- Form: tornotron/echno-web#286
- Shared types: tornotron/echno-core#40 (published 1.15.0)

abhijithanandan:
Closes #493.

Optional at creation, required at approval, as decided on the issue.

## Why approval

Compliance generation is keyed by `(state, projectType)`. #488 made the state an explicit field but left it optional, so a project could reach approval without one. Approval publishes `ProjectApprovedEvent`, the listener runs `AFTER_COMMIT`, and generation then failed there with "This project has no state set". The approval itself succeeded, so the user met the missing field as an absence of inspections, on a screen they had already left.

Requiring it at creation would block the ordinary case of a project drafted before its site paperwork is settled. Approval is the first moment the state is genuinely needed and the last moment the user is still somewhere it can be typed in.

## The address fallback still counts

Projects that predate the field often name their state in the address line, and `ComplianceGenerationService` already reads it that way. Requiring the column alone would refuse projects the generation behind the gate would have handled.

Both sides now resolve through one method, `IndianStateResolver.forProject(state, address)`. The gate cannot admit a project generation would turn away, and cannot turn away one generation would have handled. A gate that disagreed with the generator would be worse than no gate.

## The transition moved out of the patch loop

The event used to fire from inside the `updates.forEach` switch, on the `status` key. A user filling the state in on the same screen that approves sends both fields in one patch, and JSON preserves the order they were written in, so a gate in that position would have judged the patch on whichever field the payload listed first. The transition and the gate now run once, after the whole patch is applied.

## How many projects this would block

Measured on the compose staging (`.116`, `echno.in`), which is the environment with real usage:

| | count |
|---|---|
| projects | 27 |
| have the state field set | 0 |
| no field, state recovered from the address | 5 |
| neither, and not yet approved (would be blocked) | 21 |
| neither, and already approved (never re-gated) | 1 |

**21 would be refused.** They are demo and seed rows with placeholder addresses ("Emerald Heights Area", "Test 2 Address", "hgh"), not real project records. The count is high because the field only landed today, so nothing has one yet.

Worth being explicit that this removes no capability: none of those 21 could have generated compliance before this change either. They would have been approved and then failed generation silently. The gate moves the same failure to the request that can still fix it, with a message naming the field. The one already-approved row is project 27, address "Chennai", which is the exact case the state field exists for, and it is not re-gated because the check is on the transition rather than on the row.

## Not in scope

Geocoding auto-fill from the address is noted on the issue as needing a provider and a per-lookup cost. Left alone: it is Abhijith's call and independent of this.

## Verification

Every new test confirmed failing without its fix, on `lab-node-116-f`.

With `ProjectService` reverted to `development`:

```
approvingAProjectWithNoStateAnywhere_isRefusedAndPublishesNothing()  FAILED
theRefusalNamesTheFieldAndAnExampleOfWhatToPutInIt()                 FAILED
```

With the gate placed inside the patch loop instead of after it:

```
aPatchThatSetsTheStateAndApprovesInOneCall_isJudgedOnTheStateItSets()  FAILED
```

`ComplianceGenerationServiceTest` and `IndianStateResolverTest` pass unchanged, so routing both sides through `forProject` altered no existing behaviour. Full `./gradlew build`: BUILD SUCCESSFUL.

`ProjectApprovalStateTest` is a plain Mockito test in the style of `ProjectServicePaginationTest`. No Spring context, no entities, so nothing added to the context cache or the test heap.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/compliance/ComplianceGenerationService.java
+++ b/src/main/java/org/tornotron/echno_backend/compliance/ComplianceGenerationService.java
@@ -92,9 +92,10 @@ public class ComplianceGenerationService {
         // A project that states its own state is taken at its word; scanning the free-text
         // address is only the fallback for projects that predate the field, and it cannot find
         // what the address does not say (an address of "Chennai" names no state at all).
-        String state = project.getProjectState() != null
-                ? project.getProjectState()
-                : IndianStateResolver.resolve(project.getProjectAddress());
+        // Approval refuses a project this returns null for, so reaching the throw below means
+        // the state was cleared between approval and generation.
+        String state = IndianStateResolver.forProject(
+                project.getProjectState(), project.getProjectAddress());
         if (state == null) {
             throw new InvalidRequestException(
                     "This project has no state set, and its address does not name one, so the "
--- a/src/main/java/org/tornotron/echno_backend/compliance/IndianStateResolver.java
+++ b/src/main/java/org/tornotron/echno_backend/compliance/IndianStateResolver.java
@@ -55,6 +55,26 @@ public final class IndianStateResolver {
                 "'" + trimmed + "' is not an Indian state or union territory");
     }
 
+    /**
+     * The state a project's compliance rules are keyed by: its own state field where it has one,
+     * and otherwise whatever its free-text address happens to name.
+     *
+     * <p>Both the approval gate that refuses a project with no determinable state and the
+     * generation that runs straight afterwards resolve it through here, so the gate cannot admit
+     * a project that generation would then turn away, and cannot refuse one that generation
+     * would have handled.
+     *
+     * @param projectState The project's own state field, canonical or blank.
+     * @param projectAddress The project's free-text address, scanned only when the field is blank.
+     * @return The canonical state name, or null when neither route yields one.
+     */
+    public static String forProject(String projectState, String projectAddress) {
+        if (projectState != null && !projectState.isBlank()) {
+            return projectState;
+        }
+        return resolve(projectAddress);
+    }
+
     /** The canonical state name found in the address, or null if none matches. */
     static String resolve(String address) {
         if (address == null || address.isBlank()) {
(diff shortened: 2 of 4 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #525, opened 2026-08-28
Stop a controller being able to parse a payload it has not validated

### LINKED ISSUE #490: Multipart create endpoints never validate their payload

abhijithanandan:
Closes #490.

## What was left of the issue

The 11 endpoints and 32 constraints the issue names are already fixed on `development`, by the three PRs its own comment lists (#492 task and issue, #495 organization and chat, #500 attendance), plus #488 for project. I re-swept `readValue` across `src/main` and confirmed it: every creation payload is validated today, and nothing in this PR turns on a constraint that was not already running. **There is no behaviour change to any create payload, so nothing that succeeds today starts failing.**

What those PRs left behind is the reason the defect existed in the first place. Each added its own private `requireValid` to its own service, so there are now five identical copies of the same five lines, and, more to the point, nothing stops the twelfth endpoint being written exactly like the first eleven. The copies are the symptom. The pattern being trivial to write and invisible in review is the defect, and that is what this PR closes.

## The mechanism

`JsonPartBinder` reads every JSON part in the application:

```java
TaskCreationDto dto = jsonPartBinder.read(data, TaskCreationDto.class);   // parses, then validates
Map<String, Object> updates = jsonPartBinder.readUpdates(data);           // patch payloads, no bean to validate
```

`read` parses and validates in one call, so a caller cannot come away holding a payload whose constraints have not been checked. That is the property the eleven endpoints were missing, and it now holds by construction rather than by each author remembering. `readUpdates` covers the partial-update endpoints that read the part into a `Map`: there is no bean there and nothing to validate, and it is on the binder so that reading a part is one mechanism rather than two, which is what lets the guard rule below need no allowlist.

`PayloadValidator` is the single implementation of "check this bean or refuse it". The binder uses it at the request boundary; the five services keep their own call to it for any caller that reaches them another way. Both layers are deliberate: the binder guarantees the request path including an endpoint whose author never opens the service, the service guarantees the rest. It is a bean-validation pass over a small DTO, twice.

19 hand-parses across 12 controllers now go through the binder, and the five copies of `requireValid` are gone.

### Why here and not an argument resolver

The obvious alternative is a `HandlerMethodArgumentResolver` so the handler can declare `TaskCreationDto` directly. Two things rule it out. Custom resolvers are appended after Spring's own, so a parameter still annotated `@RequestPart` is claimed by `RequestPartMethodArgumentResolver` before mine is consulted, which means a new annotation. And springdoc builds the multipart request body from the handler's parameters, so a parameter it cannot interpret drops `data` out of the published OpenAPI document that Apidog syncs from. Keeping the signature and replacing the body costs one line per endpoint and changes the document not at all.

## The guard rail

`architecture/MultipartPayloadValidationTest`, alongside `EndpointAuthorizationTest` and `UnboundedRepositoryReadTest` and sharing their `@AnalyzeClasses` import cache. Two rules, because the defect had two halves:

- **`noControllerDeserializesItsOwnPayload`** bans a Jackson `readValue` inside anything meta-annotated `@Controller`. This is the half that actually skips validation. Banning the call rather than trying to prove a payload went unvalidated has no false positives and fails on the line that introduced it.
- **`noControllerAsksForAStringToBeValidated`** bans `@Valid` on a `String` parameter. It validates nothing, and it is worth failing the build over rather than deleting quietly, because it is the marker that made all eleven endpoints look validated to every reader including reviewers.

Against the parent commit they report **19 hand-parses and 13 inert `@Valid` annotations**. Both fail there and pass here.

## A fix that falls out of the consolidation

`ProjectController.createProject` built its own `new ObjectMapper()` instead of taking the configured one. A bare Jackson mapper has no `JavaTimeModule`, and `ProjectCreationDto` carries `startDate` and `endDate` as `LocalDateTime`, so that endpoint refused any project that named its dates, with `Java 8 date/time type not supported by default`. Its `/web` twin has always used the configured mapper, so the two disagreed. Both now go through the binder.

The same swap makes that endpoint tolerant of unknown properties, since the configured mapper disables `FAIL_ON_UNKNOWN_PROPERTIES` and a bare one does not. That is the looser direction and it matches every other create endpoint.

## No constraint looks wrong

The brief asked me to call out any constraint that looks wrong rather than enforce it silently. I read all 32 again against the DTOs and the web forms. There is nothing to report, because the six that were wrong were already corrected by the PRs that turned validation on, and each correction is recorded in the DTO next to the constraint:

| Constraint | Was | Is | Why |
|---|---|---|---|
| task/issue `title` | `@Size(max = 50)` | 255 | column is `VARCHAR(255)`, no form caps it |
| issue `description` | `@Size(max = 500)` | 2000 | `TEXT` column, message also claimed a minimum of ten against a constraint asking five |
| task `assigneeIds`, `tags` | `@NotNull` | optional | the service has always skipped an absent or empty list |
| `organizationAddress` | `@Size(max = 50)` | 255 | `TEXT` column behind a three-row textarea |
| `organizationPhone` | `@Size(9, 15)` | 9 to 16 | E.164 counted digits and forgot the leading plus |
| `organizationEmail` | TLD `{2,6}` | `{2,24}` | refused `.construction`, `.engineering`, `.contractors` |
| `organizationName` | `@Size(min = 3)` | 2 | refused "3M", "HP" |

The one client payload that newly fails is already filed: the issue form's "Save as Draft" sends `"description": ""` past `@NotBlank`, tracked as tornotron/echno-web#290. Nothing here changes that either way.

## Verification

Run on the lab build box (`10.24.6.116`), so the Testcontainers suites have a real CockroachDB. The tree there was checksum-matched against the branch before the run.

```
$ ./gradlew test
817 tests, BUILD SUCCESSFUL in 7m 13s
```

New tests, and what each is worth:

| Test | Broken by | Result |
|---|---|---|
| `MultipartPayloadValidationTest.noControllerDeserializesItsOwnPayload` | running it against the parent commit | **FAILED**, 19 violations |
| `MultipartPayloadValidationTest.noControllerAsksForAStringToBeValidated` | running it against the parent commit | **FAILED**, 13 violations |
| `JsonPartBinderTest.read_refusesAPayloadThatFailsAConstraint` | dropping the validate from `read` | **FAILED** |
| `JsonPartBinderTest.read_refusesAPayloadMissingARequiredField` | dropping the validate from `read` | **FAILED** |
| `JsonPartBinderTest.read_refusesAnAbsentPart` | dropping the absent-part guard | **FAILED** |
| `JsonPartBinderTest.read_returnsAValidPayload` | (holds either way, pins the pass-through) | passed |
| `JsonPartBinderTest.read_refusesAPartThatIsNotJson` | (holds either way, pins the parse failure) | passed |
| `JsonPartBinderTest.readUpdates_readsTheFieldsSent_andTreatsAnAbsentPartAsNoChange` | (holds either way, pins the patch contract) | passed |

All restored to green afterwards. Plain JUnit with a real Hibernate Validator and a real `ObjectMapper`, no new Spring context, so nothing is added to the test JVM's context cache.

The four `@WebMvcTest` slices over these controllers (`ProjectControllerWebAuthzTest`, `OrganizationWebControllerAuthzTest`, `ReadWriteGuardSymmetryTest`, `ChatStreamEndpointTest`) import the two new beans. A slice does not pick up a plain `@Component`, and without the import each fails to load its context.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/attendance/AttendanceController.java
+++ b/src/main/java/org/tornotron/echno_backend/attendance/AttendanceController.java
@@ -5,7 +5,7 @@ import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import org.springframework.security.access.prepost.PreAuthorize;
-import com.fasterxml.jackson.databind.ObjectMapper;
+import org.tornotron.echno_backend.common.payload.JsonPartBinder;
 import jakarta.validation.Valid;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.PageRequest;
@@ -38,11 +38,11 @@ import java.util.List;
 public class AttendanceController {
 
     private final AttendanceService attendanceService;
-    private final ObjectMapper objectMapper;
+    private final JsonPartBinder jsonPartBinder;
 
-    public AttendanceController(AttendanceService attendanceService, ObjectMapper objectMapper) {
+    public AttendanceController(AttendanceService attendanceService, JsonPartBinder jsonPartBinder) {
         this.attendanceService = attendanceService;
-        this.objectMapper = objectMapper;
+        this.jsonPartBinder = jsonPartBinder;
     }
 
     @PostMapping(value = "/check-in",consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
@@ -58,9 +58,9 @@ public class AttendanceController {
             @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "400", description = "The data part is not valid check-in JSON, or a field failed validation"),
             @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "403", description = "Caller is not a member of the current tenant")
     })
-    public ResponseEntity<AttendanceResponseDto> checkIn(@RequestParam("data") @Valid String data,
+    public ResponseEntity<AttendanceResponseDto> checkIn(@RequestParam("data") String data,
                                                          @RequestParam(value = "photoUrl", required = false) MultipartFile photoUrl) throws JsonProcessingException {
-        AttendanceCheckInDto dto = objectMapper.readValue(data, AttendanceCheckInDto.class);
+        AttendanceCheckInDto dto = jsonPartBinder.read(data, AttendanceCheckInDto.class);
         return ResponseEntity.status(HttpStatus.CREATED).body(attendanceService.checkIn(dto,photoUrl));
     }
 
@@ -80,9 +80,9 @@ public class AttendanceController {
             @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "403", description = "Caller is not a member of the current tenant"),
             @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "404", description = "No attendance record with the given id")
     })
-    public ResponseEntity<AttendanceResponseDto> recordClockEvent(@RequestParam("data") @Valid String data,
+    public ResponseEntity<AttendanceResponseDto> recordClockEvent(@RequestParam("data") String data,
                                                                   @RequestParam(value = "photo", required = false) MultipartFile photo) throws JsonProcessingException {
-        AttendanceClockEventDto dto = objectMapper.readValue(data, AttendanceClockEventDto.class);
+        AttendanceClockEventDto dto = jsonPartBinder.read(data, AttendanceClockEventDto.class);
         return ResponseEntity.ok(attendanceService.recordClockEvent(dto,photo));
     }
 
--- a/src/main/java/org/tornotron/echno_backend/attendance/AttendanceControllerWeb.java
+++ b/src/main/java/org/tornotron/echno_backend/attendance/AttendanceControllerWeb.java
@@ -5,7 +5,7 @@ import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import org.springframework.security.access.prepost.PreAuthorize;
-import com.fasterxml.jackson.databind.ObjectMapper;
+import org.tornotron.echno_backend.common.payload.JsonPartBinder;
 import jakarta.validation.Valid;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.PageRequest;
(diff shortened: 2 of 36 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Rebased onto current `development`, which had moved on by four commits since this branch was cut. Two of them matter here.

`b6e2f63` ("Let a new account create its first organization") changes the guard and the documented responses on both organization create endpoints, the same two methods this PR touches. The rebase kept its version: `@PreAuthorize("isAuthenticated()")`, no `@RequireSubscription`, and the 402 stay exactly as that commit left them. Only the payload read inside the method body is this PR's.

It also added `OrganizationServiceEntitlementTest`, which builds an `OrganizationService` and so had to move to `PayloadValidator` like the other nine construction sites. That is the one extra file in the amended commit.

Full suite re-run on the lab box against the rebased tree, checksum-matched to the branch first: `BUILD SUCCESSFUL in 7m 33s`, 852 test cases across the result files, none failed or errored.

PR #526, opened 2026-08-28
Cache a copy of the subscription rather than the entity

### LINKED ISSUE #485: SubscriptionCache holds entities, and caching DTOs instead would freeze the computed entitlement fields

abhijithanandan:
Closes #485.

## The decision

The cache holds an internal `SubscriptionSnapshot`: an immutable copy of the subscription row and its plan, taken inside the loading transaction, carrying the timestamps the entitlement flags are derived from rather than the flags themselves.

Neither of the two obvious options works.

**Not the entity**, which is what it held. A cached entity is detached by definition and outlives the persistence context that produced it, so anything left uninitialized on it throws `LazyInitializationException` on a later hit, in a transaction that cannot reattach it. That was survivable only while `findActiveSubscriptionByUserId` kept fetch-joining the whole graph, which no signature obliged it to do. It is also a mutable object shared by every concurrent reader of that user.

**Not `SubscriptionDto`**, for the reason the issue title gives. Which fields are computed, and when they change:

| DTO field | derived from | changes when |
|---|---|---|
| `expired` | `currentPeriodEnd` vs `Instant.now()` | on its own, as the clock passes the period end |
| `inTrial` | `status` + `trialEnd` vs `Instant.now()` | on its own, as the clock passes the trial end |
| `active` | `status` alone | only when a write changes the status |

The first two are answers to "as of when". Cached as booleans they are settled for the life of the entry, and a subscription that lapses inside the five minute window keeps reporting `expired = false` for the rest of it. Those are the fields the paywall will read.

`active` is not of that kind, and the issue lists it as though it were. It reads the status alone, so it is exactly as old as the cached value however that value is shaped. Eviction and the five minute expiry bound it; no choice of value type does, and if the paywall needs a tighter bound the lever is the TTL. A test states this beside the two that do depend on the clock.

So: **the cache stores facts, never verdicts.** `isExpired(Instant)` and `isInTrial(Instant)` take the instant to judge against, and `BillingMapper` resolves them against `Instant.now()` when it builds the response. Both are `@JsonIgnore` for the same reason, so a shared cache cannot persist an answer either.

## What must not be cached at all

Quota usage. It moves on every metered request and is the one input to an entitlement decision that could not be held behind the subscription without being wrong almost immediately. `checkQuotaAccess` still sums it from the usage table on every call, and the snapshot deliberately does not carry it.

## Point 2 of the issue, the feature id

Quota checks key on `UsageRecord.featureId`, a plain column rather than a relation, which no response DTO carries. That only forces a change to `PlanFeatureDto` if the cache has to hold the public DTO. The snapshot is internal, so it carries the id and `PlanFeatureDto` is untouched. No contract change, no core release.

## Two things that fall out

`recordUsage` no longer evicts the user's entry. Nothing it writes is part of what the cache holds, so the eviction bought no freshness and threw the entry away on the one path that runs on every metered request.

Resolving the quota period no longer goes through `map(...).findFirst()`, which throws on a plan that grants a metered feature without setting a period rather than falling through to the intended monthly default.

## The staleness the representation cannot fix

A snapshot carries a copy of the plan it was taken against, so a plan write changes what every subscriber of that plan is entitled to while their cached copies say otherwise, and no per-user eviction fires because no subscription row changed. The entity cache had this too, for the same reason: what it held was a detached plan graph read at some earlier moment. The entries are not addressable by user without a query and plan writes are rare administrative operations, so `PlanService` now empties the cache on every write, on the same evict-now-and-again-on-completion pattern as the per-user path.

## Redis

The cache is in process, so those evictions reach one replica. `RedisConfig` already carries the way out: it configures a `subscriptions` cache with the five minute TTL, in so many words, for a future `@Cacheable("subscriptions")`.

That cache would have failed on its first real value. `GenericJackson2JsonRedisSerializer`'s default mapper has no `JavaTimeModule` and refuses to write an `Instant` at all, and a snapshot is mostly period bounds and a trial window. Nothing caught it because the existing test caches a `String`. The value serializer now registers `JavaTimeModule` and writes ISO-8601 rather than numeric timestamps.

This is a serializer configuration inside the existing `@ConditionalOnProperty` bean method, not a Spring Data Redis repository, so it is not the shape of the boot crash we hit before. `RedisFullContextBootIT` (full context, `echno.cache.provider=redis`, real Redis container) and `RedisUnreachableBootIT` both pass.

Actually moving the cache onto `@Cacheable` is not in this change. The snapshot makes it possible; whether it is worth the Redis round trip on the hot entitlement path is a separate decision, and the payment gateway work will have opinions about what the cache must hold.

## The fetch joins

They stay, and now mean something different. The copy is taken inside the transaction, so without them it still comes out complete and merely costs a query per plan feature on a miss. Weakening them is a performance regression rather than the 500 it used to be. The test that pinned them says so.

## Verification

Full suite green on the lab box (`10.24.6.116`), 7m09s, including `BillingMappingIT` (19), `SubscriptionServiceFeatureAccessTest` (10), `SubscriptionSnapshotTest` (6), `RedisDistributedStateIT` (4), `RedisFullContextBootIT`, `RedisUnreachableBootIT`.

Each new test was checked against a mutation of the production code that should break it:

| Mutation | Test that failed |
|---|---|
| `isExpired`/`isInTrial` answer for a fixed moment instead of the instant asked | `expiredIsAnsweredForTheInstantAsked`, `inTrialIsAnsweredForTheInstantAsked`, `activeReadsTheStatusAloneAndNotTheClock` |
| `active` consults the period end as well as the status | `activeReadsTheStatusAloneAndNotTheClock` |
| feature lookup matches the display name instead of the code | `featureLookupFindsTheGrantAndItsFeatureId` |
| no defensive copy of the feature list | `theFeatureListIsCopiedInAndHandedOutUnmodifiable` |
| a persistent entity becomes reachable from the snapshot | `nothingReachableFromASnapshotIsAnEntity` |
| recording usage evicts the entry again | `recordUsage_doesNotEvictTheCachedSubscription` |
| the snapshot stops carrying the feature id | `recordUsage_keysTheUsageRowOnTheFeatureIdFromTheCachedValue` (and both quota tests) |
| the old quota-period resolution | `recordUsage_fallsBackToAMonthlyPeriodWhenThePlanSetsNone` |
| the response flags are settled when the snapshot is taken | `aCachedSubscriptionWhosePeriodHasEnded_readsAsExpired` |
| a plan write leaves the cached subscriptions alone | `removingAFeatureFromAPlan_invalidatesTheCachedSubscriptions` |
| the stock Redis value serializer | `aSubscriptionSnapshotIsSharedAcrossTwoCacheManagersViaRedis` |

No public API or response contract changes. `PlanFeatureDto`, `PlanDto` and `SubscriptionDto` are byte-for-byte the same shape.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/billing/components/SubscriptionCache.java
+++ b/src/main/java/org/tornotron/echno_backend/billing/components/SubscriptionCache.java
@@ -9,15 +9,43 @@ import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.support.TransactionSynchronization;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
-import org.tornotron.echno_backend.billing.Subscription;
+import org.tornotron.echno_backend.billing.snapshot.SubscriptionSnapshot;
 
 import java.time.Duration;
 
+/**
+ * Per-user cache of the active subscription, so the entitlement check on an annotated endpoint
+ * does not query the subscription and its plan graph on every request.
+ *
+ * <p>What it holds is a {@link SubscriptionSnapshot}: an immutable copy of the row and its
+ * plan, taken while the loading transaction is still open. It deliberately holds neither of
+ * the two obvious alternatives.
+ *
+ * <p>Not the {@code Subscription} entity, which is what it used to hold. An entity in a cache
+ * is detached by definition and outlives the persistence context that produced it, so anything
+ * left uninitialized on it throws {@code LazyInitializationException} on a later hit, in a
+ * transaction that cannot reattach it. That was survivable only as long as the loading query
+ * kept fetch-joining the whole graph, which no signature enforced. It is also a mutable object
+ * shared by every concurrent reader of that user, so any write path that touched it published a
+ * half-finished change to all of them at once.
+ *
+ * <p>Not {@code SubscriptionDto} either, which is the tempting fix. Two of its entitlement
+ * flags, {@code expired} and {@code inTrial}, are comparisons against the current instant, so
+ * caching them freezes them for the life of the entry and a subscription that lapses inside
+ * that window keeps reporting itself unexpired. The snapshot keeps the timestamps those flags
+ * are derived from and leaves the derivation to read time.
+ *
+ * <p>Freshness is bounded by the five minute expiry and by eviction: every write that changes
+ * a user's subscription evicts that user's entry, and every write that changes a plan evicts
+ * all of them, because a snapshot carries a copy of the plan it was taken against. The cache
+ * is in process, so those evictions reach one replica. Under more than one replica the
+ * remaining bound on the others is the expiry.
+ */
 @Component
 @Slf4j
 public class SubscriptionCache {
 
-    private final Cache<@NonNull Long, Subscription> cache;
+    private final Cache<@NonNull Long, SubscriptionSnapshot> cache;
 
     public SubscriptionCache() {
         this.cache = Caffeine.newBuilder()
@@ -27,11 +55,11 @@ public class SubscriptionCache {
                 .build();
     }
 
-    public Subscription get(Long userId) {
+    public SubscriptionSnapshot get(Long userId) {
         return cache.getIfPresent(userId);
     }
 
-    public void put(Long userId, Subscription subscription) {
+    public void put(Long userId, SubscriptionSnapshot subscription) {
         cache.put(userId, subscription);
     }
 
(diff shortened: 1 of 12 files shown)


abhijithanandan:
@coderabbitai review

PR #527, opened 2026-08-28
Refuse work that reaches tenant data with no tenant declared

### LINKED ISSUE #507: Tenant isolation fails open on a null TenantContext: both the orgFilter aspect and the load listener give up silently

abhijithanandan:
Closes #507.

## The problem

Tenant isolation had two mechanisms and both failed open on a null tenant context.

`HibernateFilterConfig` skipped the `orgFilter` and logged at debug, a level no deployed environment turns on. `TenantIsolationLoadListener` returned early on the same condition. They read as defence in depth and were not: they cover different things, query shape against primary-key loads, but gave up on identical input. A null tenant did not narrow isolation, it removed it, and left nothing behind to distinguish that from a correct run.

## The shape of the fix

The absent case now has to say which of two things it is. There are three ways to declare a tenant scope, and declaring one is mandatory for work that reaches tenant-scoped data:

- **an organization id**, set by `TenantFilter` from the request or by `TenantScopedJobRunner` from durable state,
- **a bypass** (`@BypassTenantFilter`), meaning read across organizations on purpose, already audited at WARN where it is set,
- **`@WithoutTenant("reason")`**, new here, meaning this work belongs to no organization.

Anything else is a fourth state, undeclared, and goes to the new `UnscopedAccessGuard`.

`@WithoutTenant` is deliberately weaker than `@BypassTenantFilter`. It suppresses the missing-scope failure and nothing else, so where an organization id is in force the `orgFilter` and the cross-tenant check stay fully on. That is what makes it safe on shared service code a tenant request may also call, and it is pinned by a test in both the unit suite and the IT.

## Why the two boundaries get different defaults

`echno.multitenancy.load-boundary` defaults to **DENY**. It fires when a `TenantScopedEntity` is actually read, which is the leak itself rather than a proxy for it, so the refusal is precise and attributable to the row that caused it.

`echno.multitenancy.transaction-boundary` defaults to **WARN**. It fires on every `@Transactional` method in the package tree, including the many that touch no tenant-scoped entity at all: users, organizations, plans, subscriptions. Denying there would refuse work that was never at risk.

Both take ALLOW, WARN or DENY, so a deployment can tighten ahead of the default or step back during an incident without a code change. Under WARN each event increments `echno.multitenancy.unscoped` (tagged with boundary, call site and policy) and logs the call site, rate limited per call site so an unscoped read inside a loop cannot bury what it is reporting. The counter is never rate limited, so the dashboard number stays a total rather than a sample.

**The transaction boundary should move to DENY on evidence, not on a decision.** The condition is the warning having been silent for a week under real staging traffic. That is the observation step #507 asks for, and it exists because the alternative is trusting the enumeration below, which is the next section's problem.

## Where the soft edge is

**The fail-closed guarantee rests on the completeness of the tenant-less enumeration.** That enumeration is static analysis of controller-to-service chains, including materialization through MapStruct lazy collections, so it is a claim about code paths rather than a measurement of them. If a path was missed, it answers 403 rather than leaking, which is the right direction to fail, but it is a production incident and the lever for it is `ECHNO_TENANT_LOAD_BOUNDARY=WARN`.

The paths found, all of which legitimately reach tenant-scoped data with no tenant:

| Path | What it touches | How it is declared now |
|---|---|---|
| The no-membership state: profile lookup, organization creation, invite-code redemption, the organization picker | `Employee`, `Project`, `Attachment`, `Task`, `ShiftTiming`, `ProjectInviteCode` | `TenantFilter`, when the token carries no `ORG_MEMBER_` authority |
| `GET /user/web` | `Attachment` | `TenantFilter`, pre-tenant path |
| `/billing/**` | keyed on the user, not the organization | `TenantFilter`, pre-tenant path |
| `/auth/register` | pre-tenant by definition | `TenantFilter`, pre-tenant path |
| Startup: `KeycloakInitializer` to `DevFixtureProvisioner.provision` | `Employee`, `ShiftTiming` | `@WithoutTenant`, the only annotation site in the codebase |

The request-side declarations are handled centrally in `TenantFilter` rather than by annotating the ten or so service methods they reach. The exemptions then sit in one place that can be read as a list, and the services stay honest about having no opinion on the matter.

Behaviour of those requests is otherwise unchanged, which the tests pin: no organization id, no membership check, no bypass. Only the declaration is new. `shouldNotFilter` now covers `/actuator` alone, so the three pre-tenant API paths run through `doFilterInternal` to declare themselves instead of being skipped outright; the early branch replicates the old skip exactly, plus the declaration and the `finally` clear.

## Ordering

`WithoutTenantAspect` and `TenantFilterBypassAspect` are now both ordered ahead of `HibernateFilterConfig`, which reads the scope on entry. Neither was ordered at all before, which left to chance the same one-step ordering that #508 turned out to be. A reflection test asserts the relation rather than trusting the annotation to stay put.

## Tests

60 new or changed tests in the multitenancy package, plain JUnit and Mockito except the IT, which reuses the existing `TenantIsolationIT` context and adds no new one.

Every new test was checked by breaking its own fix and confirming it goes red:

| Mutation | Test that caught it |
|---|---|
| Load listener returns early on a null org again | `TenantIsolationIT.findById_withNoTenantScopeDeclaredAtAll_isRefused`, `TenantIsolationLoadListenerTest.deniesALoadThatDeclaresNoScopeAtAll` |
| Transaction boundary logs instead of calling the guard | `HibernateFilterConfigTest.refusesATransactionThatDeclaresNoScopeAtAllWhenTheBoundaryDenies` |
| `/auth/register` stops declaring | `TenantFilterTest.selfRegistrationDeclaresItselfUnscoped` |
| `/user/web` stops declaring | `TenantFilterTest.theCurrentUserLookupDeclaresItselfUnscoped` |
| No-membership branch stops declaring | `TenantFilterTest.anAuthenticatedUserWithNoMembershipDeclaresItselfUnscoped` |
| `TenantScopedJobRunner` keeps an inherited unscoped declaration | `TenantScopedJobRunnerTest.withdrawsAnUnscopedDeclarationForTheWorkAndRestoresIt` |
| Aspect order collides with the filter aspect | `WithoutTenantAspectTest.theScopeDeclaringAspectsRunAheadOfTheFilterAspect` |
| Counter becomes rate limited like the log | `UnscopedAccessGuardTest.theCounterIsNotRateLimitedEvenThoughTheLogIs` |
| `declareUnscoped` stops requiring a reason | `TenantContextTest.anUnscopedDeclarationNeedsAReason` |

Full suite green on the lab build box: 854 tests, 141 classes, 0 failures, 0 errors.

## Scope

No API change, no schema change, no `echno-core` or `echno-web` change.

#508 needed nothing here: it was fixed and closed by #514, which merged into `development` this morning. This branch is built on top of it. The pairing rationale still applied while building, since making #507 fail closed is what would have turned #508 from a silent ordering accident into a loud one.

## Related

- #507, closed here
- #508, the ordering accident, already fixed by #514
- #514, the compliance transaction boundary, which this builds on
- #482, the background-job design where both were found
- #330, the fail-closed load-listener work whose null-context early return was half of this

## Summary by CodeRabbit

* **New Features**
  * Added explicit support for operations that intentionally run without an organization scope.
  * Added configurable handling for unscoped database access: allow, warn, or deny.
  * Added metrics and warnings for unscoped access, with denial enabled by default for data loads.
  * Improved tenant isolation by rejecting undeclared tenant scope instead of silently proceeding.
* **Bug Fixes**
  * Corrected tenant-scope handling across requests, background jobs, and database loads.
  * Preserved and restored tenant context reliably during nested operations.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/configuration/DevFixtureProvisioner.java
+++ b/src/main/java/org/tornotron/echno_backend/common/configuration/DevFixtureProvisioner.java
@@ -5,6 +5,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 import org.tornotron.echno_backend.common.enums.OrgRole;
 import org.tornotron.echno_backend.common.multitenancy.TenantContext;
+import org.tornotron.echno_backend.common.multitenancy.WithoutTenant;
 import org.tornotron.echno_backend.common.service.KeycloakGroupService;
 import org.tornotron.echno_backend.employee.Employee;
 import org.tornotron.echno_backend.employee.EmployeeRepository;
@@ -87,6 +88,8 @@ public class DevFixtureProvisioner {
      * @param displayName    the name to store on the user row and copy onto the employee row.
      * @param email          the address the organization picker matches on.
      */
+    @WithoutTenant("The dev fixture creates the organization it then works inside, so the first "
+            + "steps run before any tenant exists")
     public void provision(String keycloakUserId, String displayName, String email) {
         if (keycloakUserId == null || keycloakUserId.isBlank()) {
             log.error("No Keycloak id for the dev fixture account; skipping its application identity. "
--- a/src/main/java/org/tornotron/echno_backend/common/exception/UnscopedTenantAccessException.java
+++ b/src/main/java/org/tornotron/echno_backend/common/exception/UnscopedTenantAccessException.java
@@ -1,0 +1,22 @@
+package org.tornotron.echno_backend.common.exception;
+
+/**
+ * Thrown when work reaches tenant-scoped data with no tenant scope declared at all: no
+ * organization id, no {@code @BypassTenantFilter}, no {@code @WithoutTenant}.
+ *
+ * <p>It is not the caller asking for another organization's row, which is
+ * {@link TenantAccessDeniedException} proper. It is the application unable to say which
+ * organization the work belongs to, and refusing to guess. Before #507 that condition returned
+ * every organization's rows and logged nothing.
+ *
+ * <p>It extends {@link TenantAccessDeniedException} so a request that hits it answers 403 with
+ * the same non-committal body as any other isolation refusal, rather than describing the internal
+ * state to the caller. The detail belongs in the log, which
+ * {@code org.tornotron.echno_backend.common.multitenancy.UnscopedAccessGuard} writes.
+ */
+public class UnscopedTenantAccessException extends TenantAccessDeniedException {
+
+    public UnscopedTenantAccessException(String message) {
+        super(message);
+    }
+}
--- a/src/main/java/org/tornotron/echno_backend/common/multitenancy/HibernateFilterConfig.java
+++ b/src/main/java/org/tornotron/echno_backend/common/multitenancy/HibernateFilterConfig.java
@@ -12,6 +12,27 @@ import org.springframework.core.annotation.Order;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
 
+/**
+ * Enables the Hibernate {@code orgFilter} on the session for the duration of a
+ * {@code @Transactional} method, so queries in that transaction see only the current
+ * organization's rows.
+ *
+ * <p>Three states arrive here and each is now answered explicitly:
+ *
+ * <ul>
+ *   <li>An organization id is in force, so the filter goes on. The ordinary case.
+ *   <li>Tenant isolation is deliberately off, either a {@code @BypassTenantFilter} reading across
+ *       organizations or a {@link WithoutTenant} method that belongs to none. The filter stays
+ *       off because that is what was asked for.
+ *   <li>Nothing at all was declared. Until #507 this fell in with the case above and the
+ *       transaction ran unfiltered, recorded only in a debug line no deployed environment has on.
+ *       It now goes to {@link UnscopedAccessGuard}.
+ * </ul>
+ *
+ * <p>This aspect reads the scope on entry, which is why anything that establishes one has to be
+ * in place before the aspect runs rather than inside the method body. Setting it in the body is
+ * one step too late and leaves the transaction unfiltered, which is what #508 was.
+ */
 @Aspect
 @Component
 @RequiredArgsConstructor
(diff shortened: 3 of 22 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
Note before merge: CodeRabbit has not reviewed this PR. The commit status reads `success` but its description is "Review rate limited", which means no review ran, not that one passed. The free OSS quota was exhausted and two `@coderabbitai review` re-requests (16:35 and 17:27 UTC) were both refused with "Review rate limited".

Re-request once the quota resets and read the findings before this goes in. Given what it touches, green CI on its own is not enough here.

abhijithanandan:
Coordination note from #542, which adds a tenant-less path your enumeration should list: the compliance generation queue. Its dispatcher polls compliance_generation_jobs every 2 s through TransactionRetryTemplate with no org id, no bypass and no @WithoutTenant, and the worker reads the claimed row's dispatch data the same way before it enters TenantScopedJobRunner. None of that ever loads a tenant-scoped entity (native SQL, scalars only, by design), so your load boundary stays quiet. But every poll is an undeclared transaction at the transaction boundary: under WARN it never goes silent, which stalls the move-to-DENY-on-evidence plan, and under DENY the queue stops draining. Whichever branch lands second should put @WithoutTenant on the dispatcher poll path and the worker's dispatch-row read. Everything after the claim already runs under TenantScopedJobRunner, so nothing else needs declaring.

Issue #530, opened 2026-08-28
Site transfer and PO numbers are invented by the client, the unique constraint is global while the check is per-org, and the duplicate message is discarded


abhijithanandan:
## Reported

Anand, QA pass on `echno.in`, 28 Aug 2026. ClickUp `86d45vqh8` and `86d45vnx9`.

New site transfers and new purchase orders both collide on sequence `000001`. From the staging log for 28 Aug:

```
DuplicateResourceException: Site transfer with number TRF-2026-000001 already exists
DuplicateResourceException: Purchase Order with PO number PO-2026-000001 already exists
```

The proximate cause is in the browser and is filed as echno-web #WEB_NUMBERING. This issue is the reason the browser was in a position to cause it.

### 1. The number is supplied by the client

Both DTOs require it and the services copy it straight onto the entity:

- `siteTransfer/dto/SiteTransferCreationDto.java:18-21`, `@NotBlank private String transferNumber`, written at `SiteTransferService.java:136`
- `purchaseOrder/dto/PurchaseOrderCreationDto.java:19-22`, written at `PurchaseOrderService.java:116`

There is no reservation, no database sequence, no `max + 1`. The server never generates a number, it only checks the one it is handed. Two clients that ask at the same moment will always be able to propose the same number, whatever the frontend does.

### 2. The uniqueness check and the constraint disagree on scope

The service pre-check is organisation-scoped (`SiteTransferService.java:106`, `PurchaseOrderService.java:96`). The database constraint is global:

`db/changelog/v4.0/baseline-002-level1-tables.xml:353-354`
```xml
<column name="transfer_number" type="VARCHAR(100)">
    <constraints nullable="false" unique="true" uniqueConstraintName="uk_site_transfer_number"/>
```

Same shape for `po_number` at `baseline-003-level2-tables.xml:86`, and JPA mirrors it at `SiteTransfer.java:37`.

With one tenant on staging this does not bite. With two it will: tenant B proposing `TRF-2026-000001` passes the org-scoped check and then dies on the global constraint, surfacing through the `DataIntegrityViolationException` handler rather than as a clean duplicate. That is a multi-tenant production bug waiting for the second client.

### 3. The duplicate message is thrown away before the user sees it

`common/exception/GlobalExceptionHandler.java:177-183`:

```java
@ExceptionHandler(DuplicateResourceException.class)
@ResponseStatus(HttpStatus.CONFLICT)
public ProblemDetail handleDuplicateResourceException(...) {
[...]

Issue #529, opened 2026-08-28
Consumption is validated against a (material, project, location) row that nothing guarantees exists, and there is no way to create or correct one


abhijithanandan:
## Reported

Anand, QA pass on `echno.in`, 28 Aug 2026. ClickUp `86d45vfxq`, `86d45vrrm`, `86d45vtby`.

Recording a consumption of 1 MT of TNT Steel against project 3, storage location 14 is refused:

```
Insufficient stock for material ID 2 at project ID 3, storage location ID 14. Required: 1.00, Available: 0.00
```

The frontend half of this (the form showing an org-wide total instead of the location balance) is filed as echno-web #WEB_STOCK. This issue is the backend half, which is the more consequential one.

### 1. Nothing checks that a storage location belongs to the project it is booked against

`MaterialConsumptionService.java:129-130` resolves the storage location with `findByIdAndOrganization_Id`. That is organisation scope only. `StorageLocation` does carry a `project` (`StorageLocation.java:43-44`), so the check is available, it is just never made. The only cross-entity check in that method is the task-to-project one at lines 140-144.

On staging, storage location 14 has `project_id = NULL`. It belongs to no project at all, and the API accepted it against project 3 without complaint. Seven of the thirteen locations on staging have a null project, so this is not an isolated row.

The same gap exists on material creation, `MaterialService.java:113-127`, which resolves project and location independently.

### 2. A missing balance row is indistinguishable from a real zero

`InventoryService.getStockAtLocation` (`InventoryService.java:82-86`):

```java
return currentStockRepository
        .findByMaterialIdAndProjectIdAndStorageLocationId(materialId, projectId, storageLocationId)
        .map(CurrentStock::getCurrentQuantity)
        .orElse(0.0);
```

"This material has never been stocked here" and "this location is empty" produce the same message. The first is a user error we could name precisely, the second is a genuine stock-out. Telling them apart is most of what would have saved Anand the ticket.

### 3. There is no way to set or correct stock at a location after the material is created

This is the substance of `86d45vfxq`. Anand's framing is that Current Stock should not be directly editable in the material master. It already is not: `MaterialUpdateDto` has no `openingStock` field and `MaterialService.updateMaterial` never touches it. The actual problem is the opposite of what the ticket assumes. There is no controlled path at all.

Every writer of `current_stock` goes through `InventoryService.updateCurrentStock` (`InventoryService.java:313`). Its complete set of callers:

[...]

Issue #522, opened 2026-08-28
Rework the DTO converter component before it turns into a scaling problem


abhijithanandan:
Hrishi raised this himself on 25 August: the DTO converter component works, but the way it is put together will not hold once the entity count and the request volume grow, and he asked for it to be revisited rather than left as is.

Scope:

- Audit the converter component and write down what it actually costs per request.
- Decide whether it stays hand-rolled or moves to a mapper, and settle that once for the whole codebase rather than per module.
- Remove any per-request reflection or repeated lookups on the conversion path.
- Start with the finance and inspection entities, since those carry the widest DTOs and will show the problem first.

Raised by Hrishikesh, 25 August 2026.

abhijithanandan:
I read the whole conversion layer before answering this, because the issue asks for a decision that would touch every module and I did not want to make it from an impression. Numbers first, then where I think Hrishi is right, where the premise does not hold, and what I would actually do.

Short version: **the mechanism is already settled and does not need reworking. The cost is real but it is in three specific places, none of them where the issue says to look, and the fix is local rather than architectural.** I am recommending against a rewrite and for four scoped changes.

---

## 1. What the design actually is

The issue's premise is that the conversion layer is inconsistent. Counted, it mostly is not.

| | Count |
|---|---|
| `*Mapper` classes | 59 |
| of which MapStruct `@Mapper` | **52** |
| of which hand-written | **7** (6 in `attendance`, 1 in `billing`) |
| `*Converter` classes | 9, and **none of them convert an entity to a DTO** (7 are Spring `Converter<String, Enum>` for request params, 1 is `JwtAuthConverter`, 1 is a JPA `AttributeConverter`) |
| Static factory methods on DTOs (`fromEntity`, `of`, `toDto`) | **0** |
| Reflection-based copying (`BeanUtils.copyProperties`, ModelMapper, `objectMapper.convertValue`) | **0** |
| `*Dto` classes | 204 |

So three of the four things the issue's scope list asks for are already done:

- **"Decide whether it stays hand-rolled or moves to a mapper, and settle that once for the whole codebase."** It was settled. 52 of 59 are MapStruct, and the 17 `chore/mapstruct-*` branches from mid-August are the migration that did it. The remaining 7 are 583 lines total and are a finishing job, not a decision.
[...]

PR #538, opened 2026-08-28
Record what the material mapper costs per row

abhijithanandan:
Evidence for the converter audit on #522. Test only; no production code is touched.

## What it pins

`MaterialMapper.fillStock` is an `@AfterMapping` hook that asks `InventoryService` for two aggregates keyed on the material id. The generated `MaterialMapperImpl` calls it once per mapped material:

```java
materialDto.setReorderLevel( material.getReorderLevel() );

fillStock( material, materialDto );
```

so the cost is per row and not per request, and nothing at the call site shows it. `MaterialService.getAllMaterials()` reads

```java
return materialRepository.findAll().stream()
        .map(material -> materialMapper.toDto(material))
        .collect(Collectors.toList());
```

which looks free. `IndentItemMapper` declares `uses = MaterialMapper.class` and `IndentItemDto` carries a `MaterialDto`, so an indent pays the same cost once per line item, and `IndentMapper` then pays it once per line on every indent on the page.

Four assertions, all currently passing:

| Test | Asserts |
|---|---|
| `mappingOneMaterialCostsTwoAggregateReads` | 1 material, 2 aggregate reads |
| `mappingAPageOfFiftyMaterialsCostsOneHundredAggregateReads` | 50 materials, 100 aggregate reads |
| `theCostFollowsMaterialDtoIntoAnIndentLine` | 10 indent lines, 20 aggregate reads |
| `theWithStockVariantTakesTheAggregatesFromTheCaller` | `toWithStockDto`, 0 aggregate reads |

The last one is the point of the set. The batched shape already exists on this mapper and three call sites in `MaterialService` use it. `toDto` is the one that does not.

## Why a test rather than the fix

The fix belongs to whoever owns these files next, and right now three open branches have them open: #473 is editing `MaterialService`, `StorageLocationService` and `IndentService`, #529 is editing `InventoryService`, and #530 is editing `IndentService`. Changing `MaterialMapper` now would conflict with all three and would be reviewed as part of an argument it is not part of. The counts are the durable half; they can land alone and they are what the eventual fix flips.

`default_batch_fetch_size: 100` does not help here. It batches lazy association loads, and these are explicit repository calls.

## Cost

Plain JUnit and Mockito, no Spring context, so it adds nothing to the context cache the suite is already tight on. Runs in under a second.

```
MaterialMapperCostTest > mappingAPageOfFiftyMaterialsCostsOneHundredAggregateReads() PASSED
MaterialMapperCostTest > theWithStockVariantTakesTheAggregatesFromTheCaller() PASSED
MaterialMapperCostTest > mappingOneMaterialCostsTwoAggregateReads() PASSED
MaterialMapperCostTest > theCostFollowsMaterialDtoIntoAnIndentLine() PASSED
```

Do not merge yet. This is evidence attached to an open design question, and it should land with whatever #522 concludes.

### DIFF
--- a/src/test/java/org/tornotron/echno_backend/material/mapper/MaterialMapperCostTest.java
+++ b/src/test/java/org/tornotron/echno_backend/material/mapper/MaterialMapperCostTest.java
@@ -1,0 +1,119 @@
+package org.tornotron.echno_backend.material.mapper;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.test.util.ReflectionTestUtils;
+import org.tornotron.echno_backend.employee.mapper.EmployeeMapper;
+import org.tornotron.echno_backend.indentItem.IndentItem;
+import org.tornotron.echno_backend.indentItem.mapper.IndentItemMapper;
+import org.tornotron.echno_backend.indentItem.mapper.IndentItemMapperImpl;
+import org.tornotron.echno_backend.inventoryTransaction.InventoryService;
+import org.tornotron.echno_backend.material.Material;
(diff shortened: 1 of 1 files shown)


abhijithanandan:
Folded into #561 and superseded by it. The commit here is the first of the three on that branch, so the measurement lands as written and the next commit flips the assertions to what the same paths cost after item A: one aggregate read for a material, one for a page of fifty, one for a ten-line indent. That is what this PR said should happen when the fix landed, and it is visible in the commit history there rather than only in a query log.

PR #539, opened 2026-08-28
Check consumption against the balance row it writes, and give stock adjustments a way to correct one

### LINKED ISSUE #530: Site transfer and PO numbers are invented by the client, the unique constraint is global while the check is per-org, and the duplicate message is discarded

### LINKED ISSUE #531: Purchase orders can be created already APPROVED, and unitPrice is unvalidated so POs save with a zero total
## Reported

Anand, QA pass on `echno.in`, 28 Aug 2026. ClickUp `86d45vnx9`.

Two findings from the same screen, both separate from the duplicate PO number in #BACKEND_NUMBERING:

- A Purchase Order can be created with Status = Approved straight from the create form, bypassing approval entirely.
- Unit Price defaults to ₹0.00 on every line, the Line Total shows "—", and the PO saves with a total of zero.

### Status is client-supplied and unrestricted

`purchaseOrder/dto/PurchaseOrderCreationDto.java:31-33`:

```java
@NotBlank
private String status;
```

No enum type, no allowed-value constraint. It is resolved blindly at `PurchaseOrderService.java:118`:

```java
PurchaseOrderStatus.valueOf(creationDto.getStatus())
```

So any of the six enum values is accepted on create, including `APPROVED`, `FULLY_RECEIVED` and `CANCELLED`. Nothing checks whether the caller holds approval authority.

The frontend exposes it rather than hiding it: `features/purchase-orders/components/purchase-order-form.tsx` renders a Status select on the create form that enumerates the whole enum unfiltered, and `purchase-orders/new/page.tsx:90` sends whatever is chosen. The default is `draft`, so this needs a deliberate click, but nothing stops it.

This is an authorisation gap, not a mapping bug. The enums line up correctly across `echno-core/src/types/purchase-orders/enums.ts:12-30` and `PurchaseOrderStatus.java`.

### Unit price is unvalidated on both tiers

`purchaseOrderItem/dto/PurchaseOrderItemCreationDto.java:29-30` has no `@NotNull` and no `@Positive` on `unitPrice`. Compare `orderedQuantity` at lines 25-27, which has both. Null is then coerced to zero at `PurchaseOrderService.java:309-310` and `PurchaseOrderItemService.java:79-80`.

The frontend does not check it either: `validateForm()` in `purchase-order-form.tsx:284-299` validates `materialId` and `orderedQuantity` only.

The "—" Anand saw is cosmetic and follows from the zero. The line total cell is `item.unitPrice ? … : '—'`, and `0` is falsy, so the cell blanks and the grand total block (guarded on `totalAmount > 0`) is hidden entirely. The PO then persists with `totalAmount = 0`, which is the part that matters.

## What a fix involves

1. Type `status` as the enum in the creation DTO, and constrain create to the initial states. Approval moves to the existing `PATCH /{id}/status` path, gated on a role.
2. On the web side, drop the status select from the create form. Nothing good comes of offering it.
3. Add `@NotNull` and `@PositiveOrZero` to `unitPrice`, and decide whether a genuinely zero-priced line is allowed. If it is, it should be an explicit choice on the line rather than the default that falls out of an empty field.
4. Mirror the price rule in `validateForm()` so the user finds out before submitting.

Anand also asks for the PO to inherit project and material from a linked indent, and for the ordered quantities to be validated against it. That is real and worth doing, but it is a feature on top of this rather than part of the fix, so I have left it out.

abhijithanandan:
Closes #530
Closes #531

## What was wrong

Two issues from Anand's 28 Aug QA pass on `echno.in`, both on the procurement screens.

**#530.** The browser invented `transferNumber` and `poNumber` and sent them; the server only checked what it was handed. So every create after the first collided on `000001`, and no amount of frontend cleverness could fix it, because two tabs compute the same next number from the same list. On top of that the unique constraint was **global** while every service check was **per organization** (harmless with one tenant, a bug the second client triggers on their first document), and `GlobalExceptionHandler` threw `ex.getMessage()` away, so `PO-2026-000001 already exists` reached the user as "Data already exists". That last part is why this was first reported as a double-submit bug.

**#531.** `status` on the purchase order create payload was a free `String` resolved with `valueOf`, so a PO could be created already `APPROVED` and approval was bypassed entirely. `unitPrice` had no `@NotNull` and no `@Positive*`, was coerced from null to zero, and orders saved with a total of nothing.

## Allocation, and why it is correct under concurrency

Numbers now come from a counter row per `(organization, document type, year)` in a new `document_number_sequence` table. `DocumentNumberAllocator` takes one:

```sql
INSERT INTO document_number_sequence
    (organization_id, document_type, sequence_year, last_allocated, created_at, updated_at)
VALUES (?, ?, ?, 1, current_timestamp, current_timestamp)
ON CONFLICT (organization_id, document_type, sequence_year)
DO UPDATE SET last_allocated = document_number_sequence.last_allocated + 1,
              updated_at = current_timestamp
RETURNING last_allocated
```

**One statement is the whole point.** Read-then-write is wrong however it is ordered, because two transactions can read the same counter before either writes, and `max + 1` has the same fault with an extra table scan. As a single statement the counter row *is* the point of conflict:

- The first transaction to touch it leaves a write intent. The second blocks on that intent rather than reading around it, and only proceeds once the first has committed or aborted, incrementing whatever the first left.
- If waiting pushes the second's timestamp past something it has already read, CockroachDB aborts it with SQLSTATE `40001` rather than let it commit on a stale snapshot. `TransactionRetryTemplate` then runs the whole unit of work again, and the fresh attempt allocates the number after the winner's.

Two committed transactions cannot come away with the same number: that would need both to have read the same counter value and both to have committed, which is what SERIALIZABLE forbids.

The four create paths also nominate `23505` as retryable, following #528's pattern, so a number that somehow did collide (a counter seeded behind the rows already in the table, say) is re-allocated on the next attempt instead of surfacing as a 500 for a duplicate the user never chose. That is a backstop, not the mechanism.

**The trade.** Allocation runs inside the create transaction, so the number is spent only if the create commits and a rolled back create leaves no gap. The cost is that the counter row stays locked until commit, which serialises concurrent creates of one document type within one tenant. At the rate these documents are raised that is the right side of the trade, and it is why the allocation is taken as late in the create flow as it can be.

**Proven, not argued.** `DocumentNumberAllocatorConcurrencyIT` runs eight threads at once against a real CockroachDB and asserts they come away with exactly `000001` through `000008`. Swapped for a read-then-write implementation it fails immediately:

```
DuplicateKeyException: duplicate key value violates unique constraint "uk_document_number_sequence"
  Detail: Key (organization_id, document_type, sequence_year)=(3, 'PURCHASE_ORDER', 2026) already exists.
```

## Four of four, not two of four

`generateIndentNumber` and `generateGrnNumber` have exactly the shape of the two that broke, so indents and GRNs are allocated server-side too rather than left to break next. `lib/utils/document-number-utils.ts` can go entirely.

While there: indent create never checked its number for a duplicate at all, and GRN create checked `existsByGrnNumber` with no organization at all, so one tenant's GRN number blocked another's. Both are moot now that the server issues the number, and the constraint underneath each is per-tenant.

## API contract for the frontend unwind

This is what echno-web PR #312's four steps should be written against.

**1. Stop sending the number.** `transferNumber`, `poNumber`, `indentNumber` and `grnNumber` are **gone from the four creation DTOs**. Unknown properties are ignored on deserialization, so the currently deployed frontend keeps working unchanged until it is updated; the field is simply discarded. Nothing needs to be coordinated on deploy order.

| Endpoint | Field removed from the request |
|---|---|
| `POST /api/v1/site-transfers` (and `/web`) | `transferNumber` |
| `POST /api/v1/purchase-orders` (and `/web`) | `poNumber` |
| `POST /api/v1/indents` (and `/web`) | `indentNumber` |
| `POST /api/v1/goods-received-notes` (and `/web`) | `grnNumber` |

**2. Show it from the response.** The number is on the created resource exactly as before: `SiteTransferDto.transferNumber`, `PurchaseOrderDto.poNumber`, `IndentDto.indentNumber`, `GoodsReceivedNoteDto.grnNumber`. Nothing on the read side changed. Format is unchanged too, `PREFIX-YYYY-NNNNNN` with the same `TRF` / `PO` / `IND` / `GRN` prefixes and the same six-digit padding, so a "will be assigned on save" placeholder followed by the real value from the 201 body is all that is needed.

**3. Drop the client `required()` check.** There is nothing left for the user to supply. The PO number field was editable; it should become read-only or disappear, since a typed number is now ignored rather than honoured.

**4. Duplicate messages now arrive intact.** A 409 from `DuplicateResourceException` carries the service's own sentence in `detail` (and in the legacy `message` property), for example `Purchase Order with PO number PO-2026-000001 already exists`. The generic string survives only as a fallback for an exception with no message.

**Also for #531, on the same screens:**

- `status` on `POST /purchase-orders` is now the `PurchaseOrderStatus` enum, **optional**, and **`DRAFT` is the only accepted value**. Omitting it means `DRAFT`. Anything else is a 400 naming the status that was refused. The status select should come off the create form; approval goes through `PATCH /purchase-orders/{id}/status`, which already exists.
- `unitPrice` on a purchase order line is now **required** and must be **zero or positive**. Zero is still allowed, for a line genuinely supplied free of charge, but it has to be sent deliberately rather than fall out of an empty field. `validateForm()` should mirror that so the user finds out before submitting.
- `totalAmount` on the create payload and `totalPrice` on a line are ignored, and always were: the server computes both. Their schema descriptions now say so.

## Schema

`db/changelog/v4.0/061-document-number-sequences.xml`, four changesets:

1. Create `document_number_sequence` with a unique `(organization_id, document_type, sequence_year)` and an FK to `organization`.
2. **Seed it above the numbers already in each table**, so the first server-issued number cannot collide with a browser-issued one. Only rows in the canonical `PREFIX-YYYY-NNNNNN` shape are counted, which is exactly the right set: a number in any other shape can never be produced by the allocator, so it cannot be collided with. The year is read out of the number rather than off `created_at`, because the number is what has to stay unique. Positions are fixed rather than a capture group, since CockroachDB's `substring` with a regex returns the whole match rather than the group.
3. Add `(organization_id, <number>)` unique constraints on all four tables.
4. Drop the global ones. `DROP INDEX ... CASCADE` rather than `DROP CONSTRAINT`, because CockroachDB backs a `UNIQUE` column constraint with an index of the same name. `uk_intend_number` is dropped alongside `uk_indent_number`: that is the name the constraint carries on databases built from the v1, v2 or v3 baselines, before the intend-to-indent rename.

New constraints are added before the old ones are dropped, so the numbers are never unconstrained in between. The whole file is exercised on a real CockroachDB v26.2.4 by every integration test in the suite.

Also `echno.document-number.zone` (default `Asia/Kolkata`), because the year stamped on a document should not flip a day early on a server running UTC.

## Tests

Every new test was run against the fix removed. Nothing here passes vacuously.

| Test | Fix removed | Result without it |
|---|---|---|
| `DocumentNumberAllocatorConcurrencyIT.concurrentAllocations_eachGetTheirOwnNumber` | single upsert replaced by read-then-write | **FAILED** &mdash; `DuplicateKeyException` on `uk_document_number_sequence`, two callers computed the same next value |
| `DocumentNumberAllocatorTest.allocate_readsAndWritesTheCounterInASingleUpsert` | same | **FAILED** |
| `DocumentNumberAllocatorTest.allocate_formatsThePrefixYearAndPaddedSequence` | same | **FAILED** |
| `DocumentNumberAllocatorTest.allocate_padsToSixDigitsButDoesNotTruncateBeyondThem` | same | **FAILED** |
| `DocumentNumberAllocatorTest.allocate_usesEachTypesOwnPrefix` | same | **FAILED** |
| `DocumentNumberAllocatorTest.allocate_readsTheYearInTheConfiguredZoneNotTheContainers` | same | **FAILED** |
| `PurchaseOrderServiceTest.createPurchaseOrder_takesItsNumberFromTheServerNotTheCaller` | allocator call replaced by a literal | **FAILED** |
| `SiteTransferServiceTest.create_takesItsNumberFromTheServerNotTheCaller` | same | **FAILED** |
| `IndentServiceTest.addIndent_takesItsNumberFromTheServerNotTheCaller` | same | **FAILED** |
| `GoodsReceivedNoteServiceTest.create_takesItsNumberFromTheServerNotTheCaller` | same | **FAILED** |
| `GoodsReceivedNoteServiceTest.create_buildsItemsSavesThemAndPublishesEvent` | same | **FAILED** (it asserts the published number) |
| `PurchaseOrderServiceTest.createPurchaseOrder_asApproved_isRefused` | guard removed, `setStatus(dto.getStatus())` restored | **FAILED** |
| `PurchaseOrderServiceTest.createPurchaseOrder_asAnyOtherLaterState_isRefused` | same | **FAILED** |
| `PurchaseOrderServiceTest.createPurchaseOrder_withNoStatus_startsAsDraft` | same | **FAILED** |
| `PurchaseOrderItemCreationDtoValidationTest.missingUnitPrice_isRejected` | `@NotNull` / `@PositiveOrZero` removed | **FAILED** |
| `PurchaseOrderItemCreationDtoValidationTest.negativeUnitPrice_isRejected` | same | **FAILED** |
| `GlobalExceptionHandlerTest.duplicateResource_reportsTheDuplicateItActuallyFound` | handler restored to the fixed string | **FAILED** |

Two tests pass under those reverts and are honestly not part of the red demonstration: `DocumentNumberAllocatorConcurrencyIT.countersAreSeparatePerOrganizationAndPerType`, which covers the per-tenant scoping rather than the concurrency, and `GlobalExceptionHandlerTest.duplicateResource_withNoMessage_stillSaysSomethingUseful`, which pins the fallback.

Full suite green on the lab box against real CockroachDB.

## Not addressed

`existsByPoNumberAndOrganization_Id` and `existsByTransferNumberAndOrganization_Id` are now unused. They are left in place deliberately: #473 is in flight across the repository interfaces and a one-line deletion there is not worth the conflict. Worth sweeping once that lands.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/events/listeners/InventoryEventListener.java
+++ b/src/main/java/org/tornotron/echno_backend/common/events/listeners/InventoryEventListener.java
@@ -88,13 +88,17 @@ public class InventoryEventListener {
         MaterialConsumption consumption = event.getMaterialConsumption();
         logger.info("Handling material consumed event for consumption ID: {}", consumption.getId());
 
+        // Read the balance this movement actually draws down. With no storage location that
+        // is the project's unlocated row, not the project total: the total would put an
+        // opening and closing figure on the ledger entry that the row it moves never held.
         Double openingStock;
         if (consumption.getStorageLocation() != null) {
             openingStock = inventoryService.getStockAtLocation(
                     consumption.getMaterial().getId(), consumption.getProject().getId(),
                     consumption.getStorageLocation().getId());
         } else {
-            openingStock = inventoryService.getCurrentStock(consumption.getMaterial().getId(), consumption.getProject().getId());
+            openingStock = inventoryService.findUnlocatedStock(
+                    consumption.getMaterial().getId(), consumption.getProject().getId()).orElse(0.0);
         }
         Double quantityChanged = -consumption.getQuantity().doubleValue();
         Double closingStock = openingStock + quantityChanged;
--- a/src/main/java/org/tornotron/echno_backend/inventoryTransaction/InventoryService.java
+++ b/src/main/java/org/tornotron/echno_backend/inventoryTransaction/InventoryService.java
@@ -73,6 +73,10 @@ public class InventoryService {
     /**
      * Returns current stock for a material at one storage location within a project.
      *
+     * <p>Flattens a missing balance row to zero. Use {@link #findStockAtLocation} where the
+     * difference matters: a material that has never been stocked at a location and a location
+     * that has genuinely run out are different situations and deserve different messages.
+     *
      * @param materialId The material to look up.
      * @param projectId The project the location belongs to.
      * @param storageLocationId The storage location to read.
@@ -80,10 +84,43 @@ public class InventoryService {
      */
     @Transactional(readOnly = true)
     public Double getStockAtLocation(Long materialId, Long projectId, Long storageLocationId) {
+        return findStockAtLocation(materialId, projectId, storageLocationId).orElse(0.0);
+    }
+
+    /**
+     * Returns the balance for a material at one storage location, empty when no row exists.
+     *
+     * <p>The empty result means "this material has never been stocked here", which reads the
+     * same as a zero balance once flattened but is a different fact about the data.
+     *
+     * @param materialId The material to look up.
+     * @param projectId The project the location belongs to.
+     * @param storageLocationId The storage location to read.
+     * @return The quantity on hand, or empty when the balance row does not exist.
+     */
+    @Transactional(readOnly = true)
+    public Optional<Double> findStockAtLocation(Long materialId, Long projectId, Long storageLocationId) {
         return currentStockRepository
                 .findByMaterialIdAndProjectIdAndStorageLocationId(materialId, projectId, storageLocationId)
-                .map(CurrentStock::getCurrentQuantity)
-                .orElse(0.0);
+                .map(CurrentStock::getCurrentQuantity);
+    }
+
+    /**
+     * Returns the project balance held against no storage location, empty when no row exists.
+     *
+     * <p>This is the row a movement with no storage location reads and writes. It is not the
+     * project total: stock booked into a storage location lives in its own row and is not
+     * reachable from here.
+     *
+     * @param materialId The material to look up.
+     * @param projectId The project to read.
+     * @return The unlocated quantity on hand, or empty when the balance row does not exist.
+     */
+    @Transactional(readOnly = true)
+    public Optional<Double> findUnlocatedStock(Long materialId, Long projectId) {
+        return currentStockRepository
+                .findByMaterialIdAndProjectIdAndStorageLocationIsNull(materialId, projectId)
+                .map(CurrentStock::getCurrentQuantity);
     }
 
     /**
@@ -123,18 +160,64 @@ public class InventoryService {
     /**
      * Checks that a storage location holds at least the required quantity of a material.
      *
+     * <p>A balance row that does not exist is reported separately from a row that has run
+     * down to zero. Both leave nothing to draw on, but only the second is a stock-out; the
+     * first means the material has never been received into that location, which is a
+     * different thing to tell the user and a different thing for them to fix.
+     *
      * @param materialId The material to check.
      * @param projectId The project the location belongs to.
      * @param storageLocationId The storage location to check.
      * @param requiredQuantity The quantity that must be available.
-     * @throws InsufficientStockException if the quantity on hand at the location is below the required amount.
+     * @throws InsufficientStockException if no balance exists at the location, or the quantity on hand is below the required amount.
      */
     public void validateSufficientStockAtLocation(Long materialId, Long projectId, Long storageLocationId, Double requiredQuantity) {
-        Double currentStock = getStockAtLocation(materialId, projectId, storageLocationId);
+        Optional<Double> balance = findStockAtLocation(materialId, projectId, storageLocationId);
+        if (balance.isEmpty() && requiredQuantity > 0) {
+            throw new InsufficientStockException(
+                String.format("No stock of material ID %d has ever been held at project ID %d, storage location ID %d, "
+                        + "so there is none to draw on. Required: %.2f. Receive the material into this location with a "
+                        + "goods receipt or a transfer in, or record a stock adjustment, before drawing from it.",
+                    materialId, projectId, storageLocationId, requiredQuantity)
+            );
+        }
+        Double currentStock = balance.orElse(0.0);
         if (currentStock < requiredQuantity) {
             throw new InsufficientStockException(
                 String.format("Insufficient stock for material ID %d at project ID %d, storage location ID %d. Required: %.2f, Available: %.2f",
                     materialId, projectId, storageLocationId, requiredQuantity, currentStock)
+            );
+        }
+    }
+
+    /**
+     * Checks the project balance held against no storage location.
+     *
+     * <p>This is the check for a movement that names no location, because that is the row the
+     * movement goes on to write. {@link #validateSufficientStock} is deliberately not used
+     * here: it sums every row on the project, so it passes on stock sitting in storage
+     * locations that the unlocated write will never touch, and the balance goes negative.
+     *
+     * @param materialId The material to check.
+     * @param projectId The project to check within.
+     * @param requiredQuantity The quantity that must be available.
+     * @throws InsufficientStockException if no unlocated balance exists, or it is below the required amount.
+     */
+    public void validateSufficientUnlocatedStock(Long materialId, Long projectId, Double requiredQuantity) {
+        Optional<Double> balance = findUnlocatedStock(materialId, projectId);
+        if (balance.isEmpty() && requiredQuantity > 0) {
+            throw new InsufficientStockException(
+                String.format("Material ID %d holds no stock at project ID %d outside a storage location. Required: %.2f. "
+                        + "Any stock the project holds sits in its storage locations, so name the location to draw from.",
+                    materialId, projectId, requiredQuantity)
+            );
+        }
+        Double currentStock = balance.orElse(0.0);
+        if (currentStock < requiredQuantity) {
+            throw new InsufficientStockException(
+                String.format("Insufficient stock for material ID %d at project ID %d outside a storage location. "
+                        + "Required: %.2f, Available: %.2f",
+                    materialId, projectId, requiredQuantity, currentStock)
             );
         }
     }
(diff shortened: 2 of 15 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #540, opened 2026-08-28
Allocate document numbers on the server, and scope their uniqueness to the tenant

### LINKED ISSUE #482: Compliance generation runs the model inside the request and inside a transaction, so it cannot scale past ~20 rules

abhijithanandan:
Closes #482.

#514 took the model call out of the transaction and #528 made a cut-short answer fail loudly instead of writing nothing. Both left the run itself on the request thread, where it dies at the sixty-second edge timeout no matter how well it behaves. This moves it off, and splits the rule catalogue so no single call has to carry all of it.

## The token ceiling, measured

This was recorded as blocked on a key. It was not: the key is configured on staging, and the numbers below come from real calls to the configured endpoint (`llama-4-maverick` on DigitalOcean Gradient) through the lab proxy, using a prompt built exactly the way `buildUserPrompt` builds it, over the seeded rule catalogue extended to the sizes shown.

| Rules | prompt tokens | completion tokens | tokens/rule | elapsed | finish_reason | elements parsed |
|---|---|---|---|---|---|---|
| 6 | 695 | 339 | 56.5 | 13.2 s | stop | 6 |
| 8 | 825 | 451 | 56.4 | 20.3 s | stop | 8 |
| 10 | 958 | 573 | 57.3 | 17.2 s | stop | 10 |
| 12 | 1096 | 701 | 58.4 | 23.4 s | stop | 12 |
| 21 | 1710 | 1209 | 57.6 | 38.4 s | stop | 21 |
| 40 | 3048 | 2376 | 59.4 | 58.3 s | stop | 40 |
| 60 | 4444 | 3187 | 53.1 | 108.5 s | stop | 60 |
| **72** | 5286 | **4096 (capped)** | - | 257.7 s | **length** | **none: the array never closes** |

**About 57 completion tokens per rule.** The 72-rule run is the ceiling landing exactly where that predicts: `finish_reason: length`, the budget spent to the token, and a response whose array never closes, which is precisely the second of #528's three truncation checks firing on live data. Two things worth noting beyond the headline:

- **Wall clock is the tighter limit, and it bites first.** At about 1.7 s per rule, 40 rules already takes 58 s. So the sixty-second ceiling arrives at roughly 40 rules and the token ceiling at roughly 72. The issue estimated both at 30 to 35; the shape of the argument was right and the numbers were pessimistic.
- **The model gets terser as it approaches the cap** (53.1 tokens/rule at 60 against 57 to 59 below it), so answer quality degrades before truncation does. That is a further argument for a batch size nowhere near the cap.

**Batch size: 10.** Ten rules costs about 600 completion tokens, under a sixth of the 4096 budget, and takes about 20 seconds against a 60-second read timeout. Twenty would still fit the token budget but would sit at roughly two thirds of the read timeout, which is not enough margin for one slow call. Both numbers are configuration (`compliance.ai.batch-size`), because the figure that has to hold is not the average rule but the worst rule in someone's catalogue.

**The batches run concurrently, also on evidence.** The issue asked whether chunked calls serialise per API key. They do not: four simultaneous ten-rule calls finished in 22 s where the same four run one after another took 91. So a 70-rule catalogue costs roughly one batch of wall clock rather than seven. `compliance.ai.batch-concurrency` defaults to 4, and setting it to 1 is the right move for an endpoint that rate-limits per key.

## The queue

New table `compliance_generation_jobs` (changelog 061). `POST /compliance/jobs?projectId=` inserts a row and answers with it; `GET /compliance/jobs/{id}` is the poll; `GET /compliance/jobs?projectId=` returns the latest run for a project, which is how a page that was reloaded mid-run finds what it was watching.

**Status model.** Five states, because a caller has five things to show and no fewer:

| status | means |
|---|---|
| `queued` | accepted, waiting for a worker |
| `running` | a worker has it; the progress counters move here |
| `succeeded` | every rule assessed, compliances created |
| `nothing-to-report` | every rule assessed, nothing to create (none applied, or all already existed) |
| `failed` | did not cover every rule; nothing created |

Splitting `succeeded` from `nothing-to-report` is the whole point of the set. Before #528, "the answer was cut short and we lost it" and "there is nothing outstanding for this project" both reached the user as an empty list. A job that recorded both as one success state would have put that ambiguity straight back, one layer further from anyone who could notice. `failed` is the third of the three, and it never carries a count.

**Progress.** `batchesDone/batchesTotal` and `rulesAssessed/rulesTotal`, written after each batch in their own short transaction. They are progress and never a partial result: a run that stops at batch three has assessed some rules and produced nothing.

**Preconditions stay at accept time.** No project, no project type, no recognisable state, no rules for the jurisdiction, AI not configured: all decidable from the database and configuration, so they are raised on the request as the 400 or 404 they have always been. Only failures that genuinely need the model come back later as a failed job.

**One job per project.** A partial unique index on `(organization_id, project_id) WHERE status IN ('QUEUED','RUNNING')`. A second click joins the run in flight rather than starting a second one, and the loser of an insert race reads back the winner and hands that to its caller. This is #510's race closed one step earlier, where it matters more, because a duplicate run spends a minute of inference before it gets far enough to collide.

**Which replica.** Every replica polls every two seconds; a claim is `UPDATE ... SET status='RUNNING' WHERE id=? AND status='QUEUED'` and only one can win. No leader (a replica that wrongly believes it is one is a failure mode this design cannot have), no advisory locks (CockroachDB has none), no broker (Redis is optional and off by default here, so a Redis-only queue would be a second configuration axis for every environment). The claim writes a lease that the worker pushes forward as it finishes each batch, so progress and liveness are the same signal and cannot get out of step. A dead replica's lease goes stale and the next poll on any replica returns the row to the queue, or fails it if its attempts are gone. The attempt is counted by the claim, so a job that kills its worker before the worker writes anything still stops eventually.

**Tenancy.** The poller cannot have a tenant, since its job is to find out which tenant to become, so its queries are native and return bare scalars, never entities: a Hibernate `@Filter` does not apply to native SQL, so the intent is in the query rather than in ambient thread state a later edit could widen. Everything after the claim runs inside `TenantScopedJobRunner`, which refuses a null org id rather than running unscoped.

**Approval queues too.** `ComplianceGenerationListener` inserts a job instead of running generation on an `@Async` thread. That thread was watched by nothing and outlived by nothing: a replica restarted mid-run left an approved project with no compliances and no record that generation had ever been attempted.

## Partial batch failure

**A failed batch fails the whole run and creates nothing.** Batch 3 of 5 failing means the run has no opinion at all about the rules in batches 3 to 5. Handing back batches 1 and 2 would create their compliances and show the user a finished result silently missing three fifths of the jurisdiction, which is the exact failure this module exists to prevent, and is #528's rule applied one level up. The job goes to `failed`, or back to `queued` if it has attempts left, with the reason on the row either way so a caller polling through a retry is told what is going wrong. Re-running is cheap and idempotent per rule, which is what makes throwing the work away affordable.

## `POST /compliance/regenerate` is still there

Unchanged, still synchronous, still 200/409/502, so `echno-web` #315 keeps working. It is marked in the OpenAPI description as the shape to move off, and should be deleted once the web client is on the job endpoints. Batching pushes its ceiling out (a 70-rule catalogue now runs in about the time one batch takes) but does not remove it, and it will not survive a big enough catalogue.

## Tests

Every new test was run against the code without its fix to confirm it fails for the reason it claims. See the failing-test table in the review notes.

`ComplianceGenerationServiceIT` gained the job-table cases rather than getting a sibling IT, so the suite keeps one cached Spring context instead of two. Those are the cases only a real CockroachDB can settle: that the partial index predicate holds, that finished jobs do not block a new run, that two claims cannot both win, and that lease recovery moves a row the way it is supposed to. Everything else is plain JUnit and Mockito with no context at all.

## Not done here

- **The web client.** Nothing in `echno-web` uses the job endpoints yet; that is its own change, and until it lands the button still goes through `regenerate`.
- **The nightly sweep.** When Anand's catalogue lands, every already-approved project in an affected jurisdiction is silently missing compliances and nothing regenerates them. That needs a change timestamp on `ComplianceRule` and is a separate piece of work.
- **The hardcoded 60 s read timeout** in `requestFactory()`. It is no longer the binding constraint at a batch size of ten, but it is still a constant where the rest of this is configuration.

## Seed verified against the real staging data

The integration tests exercise the migration on a real CockroachDB v26.2.4, but they start from an empty database, so the seed statements run against zero rows there. The arithmetic that matters is the arithmetic on rows that already exist, so it was checked read-only against the live `echno.in` database:

| Document | org | rows | matching the canonical shape | max sequence | year read |
|---|---|---|---|---|---|
| Purchase order | 2 | 1 | 1 | 1 | 2026 |
| Site transfer | 2 | 1 | 1 | 1 | 2026 |
| Indent | 2 | 5 | 5 | 5 | 2026 |
| GRN | 2 | 5 | 5 | 5 | 2026 |

Every existing row is in the canonical shape, and the offsets pull the year and the sequence out of all of them. The counters therefore start at 1, 1, 5 and 5, and the first numbers the server issues are `PO-2026-000002`, `TRF-2026-000002`, `IND-2026-000006` and `GRN-2026-000006`. None collides with a row already there, which was the one thing that could have made the first create after deploy fail.

## Not mine, and pre-existing: the mapper's aggregate queries

`MaterialMapper.fillStock` runs two aggregate queries per material in an `@AfterMapping` hook, and `IndentItemMapper` declares `uses = MaterialMapper`, so any indent DTO pays two queries per line. The create response built here goes through that mapper, exactly as it did before this change. It is #522's to fix, not this PR's, and the cost is unchanged either way. Noting it so the next person profiling an indent create does not attribute it to the numbering work. Purchase orders are not affected by that particular path: `PurchaseOrderMapper` uses only `EmployeeMapper`.

## Review

No CodeRabbit review ran on this PR. The bot's check reports `pass` with "Review rate limited", and three separate `@coderabbitai review` requests across an hour, including one after the stated reset window, were each refused with the free-OSS quota notice. Merged on green CI plus my own reading, and logged on ClickUp `86d45heqm` for the retro review pass once the quota resets.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/documentnumber/DocumentNumberAllocator.java
+++ b/src/main/java/org/tornotron/echno_backend/common/documentnumber/DocumentNumberAllocator.java
@@ -1,0 +1,100 @@
+package org.tornotron.echno_backend.common.documentnumber;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.stereotype.Component;
+
+import java.time.LocalDate;
+import java.time.ZoneId;
+
+/**
+ * Hands out the next document number for a tenant, a document family and a year.
+ *
+ * <p>Numbers used to be invented by the browser from the list it happened to have loaded, so
+ * two people on the New Purchase Order screen at the same moment proposed the same number and
+ * the second save was rejected. The number is issued here instead, from a counter row per
+ * {@code (organization, document type, year)} in {@code document_number_sequence}.
+ *
+ * <h2>Why one statement</h2>
+ *
+ * <p>The allocation is a single {@code INSERT ... ON CONFLICT ... DO UPDATE ... RETURNING}.
+ * Read-then-write would be wrong here whatever isolation the database ran at, because two
+ * transactions can read the same counter before either writes. As one statement, the row is
+ * the point of conflict:
+ *
+ * <ul>
+ *   <li>The first transaction to touch the counter row leaves a write intent on it. The second
+ *       blocks on that intent rather than reading around it, and only proceeds once the first
+ *       has committed or aborted, at which point it increments the value the first left.</li>
+ *   <li>If waiting pushes the second transaction's timestamp past something it has already
+ *       read, CockroachDB aborts it with SQLSTATE 40001 rather than let it commit on a stale
+ *       snapshot. The whole unit of work then runs again through
+ *       {@link org.tornotron.echno_backend.common.retry.TransactionRetryTemplate}, which
+ *       re-reads the counter and allocates the next number after the winner's.</li>
+ * </ul>
+ *
+ * <p>Either way, two committed transactions cannot come away with the same number: that would
+ * require both to have read the same counter value and both to have committed, which is exactly
+ * what SERIALIZABLE forbids.
+ *
+ * <h2>What the caller still owes</h2>
+ *
+ * <p>This runs in the caller's transaction, so the counter advances only if the create commits;
+ * a rolled back create returns its number to the pool rather than leaving a gap. The cost is
+ * that the counter row stays locked until the caller commits, which serialises concurrent
+ * creates of the same document type within one tenant. That is the intended trade at the volume
+ * these documents are raised, and it is why the lock is taken as late as the create flow allows.
+ *
+ * <p>The per-organization unique index on the number column is the backstop, not the mechanism.
+ * If a number ever were issued twice (a counter seeded behind the rows already in the table, say)
+ * the insert is rejected with SQLSTATE 23505, and the callers nominate that as a retryable
+ * conflict so the next attempt allocates a fresh number instead of answering the user with a 500.
+ */
+@Component
+public class DocumentNumberAllocator {
+
+    /**
+     * Allocates and formats in one round trip. {@code last_allocated} holds the highest number
+     * issued so far, so a first allocation inserts 1 and every later one returns the value it
+     * has just written.
+     */
+    private static final String ALLOCATE_SQL = """
+            INSERT INTO document_number_sequence
+                (organization_id, document_type, sequence_year, last_allocated, created_at, updated_at)
+            VALUES (?, ?, ?, 1, current_timestamp, current_timestamp)
+            ON CONFLICT (organization_id, document_type, sequence_year)
+            DO UPDATE SET last_allocated = document_number_sequence.last_allocated + 1,
+                          updated_at = current_timestamp
+            RETURNING last_allocated
+            """;
+
+    /** Matches the six-digit padding the browser used, so old and new numbers sort together. */
+    private static final String NUMBER_FORMAT = "%s-%d-%06d";
+
+    private final JdbcTemplate jdbcTemplate;
+    private final ZoneId zone;
+
+    public DocumentNumberAllocator(
+            JdbcTemplate jdbcTemplate,
+            @Value("${echno.document-number.zone:Asia/Kolkata}") String zone) {
+        this.jdbcTemplate = jdbcTemplate;
+        this.zone = ZoneId.of(zone);
+    }
+
+    /**
+     * Issues the next number of {@code type} for {@code organizationId} in the current year.
+     *
+     * <p>Must be called inside a transaction, and inside the same one that saves the document:
+     * the number is only spent if that transaction commits.
+     *
+     * @param type           the document family, which decides the prefix and the counter row
+     * @param organizationId the tenant the counter belongs to
+     * @return the formatted number, for example {@code PO-2026-000007}
+     */
+    public String allocate(DocumentNumberType type, Long organizationId) {
+        int year = LocalDate.now(zone).getYear();
+        Long sequence = jdbcTemplate.queryForObject(
+                ALLOCATE_SQL, Long.class, organizationId, type.name(), year);
+        return NUMBER_FORMAT.formatted(type.getPrefix(), year, sequence);
+    }
+}
--- a/src/main/java/org/tornotron/echno_backend/common/documentnumber/DocumentNumberType.java
+++ b/src/main/java/org/tornotron/echno_backend/common/documentnumber/DocumentNumberType.java
@@ -1,0 +1,36 @@
+package org.tornotron.echno_backend.common.documentnumber;
+
+/**
+ * The document families that carry a server-allocated number, and the prefix each one wears.
+ *
+ * <p>The name of the constant is what goes into the {@code document_type} column of
+ * {@code document_number_sequence}, so renaming one orphans that tenant's counter and starts
+ * its numbering again from one. Add constants, do not rename them.
+ *
+ * <p>The prefixes are the ones the browser used to generate, kept identical so numbers issued
+ * before and after the move to server-side allocation sort and read the same way.
+ */
+public enum DocumentNumberType {
+
+    /** Site transfers, {@code TRF-2026-000001}. */
+    SITE_TRANSFER("TRF"),
+
+    /** Purchase orders, {@code PO-2026-000001}. */
+    PURCHASE_ORDER("PO"),
+
+    /** Material indents, {@code IND-2026-000001}. */
+    INDENT("IND"),
+
+    /** Goods received notes, {@code GRN-2026-000001}. */
+    GOODS_RECEIVED_NOTE("GRN");
+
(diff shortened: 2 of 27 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Surveyed the rest of the create endpoints for the same client-supplied-status weakness while in here, since #531 was unlikely to be the only one. It is not: sixteen others let the caller choose the lifecycle state on create, and two of them matter beyond hygiene.

**Project** (`ProjectCreationDto.java:60` → `ProjectService.java:113`) takes `status: "approved"` on create and skips both halves of the approval contract the patch path enforces: `requireStateForApproval` and the `ProjectApprovedEvent` publish that generates the project's compliance inspections. The project ends up permanently approved with its compliance inspections never generated, and no later transition can re-fire the event because `onApproval` returns early once the status is already approved.

**Issue** (`IssueCreationDto.java:39` → `IssueService.java:87`) is a genuine escalation: any tenant member can create an issue, but moving one to `resolved`/`closed` afterwards needs `system-admin` or `project-manager`. Create lets a member land it in that terminal state directly.

Below those: StockAdjustment accepts `"APPROVED"` as free text while `approvedBy`/`approvedAt` stay null; SiteTransfer accepts `COMPLETED` on create while publishing stock movement unconditionally either way; Expense, Receipt, SubContract and Asset are free strings with no state machine behind them at all.

Full table and ranking in #541. Deliberately not fixed here: this PR is already two issues wide, and Project and Issue each want their own change rather than being folded into a numbering fix.

abhijithanandan:
One more thing for whoever picks up the frontend unwind: `echno-core` carries the number on all four create types and copies it in each mapper, so it has to move first.

- `src/types/purchase-orders/purchase-order-create.ts:18,60` (`poNumber`)
- `src/types/site-transfers/site-transfer-create.ts:20,62` (`transferNumber`)
- `src/types/indents/indent-create.ts:17,50` (`indentNumber`)
- `src/types/grn/grn-create.ts:18,64` (`grnNumber`)

Also in `purchase-order-create.ts`, `status` should become optional and, if kept at all, narrowed to `DRAFT`; the backend refuses anything else on create now.

Removing the field from a create type is a breaking change to the package, so it wants its own `echno-core` release before `echno-web` follows. Nothing is time-critical: the backend already ignores an unknown property, so the deployed frontend keeps working until then.

abhijithanandan:
@coderabbitai review

PR #542, opened 2026-08-28
Run compliance generation as a queued job, a batch of rules at a time

### LINKED ISSUE #482: Compliance generation runs the model inside the request and inside a transaction, so it cannot scale past ~20 rules

abhijithanandan:
Closes #482.

#514 took the model call out of the transaction and #528 made a cut-short answer fail loudly instead of writing nothing. Both left the run itself on the request thread, where it dies at the sixty-second edge timeout no matter how well it behaves. This moves it off, and splits the rule catalogue so no single call has to carry all of it.

## The token ceiling, measured

This was recorded as blocked on a key. It was not: the key is configured on staging, and the numbers below come from real calls to the configured endpoint (`llama-4-maverick` on DigitalOcean Gradient) through the lab proxy, using a prompt built exactly the way `buildUserPrompt` builds it, over the seeded rule catalogue extended to the sizes shown.

| Rules | prompt tokens | completion tokens | tokens/rule | elapsed | finish_reason | elements parsed |
|---|---|---|---|---|---|---|
| 6 | 695 | 339 | 56.5 | 13.2 s | stop | 6 |
| 8 | 825 | 451 | 56.4 | 20.3 s | stop | 8 |
| 10 | 958 | 573 | 57.3 | 17.2 s | stop | 10 |
| 12 | 1096 | 701 | 58.4 | 23.4 s | stop | 12 |
| 21 | 1710 | 1209 | 57.6 | 38.4 s | stop | 21 |
| 40 | 3048 | 2376 | 59.4 | 58.3 s | stop | 40 |
| 60 | 4444 | 3187 | 53.1 | 108.5 s | stop | 60 |
| **72** | 5286 | **4096 (capped)** | - | 257.7 s | **length** | **none: the array never closes** |

**About 57 completion tokens per rule.** The 72-rule run is the ceiling landing exactly where that predicts: `finish_reason: length`, the budget spent to the token, and a response whose array never closes, which is precisely the second of #528's three truncation checks firing on live data. Two things worth noting beyond the headline:

- **Wall clock is the tighter limit, and it bites first.** At about 1.7 s per rule, 40 rules already takes 58 s. So the sixty-second ceiling arrives at roughly 40 rules and the token ceiling at roughly 72. The issue estimated both at 30 to 35; the shape of the argument was right and the numbers were pessimistic.
- **The model gets terser as it approaches the cap** (53.1 tokens/rule at 60 against 57 to 59 below it), so answer quality degrades before truncation does. That is a further argument for a batch size nowhere near the cap.

**Batch size: 10.** Ten rules costs about 600 completion tokens, under a sixth of the 4096 budget, and takes about 20 seconds against a 60-second read timeout. Twenty would still fit the token budget but would sit at roughly two thirds of the read timeout, which is not enough margin for one slow call. Both numbers are configuration (`compliance.ai.batch-size`), because the figure that has to hold is not the average rule but the worst rule in someone's catalogue.

**The batches run concurrently, also on evidence.** The issue asked whether chunked calls serialise per API key. They do not: four simultaneous ten-rule calls finished in 22 s where the same four run one after another took 91. So a 70-rule catalogue costs roughly one batch of wall clock rather than seven. `compliance.ai.batch-concurrency` defaults to 4, and setting it to 1 is the right move for an endpoint that rate-limits per key.

## The queue

New table `compliance_generation_jobs` (changelog 062). `POST /compliance/jobs?projectId=` inserts a row and answers with it; `GET /compliance/jobs/{id}` is the poll; `GET /compliance/jobs?projectId=` returns the latest run for a project, which is how a page that was reloaded mid-run finds what it was watching.

**Status model.** Five states, because a caller has five things to show and no fewer:

| status | means |
|---|---|
| `queued` | accepted, waiting for a worker |
| `running` | a worker has it; the progress counters move here |
| `succeeded` | every rule assessed, compliances created |
| `nothing-to-report` | every rule assessed, nothing to create (none applied, or all already existed) |
| `failed` | did not cover every rule; nothing created |

Splitting `succeeded` from `nothing-to-report` is the whole point of the set. Before #528, "the answer was cut short and we lost it" and "there is nothing outstanding for this project" both reached the user as an empty list. A job that recorded both as one success state would have put that ambiguity straight back, one layer further from anyone who could notice. `failed` is the third of the three, and it never carries a count.

**Progress.** `batchesDone/batchesTotal` and `rulesAssessed/rulesTotal`, written after each batch in their own short transaction. They are progress and never a partial result: a run that stops at batch three has assessed some rules and produced nothing.

**Preconditions stay at accept time.** No project, no project type, no recognisable state, no rules for the jurisdiction, AI not configured: all decidable from the database and configuration, so they are raised on the request as the 400 or 404 they have always been. Only failures that genuinely need the model come back later as a failed job.

**One job per project.** A partial unique index on `(organization_id, project_id) WHERE status IN ('QUEUED','RUNNING')`. A second click joins the run in flight rather than starting a second one, and the loser of an insert race reads back the winner and hands that to its caller. This is #510's race closed one step earlier, where it matters more, because a duplicate run spends a minute of inference before it gets far enough to collide.

**Which replica.** Every replica polls every two seconds; a claim is `UPDATE ... SET status='RUNNING' WHERE id=? AND status='QUEUED'` and only one can win. No leader (a replica that wrongly believes it is one is a failure mode this design cannot have), no advisory locks (CockroachDB has none), no broker (Redis is optional and off by default here, so a Redis-only queue would be a second configuration axis for every environment). The claim writes a lease that the worker pushes forward as it finishes each batch, so progress and liveness are the same signal and cannot get out of step. A dead replica's lease goes stale and the next poll on any replica returns the row to the queue, or fails it if its attempts are gone. The attempt is counted by the claim, so a job that kills its worker before the worker writes anything still stops eventually.

**Tenancy.** The poller cannot have a tenant, since its job is to find out which tenant to become, so its queries are native and return bare scalars, never entities: a Hibernate `@Filter` does not apply to native SQL, so the intent is in the query rather than in ambient thread state a later edit could widen. Everything after the claim runs inside `TenantScopedJobRunner`, which refuses a null org id rather than running unscoped.

**Approval queues too.** `ComplianceGenerationListener` inserts a job instead of running generation on an `@Async` thread. That thread was watched by nothing and outlived by nothing: a replica restarted mid-run left an approved project with no compliances and no record that generation had ever been attempted.

## Partial batch failure

**A failed batch fails the whole run and creates nothing.** Batch 3 of 5 failing means the run has no opinion at all about the rules in batches 3 to 5. Handing back batches 1 and 2 would create their compliances and show the user a finished result silently missing three fifths of the jurisdiction, which is the exact failure this module exists to prevent, and is #528's rule applied one level up. The job goes to `failed`, or back to `queued` if it has attempts left, with the reason on the row either way so a caller polling through a retry is told what is going wrong. Re-running is cheap and idempotent per rule, which is what makes throwing the work away affordable.

## `POST /compliance/regenerate` is still there

Unchanged, still synchronous, still 200/409/502, so `echno-web` #315 keeps working. It is marked in the OpenAPI description as the shape to move off, and should be deleted once the web client is on the job endpoints. Batching pushes its ceiling out (a 70-rule catalogue now runs in about the time one batch takes) but does not remove it, and it will not survive a big enough catalogue.

## Tests

Every new test was run against the code without its fix to confirm it fails for the reason it claims. See the failing-test table in the review notes.

`ComplianceGenerationServiceIT` gained the job-table cases rather than getting a sibling IT, so the suite keeps one cached Spring context instead of two. Those are the cases only a real CockroachDB can settle: that the partial index predicate holds, that finished jobs do not block a new run, that two claims cannot both win, and that lease recovery moves a row the way it is supposed to. Everything else is plain JUnit and Mockito with no context at all.

## Not done here

- **The web client.** Nothing in `echno-web` uses the job endpoints yet; that is its own change, and until it lands the button still goes through `regenerate`.
- **The nightly sweep.** When Anand's catalogue lands, every already-approved project in an affected jurisdiction is silently missing compliances and nothing regenerates them. That needs a change timestamp on `ComplianceRule` and is a separate piece of work.
- **The hardcoded 60 s read timeout** in `requestFactory()`. It is no longer the binding constraint at a batch size of ten, but it is still a constant where the rest of this is configuration.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/events/listeners/ComplianceGenerationListener.java
+++ b/src/main/java/org/tornotron/echno_backend/common/events/listeners/ComplianceGenerationListener.java
@@ -10,14 +10,24 @@ import org.tornotron.echno_backend.common.events.ProjectApprovedEvent;
 import org.tornotron.echno_backend.common.exception.InvalidRequestException;
 import org.tornotron.echno_backend.common.exception.ResourceNotFoundException;
 import org.tornotron.echno_backend.common.multitenancy.TenantScopedJobRunner;
-import org.tornotron.echno_backend.compliance.ComplianceGenerationService;
+import org.tornotron.echno_backend.compliance.job.ComplianceGenerationJobService;
 
 /**
- * Runs AI compliance generation when a project is approved. It fires only AFTER the
+ * Queues AI compliance generation when a project is approved. It fires only AFTER the
  * approving transaction commits (so the project row is durable) and on a separate
- * async thread (so the slow AI call never blocks the approving request). That async
- * thread has no tenant context, so the organization id is read from the event and
- * established for the duration of the work by {@link TenantScopedJobRunner}.
+ * async thread. That async thread has no tenant context, so the organization id is read
+ * from the event and established for the duration of the work by
+ * {@link TenantScopedJobRunner}.
+ *
+ * <h2>Why it queues rather than generates</h2>
+ *
+ * <p>It used to call generation directly, which meant the approval path carried the whole
+ * model call on an {@code @Async} thread. Nothing was watching that thread and nothing
+ * outlived it: a replica restarted mid-run left the approved project with no compliances,
+ * no record that generation had been attempted, and nobody any the wiser. Approval now
+ * inserts a job instead, so the work survives a restart, so a second approval of the same
+ * project cannot start a second run against it, and so the same progress a project manager
+ * sees on the manual path is available for this one.
  *
  * <h2>Why this listener owns no transaction</h2>
  *
@@ -34,18 +44,18 @@ import org.tornotron.echno_backend.compliance.ComplianceGenerationService;
  * have failed open on had it stopped holding.
  *
  * <p>The transaction also spanned the external model call, so the approval path pinned a
- * pool connection for its full 34 to 47 seconds. Being {@code @Async} moved the wait off
+ * pool connection for the model call's full length. Being {@code @Async} moved the wait off
  * the request thread; it did nothing about the connection.
  *
- * <p>So the listener now does what a listener should: establish the tenant, delegate, and
- * let {@link ComplianceGenerationService} own its own transaction boundaries.
+ * <p>So the listener now does what a listener should: establish the tenant and delegate,
+ * leaving the transaction boundaries to whoever owns the work.
  */
 @Slf4j
 @Component
 @RequiredArgsConstructor
 public class ComplianceGenerationListener {
 
-    private final ComplianceGenerationService complianceGenerationService;
+    private final ComplianceGenerationJobService complianceGenerationJobService;
     private final TenantScopedJobRunner tenantScopedJobRunner;
 
     @Async
@@ -59,17 +69,18 @@ public class ComplianceGenerationListener {
             return;
         }
 
-        log.info("Project {} approved; generating compliance inspections for organization {}", projectId, orgId);
+        log.info("Project {} approved; queueing compliance generation for organization {}", projectId, orgId);
         try {
             tenantScopedJobRunner.runForTenant(orgId,
-                    () -> complianceGenerationService.generateForProject(projectId, orgId));
+                    () -> complianceGenerationJobService.submit(projectId, orgId));
         } catch (InvalidRequestException | ResourceNotFoundException e) {
             // Expected on auto-approval: a project approved without a state in its address, or
             // before any rules exist for its jurisdiction, is normal. Log quietly and move on;
-            // a project manager can regenerate once the precondition is met.
+            // a project manager can regenerate once the precondition is met. Those checks now
+            // happen at accept time, so they still surface here rather than inside a worker.
             log.info("Compliance auto-generation skipped for project {}: {}", projectId, e.getMessage());
         } catch (Exception e) {
-            log.error("Compliance generation failed for project {} in organization {}: {}",
+            log.error("Compliance generation could not be queued for project {} in organization {}: {}",
                     projectId, orgId, e.getMessage(), e);
         }
     }
--- a/src/main/java/org/tornotron/echno_backend/compliance/ComplianceGenerationProgress.java
+++ b/src/main/java/org/tornotron/echno_backend/compliance/ComplianceGenerationProgress.java
@@ -1,0 +1,32 @@
+package org.tornotron.echno_backend.compliance;
+
+/**
+ * Told how far a generation run has got, once per completed batch.
+ *
+ * <p>It exists so the slow part of generation can report on itself without knowing anything
+ * about jobs, rows or HTTP. The synchronous endpoint passes {@link #NONE} and the numbers go
+ * nowhere; the queue passes an implementation that writes them to the job row so a polling
+ * caller sees the run advance.
+ *
+ * <p>What it reports is progress and nothing else. A run that stops at batch three of five
+ * has moved this counter three times and still has no result, so nothing downstream may read
+ * these numbers as a partial answer. The result is the return of the generation call, and
+ * there is no such thing as a partial one.
+ */
+@FunctionalInterface
+public interface ComplianceGenerationProgress {
+
+    /** Reports progress that goes nowhere, for callers that are watching the call itself. */
+    ComplianceGenerationProgress NONE = (batchesDone, batchesTotal, rulesAssessed, rulesTotal) -> {
+    };
+
+    /**
+     * One batch of candidate rules has been assessed.
+     *
+     * @param batchesDone   batches finished, 1-based and counting up to {@code batchesTotal}
+     * @param batchesTotal  batches this run was split into
+     * @param rulesAssessed candidate rules covered so far
+     * @param rulesTotal    candidate rules in the run
+     */
+    void batchCompleted(int batchesDone, int batchesTotal, int rulesAssessed, int rulesTotal);
+}
--- a/src/main/java/org/tornotron/echno_backend/compliance/ComplianceGenerationService.java
+++ b/src/main/java/org/tornotron/echno_backend/compliance/ComplianceGenerationService.java
@@ -74,12 +74,23 @@ import java.util.stream.Collectors;
  *
  * <h2>Why this class owns no transaction of its own</h2>
  *
- * <p>The model call takes 34 to 47 seconds today and grows with the size of the curated
- * rule set. Running it inside a transaction pinned one of twenty pool connections for
- * that whole time while doing nothing but waiting on a third party, so a handful of
- * concurrent users could exhaust the pool. {@link #generateForProject} is therefore
- * three phases: read the inputs in one transaction, call the model with no transaction
- * and no connection held, then persist in a second transaction.
+ * <p>The model call costs roughly 1.7 seconds a rule, measured against the configured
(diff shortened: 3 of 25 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
Failing-test evidence for the new tests. Each fix was removed from the production code in turn and the affected class re-run on a lab CockroachDB, to confirm the test fails for the reason it claims rather than passing by accident.

| Fix removed | Tests that then fail |
|---|---|
| `batch()` always returns one batch (no batching) | `splitsTheCatalogueIntoBatchesOfTheConfiguredSize`, `reportsProgressAfterEachBatch`, `progressStopsAtTheFailedBatch`, `oneFailedBatchFailsTheWholeRunAndReturnsNothing`, `aBatchThatComesBackShortIsRefused`, `batchCountMatchesTheNumberOfCallsMade` |
| A failed batch returns the batches that did succeed | `oneFailedBatchFailsTheWholeRunAndReturnsNothing`, `progressStopsAtTheFailedBatch`, `aBatchThatComesBackShortIsRefused` |
| Progress is never reported | `reportsProgressAfterEachBatch`, `progressStopsAtTheFailedBatch` |
| Answers collected in completion order, not batch order | `returnsAnswersInBatchOrderEvenWhenTheCallsFinishOutOfOrder`, `reportsProgressAfterEachBatch` |
| `NOTHING_TO_REPORT` collapsed into `SUCCEEDED` | `aRunThatCoveredEveryRuleAndCreatedNothingIsNothingToReport` |
| Precondition failures retried like any other | `aPreconditionFailureIsTerminalWithoutSpendingTheRemainingAttempts` |
| The still-holds-the-claim guard on terminal writes | `doesNotWriteAnOutcomeOntoAJobItNoLongerHolds` |
| Progress written without extending the lease | `progressWritesCarryTheLeaseAndTheClaimHolder` |
| The claim's row count ignored | `doesNotRunAJobWhoseClaimAnotherReplicaWon` |
| `poll()` lets an exception escape | `aFailedPassDoesNotEscapeAndCancelTheSchedule` |
| No slot accounting | `takesNoMoreWorkThanItHasSlotsFor`, `aWorkerThatThrowsStillGivesItsSlotBack`, `freesTheSlotWhenTheClaimIsLostSoTheNextPassCanUseIt` |
| Slot released only on the happy path | `aWorkerThatThrowsStillGivesItsSlotBack` |
| The active-job check skipped at submit | `aSecondRequestJoinsTheRunAlreadyInFlight` |
| SQLSTATE 23505 not recognised as a lost race | `aRequestThatLosesTheInsertRaceReturnsTheWinnersJob`, `aRequestThatLosesToAJobThatHasSinceFinishedReturnsThatFinishedJob` |
| Preconditions not checked at accept time | 5 of the 6 `ComplianceGenerationJobServiceTest` cases, including `aPreconditionFailureIsRaisedAtAcceptTimeAndQueuesNothing` |
| `AND status='QUEUED'` dropped from the claim | `onlyOneClaimOfTwoCanWin` (against a real CockroachDB) |
| The partial unique index made non-unique | `theDatabaseRefusesASecondJobForAProjectThatAlreadyHasOneInFlight` (against a real CockroachDB) |

Full suite on the branch is green: 87 compliance tests locally on a lab CockroachDB, and CI `build-and-test` passing.

Note on review coverage: CodeRabbit reported `Review rate limited` on this PR and its check went green without a review actually running, so nothing here has been through it yet.

abhijithanandan:
Housekeeping before this merges: **the changelog number needs bumping.**

This PR claims `061`, but `061` was taken by #540 (server-allocated document numbers), which merged earlier today. #545 (asset movement history, also open) has already rebased around it and taken `063`, `064` and `065`, deliberately leaving `062` free for this one.

So `062` is yours if you take it now. If anything else lands first, check the current maximum again rather than assuming, because three PRs have added migrations today and the numbering has moved twice.

Nothing else about the change is affected, and CI would not have caught it: Liquibase only complains when two changesets share an id at run time, which is after merge.

abhijithanandan:
Rebased onto current development and folded in the review fixes:

- Changelog renumbered 061 to 062 (file, changeset ids, master include): 061 went to #540 while this sat open. Liquibase would only have complained at deploy time, so CI green was no proof here.
- Claim guard hardened: claimed_by alone does not identify a claim, because a job whose lease lapsed can be requeued and re-claimed by the same replica. recordProgress and the terminal writes now also match on attempt, which every claim increments. New unit test plus an IT (aSupersededRunOnTheSameReplicaCannotWriteProgress) proving the stale write matches nothing.
- The worker revalidates preconditions at run time. The row outlives the configuration it was accepted under; a restart that lost the AI key would have fed the worker an empty list and recorded nothing-to-report for a run that asked nothing.
- Now that #527 is in: the poller and the worker's dispatch-row read carry @WithoutTenant, so the queue's tenant-less transactions are declared instead of tripping the transaction boundary every 2 s. Everything after the claim still runs under TenantScopedJobRunner, which withdraws the declaration.
- No paging parameters on the new endpoints, so the #555 guardrail is untouched.

Full suite green on the lab box (1103 tests) before the #527/#555 rebase; rerunning now on the rebased tree.

PR #543, opened 2026-08-28
Print the QA/QC and non-conformance reports, and mark up the evidence

### LINKED ISSUE #502: Inspection Phase 1: QA/QC and NCR reports, and annotated defect images (spec 3.6)
Phase 1 of the inspection module (spec `docs/specs/2026-08-26-inspection-module.md`, section 3.6) lists reports as in scope. #497, #498 and #499 deliver the rest of Phase 1 and deliberately leave this out, so it is filed here rather than left to the phase plan to remember.

## What is missing

Section 3.6 asks for three things, and Echno already generates PDF reports, so this is an addition to that path rather than a new one:

- **A QA/QC inspection report** (FR-REP-01..03). The inspection, its check points with acceptance criterion, tolerance, measurement and the computed deviation, its defects, and the summary counts. Everything it needs is on the record already: #497 added `acceptanceCriterion`, `tolerance` and `deviation` to `InspectionCheckItem` for exactly this.
- **An NCR report** (FR-REP-10..13). The non-conformance with its full accountability trail: who raised it, which site engineer owns it, the corrective action remarks and the date they were reported, who verified it and when, who closed it and when. #498 records all of that on `Ncr` precisely so a report can print it. A punch-list report is the same query the `?open=true` filter on `GET /api/v1/ncrs/web` already answers.
- **Annotated defect images** (FR-REP-10). The defect already stores photo URLs and a location string, so the annotation is the new part.

## What is explicitly not here

Defect density and safety heatmaps (FR-REP-11) depend on element or spatial coordinates and belong with the BIM phase, per section 3.6 of the spec. Do not pull them forward into this.

## Notes for whoever picks it up

- The existing PDF path is `openhtmltopdf` plus `openpdf`; follow whatever the finance and inventory reports already do rather than introducing a third approach.
- Reads must be bounded. The `UnboundedRepositoryReadTest` ratchet forbids a no-argument `findAll()` in production code, and a report over an inspection's children is exactly the shape that tempts one.
- Guards go through the `inspectionSecurity` bean #499 introduces, not a fresh role list. A report is a read, so `canRead()` is the guard unless there is a reason it is not.
- Annotated images are the only piece with a real unknown in it: where the annotation is drawn (server side into a derived image, or client side over the stored photo) is a decision, not a detail.

Blocked on nothing once #498 and #499 land.

abhijithanandan:
Closes #502. Section 3.6 of `docs/specs/2026-08-26-inspection-module.md` asks for three things and this delivers those three. Built on the deployed inspection model, `echno-core`'s existing `Inspection` contract and the existing services: no parallel model, no second report path.

## What the spec scoped, and what is here

| Spec | Built |
|---|---|
| QA/QC inspection report (FR-REP-01..03) | `GET /api/v1/inspections/web/{id}/pdf` |
| NCR report with its trail (FR-REP-10..13) | `GET /api/v1/ncrs/web/{id}/pdf` |
| Punch list (FR-REP-08, the `?open=true` query) | `GET /api/v1/ncrs/web/punch-list/pdf` |
| Annotated defect images (FR-REP-10) | `GET`/`PUT /api/v1/inspections/web/{id}/annotations`, printed as plates on the QA/QC report |
| Defect density and safety heatmaps (FR-REP-11) | **Not here**, deliberately. They need element or spatial coordinates and belong with the BIM phase. |

Every endpoint is behind `@inspectionSecurity`, the bean #499 introduced. The three reports are reads, so `canRead()`; drawing on an inspection is `canManageInspections()`, the authority that records one.

## Where an annotation lives, and why

The answer is forced rather than chosen, and finding out why is most of the work in this PR.

The obvious model is a child collection on `InspectionDefect`. **It does not work.** `InspectionService.update` clears the inspection's defects and rebuilds them from the payload, so with `orphanRemoval` every defect row is deleted and reinserted under a new id each time an inspector saves. Anything hanging off that row is destroyed with it. The `ncrs` changelog already records this ("an inspection's defects are rebuilt wholesale on every save") and is why an NCR keeps a scalar `defectId` rather than a foreign key. An annotation modelled that way would look like it worked until somebody edited a remark.

So an annotation hangs off the thing that is stable and that it actually describes: **the photograph**. `inspection_defect_annotations` is keyed by `(organization_id, inspection_id, photo)`, where `photo` is byte for byte the string in the defect's `photos` list. A defect rebuild carries the same photo strings through unchanged, so the marks survive it. There is a regression test for exactly that, and it asserts the defect id really did change.

Keying on the photo rather than its index in the list is the other half. The list is an element collection that the update path replaces wholesale, so index-keying would silently re-point a mark at a different photo whenever the list is reordered.

**Does an annotation survive its image being replaced? No, and it must not.** A replacement arrives under a new key, so the old marks match nothing and `DefectAnnotationService.removeOrphaned` deletes them from the save path. Carrying one over would be worse than losing it: the coordinates describe a region of specific pixels, so the mark would keep reading as evidence while pointing at bare wall, and a reader could not tell.

**Coordinates are fractions of the image, not pixels.** Two points in `[0, 1]`, read as opposite corners for a rectangle or ellipse and as tail then head for an arrow. One set of numbers therefore drives a thumbnail in the web app and a half-page plate in the report, and nothing stored depends on the resolution of the file or the device that took it. A `CHECK` constraint enforces the range in the schema as well as in the request validation: the validation stops a bad payload, the constraint stops a bad row whatever wrote it.

**Vector, not a derived image.** The annotation is stored as geometry and drawn at render time, server side into the PDF as CSS-positioned boxes and client side over the same photo. A derived, burnt-in image would double the object storage, need invalidating on every edit, and could not be un-drawn.

One rendering compromise: an arrow prints in the PDF as a pin on its head rather than a drawn line. This build has no SVG support (`openhtmltopdf-svg-support` is not a dependency) and a rotated line would rest on transform behaviour that a PDF renderer is a poor place to depend on. What an arrow says is where it lands. Both points stay on the record, so a browser draws the full arrow from the same row.

## Report machinery reused

- **`pdfTemplateEngine`** (the qualified `SpringTemplateEngine` in `ThymeleafConfig`) and **openhtmltopdf**, exactly as `ConstructionInvoicePdfService` uses them: formatting done in Java so the templates only place display-ready strings, `PdfLine`-style display records per row, inline `<style>` with `@page`.
- **`UnpagedResultCap`** for the 500-row ceiling and the `truncated` / `rowCap` context variables the site progress report already uses.
- **`NcrService.findAll(..., open=true, ...)`** for the punch list, rather than a second query.
- **`InspectionService.findById` / `NcrService.findById`** for the data, so a report can only be produced for a record in the caller's tenant.

Two things extracted rather than copied a third time:

- **`PdfRenderer`** holds the four-line `PdfRendererBuilder` block that was duplicated between the invoice service and the site progress controller. `ConstructionInvoicePdfService` moves onto it (byte-identical: it already used fast mode). `ReportController` is deliberately **not** moved: it does not use fast mode, and switching a live report's renderer mode is a change to its output that does not belong in a feature PR.
- **`ReportText`** holds the date, timestamp, enum-humanising and placeholder helpers that were private to the invoice service.

## Bounded reads

No new unbounded read. `UnboundedRepositoryReadTest` passes untouched, and its allowlist is unchanged.

Two reads here are unbounded in principle and are capped explicitly:

- **An inspection's check points and defects.** A report over an aggregate is not a `findAll()`, so the ratchet cannot see it, but an inspection may carry any number of either. Both tables stop at `UnpagedResultCap.MAX_ROWS`, print the **true** totals whether they fit or not, and the document flags itself when it is showing only part. There is a test that renders 700 of each.
- **The punch list**, taken as one `PageRequest.of(0, MAX_ROWS)` page with the same truncation note.

The annotations for a report are read one capped page at a time; `DefectPhotoAnnotationRepository` has no unpaginated read at all, and a write is bounded at 400 marks per inspection by request validation.

**Photographs are capped separately and much lower** (`echno.inspection.report.max-annotated-photos`, default 12, and `max-photo-bytes`, default 2 MB), because embedding an image costs bytes rather than rows and one site photo outweighs every table on the page. Only photographs that carry a mark are read: the report's subject is annotated evidence, and reading the rest would bury it. Anything that will not load prints a placeholder naming the gap rather than failing the report.

## Security note on embedding

Photographs are read out of object storage through `FileStorageService` and inlined as `data:` URIs, not left as an `` for the renderer to fetch. The bucket is private, so a plain URL is a 403 anyway, and openhtmltopdf's default user agent opens a URL stream with no timeout, which would hand a request-scoped thread an unbounded outbound fetch driven by a string out of the database. `keyForStoredReference` resolves a stored reference to a key in the configured bucket or to nothing, so a reference on any other host is never dereferenced. That is the tested behaviour, including the metadata-service address and a traversal attempt.

## One model gap closed

The issue says the trail records "who verified it and when". It recorded `verifiedAt` and the remarks but **not the person**, so a report could say the work had been verified without saying by whom, which is the one thing sign-off is for. `ncrs.verified_by_id` is added and written by `verify` and `reject`. Nullable and not backfilled: reports verified before the column existed have no such record, and inventing one would be worse than showing the gap.

## Migrations

- `v4.0/061-create-defect-photo-annotations.xml` (table, indexes, coordinate `CHECK`)
- `v4.0/062-add-ncr-verified-by.xml` (one nullable column)

Both guarded by preconditions and included in the master changelog.

## Every new test, and the mutation that makes it fail

Each row is a real run: the change was applied to the working tree, the named test run, then reverted.

| Mutation | Test that failed |
|---|---|
| `AnnotationGeometry` places a box at `x1,y1` with `x2-x1` instead of `min`/`abs` | `normalizesABoxDrawnFromItsFarCornerBack` |
| An arrow's pin is centred on its tail instead of its head | `placesAnArrowAsAPinCentredOnItsHead` |
| The replace path drops its "photo must be on a defect" check | `refusesAMarkOnAPhotoNoDefectOnTheInspectionCarries`, `refusesTheWholePayloadBeforeItDeletesAnything` |
| `removeOrphaned` becomes a no-op | `dropsEveryMarkWhenNoPhotographIsLeft`, `sweepsTheMarksWhosePhotoIsNoLongerAttached` |
| `InspectionService.update` stops calling the sweep | `anUpdateKeepsTheMarksOnThePhotosItKeepsAndDropsTheRest`, `anUpdateThatRemovesEveryPhotoDropsEveryMark` |
| `keyForStoredReference` drops the scheme and traversal guard | `refusesAReferenceThatDoesNotNameAnObjectInThisBucket`, `refusesAKeyThatTriesToClimbOutOfItsPrefix` |
| Changeset `061-02` drops the coordinate `CHECK` | `theDatabaseRefusesACoordinateOffTheImage` |
| The inspection report is guarded by `canManageInspections()` instead of `canRead()` | `theInspectionReportIsARead` |
| `qa-qc-report.html` gets one unclosed tag | all six `InspectionReportPdfServiceTest` cases |
| The report reads every defect photograph, not only the marked ones | `readsOnlyThePhotographsThatCarryAMark` |
| The punch list resolves employee names one row at a time | `resolvesEveryNameOnThePunchListInOneQuery` |
| The NCR report names its inspection by loading the aggregate | `namesTheInspectionWithoutLoadingIt` |
| The photograph never reaches the plate (a valid PDF with nothing on it) | `rendersAFullyPopulatedInspection` |
| The annotation list element loses its `@NotNull` | `refusesANullEntryInTheList` |
| The object GET drops its byte range and length check | `asksForOneByteMoreThanTheCapSoAnOversizedObjectCannotBeAllocated`, `refusesAnObjectThatGrewPastTheCapBetweenTheHeadAndTheGet` |
| `reject` records who re-inspected but not when | `aRejectionIsStampedLikeTheAcceptanceItRefuses` |

## CodeRabbit review

Reviewed (`success | Review completed`), three findings, all valid and all fixed in `7527bf9`:

1. **Null entry in the annotation list.** `List<@Valid ...>` skips null container elements, so `{"annotations": [null]}` passed validation and the service dereferenced it: a 500 where a 400 belongs. `@NotNull` added on the element, with a validator test.
2. **The photograph cap only bound the HEAD.** An object replaced between the HEAD and the GET could still be fetched whole. The GET now asks for `bytes=0-maxBytes`, one past the limit, and rechecks the returned length, so an oversized object is refused rather than either allocated or truncated into a corrupt image. Both halves are needed and the tests cover both.
3. **`reject` recorded who re-inspected but not when.** That was a gap this PR introduced by adding `verifiedById` without a matching timestamp: on a report accepted and reopened once already, the rejection rendered against the earlier acceptance's date. Both fields now mean "the last re-inspection decision" and both are written on accept and on reject; the report's row is labelled "Re-inspected" rather than "Verified" to match.

## Test heap

`build.gradle` is untouched: `maxHeapSize` stays at 1024m and there is still no `forkEvery`, so one JVM caches every distinct test configuration for the whole run.

- **One new `@Entity`**, `DefectPhotoAnnotation`. That is the multiplier across every cached context, so the feature was designed to add one rather than three (no `DefectPhoto` entity, no annotation-revision table).
- **No new Spring context.** `DefectAnnotationServiceIT` repeats `InspectionServiceIT`'s annotations and `@Import` list to the letter, and the two new beans were added to all four existing inspection ITs at the same time, so the five classes still share one cached context. No `@WebMvcTest` was added: the guard assertions are reflection over the controller, which costs nothing.
- Everything else is plain JUnit or plain Mockito. The two renderer tests build the template engine straight from `ThymeleafConfig`, following `ConstructionInvoicePdfServiceTest`.

## Verification

Full `./gradlew test` on lab `.116` (Testcontainers, CockroachDB v26.2.4): **BUILD SUCCESSFUL, 0 failures.** The run before the last commit caught a real regression, `InspectionProjectImmutabilityTest` building `InspectionService` with `@InjectMocks` and getting a null collaborator; that is the second commit.

## Follow-on, not in this PR

`echno-core` needs the `DefectAnnotationShape` enum, the `DefectPhotoAnnotation` type and parser, the annotation service methods, and `verifiedById` on the NCR type before `echno-web` can consume any of this. That is the usual backend, then core release, then web sequence.

## Summary by CodeRabbit

- **New Features**
  - Added downloadable QA/QC inspection reports with defect-photo annotations, captions, placeholders, and truncation notices.
  - Added downloadable NCR reports and punch lists with status, accountability, and summary details.
  - Added support for creating, viewing, replacing, and clearing annotations on defect photos using rectangles, ellipses, and arrows.
  - NCR verification and rejection actions now record the responsible employee.
- **Bug Fixes**
  - Improved report formatting, missing-value handling, and photo loading reliability.
  - Added safeguards against unsafe or oversized stored files.


abhijithanandan:
@coderabbitai review

PR #557, opened 2026-08-29
Delete dead file-URL builder and three unused exists-by-number methods

### LINKED ISSUE #550: buildFileUrl omits the bucket segment, and is dead code: delete it or fix the addressing
## What is wrong

`FileStorageService.buildFileUrl` builds an object URL by joining the CDN endpoint straight onto the key, with no bucket segment. MinIO is addressed path-style, so the URL it produces points at a path that does not exist.

It is also dead code, which changes what should be done about it. See below.

## What I verified (29 Aug 2026, `development`)

```java
private String buildFileUrl(String key) {
    String endpoint = cdnEndpoint.endsWith("/") ? cdnEndpoint : cdnEndpoint + "/";
    return endpoint + key;
}
```

`cdnEndpoint` is `${DIGITAL_OCEAN_SPACES_CDN_ENDPOINT}`, and on the IITM environments that is the MinIO origin, not a virtual-hosted bucket host:

- k8s: `DIGITAL_OCEAN_SPACES_CDN_ENDPOINT` is set from `echno.storageOrigin` in `templates/backend.yaml`, with `DIGITAL_OCEAN_SPACES_BUCKET_NAME` = `echno-objects` (`k8s/scripts/create-app-secrets.sh` has it as `http://echno-minio:9000`).
- compose: `object_store_cdn_endpoint` in `ansible/roles/apps/templates/backend.env.j2`, same shape.

Path-style access needs `<endpoint>/echno-objects/<key>`. The method produces `<endpoint>/<key>`.

**The picture changes here.** `grep -rn "buildFileUrl" src/` returns exactly one line, the declaration itself. It is `private`, and nothing in the codebase calls it. Nothing stores its output, nothing serves it, no test exercises it. Every path that actually hands a URL to the browser goes through `generateDownloadUrl` (presigned GET) or `generateUploadUrl` (presigned PUT), and both build their requests from `bucketName` + `key` through the SDK, which addresses the bucket correctly.

It is also **pre-existing and not introduced by the recent object-store work.** The behaviour is identical on compose and on k8s; the presigned-upload changes did not touch this method.

## Why it matters

Not much operationally, and that is the finding. Nothing is broken today because nothing calls it. What it is is a loaded gun: it looks like the sanctioned way to get a URL for a stored object, it is sitting in the same class as the two methods that are correct, and the first person who needs "the URL of this file" will find it and use it. At that point it produces a 404 that looks like a storage misconfiguration rather than a missing path segment, on an environment where object storage is already the fiddliest part of the stack.

## What a fix would involve

Two honest options.

**Delete it.** This is my preference. It has never been called, and the correct way to hand out an object URL is a presigned GET, since the bucket is private and a direct URL would not be readable anyway. A dead private method that would be wrong if used is worth less than nothing.

**Or fix it and keep it**, if there is a case for a non-presigned URL (a public-read prefix, for instance). Then it needs the bucket segment and a decision on addressing style, because the same string has to work against DO Spaces (virtual-hosted, bucket in the hostname) and MinIO (path-style, bucket in the path). That is configuration, not string concatenation, and it should come from the S3 client's own endpoint resolution rather than being rebuilt by hand.

Either way it should not stay as it is.

### LINKED ISSUE #551: Two exists-by-number repository methods are unused since #540 moved numbering server-side
## What is wrong

Two repository methods became unused when #540 moved document numbering server-side, and are still declared.

- `PurchaseOrderRepository.existsByPoNumberAndOrganization_Id(String, Long)` (line 19)
- `SiteTransferRepository.existsByTransferNumberAndOrganization_Id(String, Long)` (line 22)

Low priority, pure cleanup.

## What I verified (29 Aug 2026, `development`)

`grep -rn` over `src/` (main and test) returns exactly one hit each, the declaration. No caller, no test, no reference from a `@Query` or a specification.

## Why it keeps

They were left in deliberately. #537 was live in the same two interfaces at the time #540 was written, and deleting a method out from under an open branch is a conflict for no benefit. Both PRs are merged now, so the reason to leave them has gone.

## What a fix would involve

Delete the two declarations. Nothing else. Spring Data derives these from the method name, so there is no implementation to remove and no migration involved. Worth confirming as you go that #540's server-side numbering has a uniqueness guarantee of its own (a constraint or a sequence) rather than these having been the only check, though the PR reads as though it does.

Good first issue.

abhijithanandan:
Closes #550. Closes #551.

Two small deletions that were verified dead before filing, and re-verified on `development` at `d63a092` before touching anything.

## `FileStorageService.buildFileUrl` (#550)

The method joined the CDN endpoint straight onto the object key with no bucket segment. The object store is addressed path-style, so the URL it produced pointed at a path that does not exist. It was also `private` with no callers.

`grep -rn "buildFileUrl"` over the whole repository returns exactly one line, the declaration itself. Every URL that reaches a browser goes through `generateDownloadUrl` or `generateUploadUrl`, and both build their requests from `bucketName` plus `key` through the SDK, which addresses the bucket correctly. The bucket is private, so a plain object URL would not be readable even if the path were right. Deleted rather than repaired, per the issue's own recommendation.

The `cdnEndpoint` field stays: it is still read at line 229.

## The exists-by-number repository methods (#551)

Deleted:

- `PurchaseOrderRepository.existsByPoNumberAndOrganization_Id`
- `SiteTransferRepository.existsByTransferNumberAndOrganization_Id`
- `GoodsReceivedNoteRepository.existsByGrnNumber`

Each returned exactly one hit across the whole repository, its own declaration. No caller in `src/main`, no stub in `src/test`, no reference from a `@Query`, a specification, or an ArchUnit allowlist.

The third one is the addition the issue did not list. #540 moved **four** document families to server-side allocation (`DocumentNumberType`: `SITE_TRANSFER`, `PURCHASE_ORDER`, `INDENT`, `GOODS_RECEIVED_NOTE`), so the GRN and indent equivalents were worth checking too. The GRN one is dead in the same way, and was not organization-scoped either, so it would have answered across tenants had anyone picked it up.

**The indent equivalent stays.** `IndentRepository.existsByIndentNumberAndOrganization_Id` is still called from `IndentService.updateIndent`, which lets a caller rename an indent and checks the new number for a collision first. That path is live, so the method is not dead.

`StockAdjustmentRepository.existsByAdjustmentNumberAndOrganization_Id` is also unreferenced, but stock adjustments are not one of #540's four families and separate work is open in that package, so it is left alone rather than dragged into an unrelated cleanup.

## Uniqueness is still guaranteed

The issue asked for this to be confirmed rather than assumed. Server-side allocation is a single `INSERT ... ON CONFLICT ... DO UPDATE ... RETURNING` against a counter row per `(organization, document type, year)`, so two committed transactions cannot come away with the same number. Behind that, `po_number`, `transfer_number` and `grn_number` each carry a unique constraint in the changelog. The deleted methods were never the mechanism, and they were not the backstop either.

## Verification

Full `./gradlew build` on a lab node: BUILD SUCCESSFUL, whole test suite including the ArchUnit checks.

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal file-storage and record-management components.
  * Removed unused duplicate-checking operations for purchase orders, goods received notes, and site transfers.
  * Existing file uploads, downloads, storage access, and record workflows remain unchanged.

* **Bug Fixes**
  * No direct user-facing bug fixes were included in this release.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/service/FileStorageService.java
+++ b/src/main/java/org/tornotron/echno_backend/common/service/FileStorageService.java
@@ -358,17 +358,6 @@ public class FileStorageService {
     }
 
     /**
-     * Builds the public URL for a stored file.
-     *
-     * @param key The storage key
-     * @return The public URL
-     */
-    private String buildFileUrl(String key) {
-        String endpoint = cdnEndpoint.endsWith("/") ? cdnEndpoint : cdnEndpoint + "/";
-        return endpoint + key;
-    }
-
-    /**
      * Rolls back uploaded files by deleting them.
      *
      * @param keys The keys of files to delete
--- a/src/main/java/org/tornotron/echno_backend/goodsReceivedNote/GoodsReceivedNoteRepository.java
+++ b/src/main/java/org/tornotron/echno_backend/goodsReceivedNote/GoodsReceivedNoteRepository.java
@@ -15,6 +15,4 @@ public interface GoodsReceivedNoteRepository extends JpaRepository<GoodsReceived
     List<GoodsReceivedNote> findByVendorId(Long vendorId);
 
     List<GoodsReceivedNote> findByReceivedOnBetween(LocalDateTime startDate, LocalDateTime endDate);
-
-    boolean existsByGrnNumber(String grnNumber);
 }
--- a/src/main/java/org/tornotron/echno_backend/purchaseOrder/PurchaseOrderRepository.java
+++ b/src/main/java/org/tornotron/echno_backend/purchaseOrder/PurchaseOrderRepository.java
@@ -16,7 +16,5 @@ public interface PurchaseOrderRepository extends JpaRepository<PurchaseOrder, Lo
 
     List<PurchaseOrder> findByStatus(PurchaseOrderStatus status);
 
-    boolean existsByPoNumberAndOrganization_Id(String poNumber, Long organizationId);
-
     Optional<PurchaseOrder> findByIdAndOrganization_Id(Long id, Long organizationId);
 }
--- a/src/main/java/org/tornotron/echno_backend/siteTransfer/SiteTransferRepository.java
+++ b/src/main/java/org/tornotron/echno_backend/siteTransfer/SiteTransferRepository.java
@@ -19,7 +19,5 @@ public interface SiteTransferRepository extends JpaRepository<SiteTransfer, Long
 
     List<SiteTransfer> findByIssueDateBetween(LocalDateTime startDate, LocalDateTime endDate);
 
-    boolean existsByTransferNumberAndOrganization_Id(String transferNumber, Long organizationId);
-
     Optional<SiteTransfer> findByIdAndOrganization_Id(Long id, Long organizationId);
 }


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #558, opened 2026-08-29
Move a site transfer against the balance row it writes

### LINKED ISSUE #556: A site transfer with no storage location reports the project total but moves the unlocated row
Found while landing #554 (issue #533). This is the same mismatch #539 fixed for consumption, still present on the site transfer path.

## What is wrong

`CurrentStock` is keyed by `(material, project, storage location)`, and a null storage location is a row of its own rather than a rollup. `InventoryService.findUnlocatedStock` says so in as many words:

> This is the row a movement with no storage location reads and writes. It is not the project total: stock booked into a storage location lives in its own row and is not reachable from here.

`InventoryEventListener.handleSiteTransferCreated` does not follow that. With no sending storage location it reads the opening figure from `getCurrentStock`, which is `sumCurrentQuantityByMaterialAndProject`, the project total across every location:

```java
if (transfer.getSendingStorageLocation() != null) {
    sendingOpeningStock = inventoryService.getStockAtLocation(
            item.getMaterial().getId(), transfer.getSendingProject().getId(),
            transfer.getSendingStorageLocation().getId());
} else {
    sendingOpeningStock = inventoryService.getCurrentStock(
            item.getMaterial().getId(), transfer.getSendingProject().getId());
}
```

Then it debits a different quantity:

```java
inventoryService.updateCurrentStock(
        item.getMaterial(), transfer.getSendingProject(), transfer.getSendingStorageLocation(),
        transfer.getOrganization(), sendingQuantityChanged, null);
```

`updateCurrentStock` with a null location locks and writes the unlocated row. So the `TRANSFER_OUT` ledger entry states an opening and a closing figure for a balance the movement never touches. The receiving side has the same shape: `getCurrentStock` for the opening figure, the unlocated row for the credit.

## The matching hole in validation

`SiteTransferService.createSiteTransferInTransaction` picks its check off the same branch:

```java
if (creationDto.getSendingStorageLocationId() != null) {
    inventoryService.validateSufficientStockForMultipleItemsAtLocation(
            requiredQuantities, sendingProject.getId(), creationDto.getSendingStorageLocationId());
} else {
    inventoryService.validateSufficientStockForMultipleItems(requiredQuantities, sendingProject.getId());
}
```

`validateSufficientStockForMultipleItems` sums every row on the project. A project holding 400 in a storage location and nothing outside one passes a check for 30, and the unlocated row it then debits goes to -30. `validateSufficientUnlocatedStock` exists for exactly this and carries the warning in its own javadoc, but there is no multiple-item form of it, so the site transfer path had nothing to call.

## The live evidence

Staging (`echno.in`) has one movement row that already carries a wrong opening figure, and it is the pre-#539 consumption path rather than a transfer:

```
 id | transaction_type | material_id | project_id | storage_location_id | opening_stock | quantity_changed | closing_stock | reference_number
 16 | USE              |           8 |           7 |                NULL |           400 |              -30 |           370 | MC-1
```

Material 8 at project 7 held 400 at storage location 2 and nothing outside a location. The entry claims 400 down to 370; the row it actually debited went 0 to -30, and that is `current_stock` id 15, the negative balance flagged as item 4 of #529. The origin was never identified there. This is it.

No transfer row is wrong yet, only because no transfer has been created without a sending storage location. The web form makes both locations mandatory, so today the bug is reachable through the API and not the UI. That is a property of the current frontend, not a guard.

## What to do

Follow #539 rather than invent a second approach.

- Read the sending and receiving opening figures from `findUnlocatedStock(...).orElse(0.0)` when the side names no location, matching what `handleMaterialConsumed` now does.
- Add the multiple-item form of the unlocated check to `InventoryService` and call it from `SiteTransferService` on the no-sending-location branch.

`handleGrnCreated` reads its opening figure the same wrong way. A GRN only ever adds, so it cannot drive a balance negative, but the figure it writes on the ledger is still not the row it credits, and it should be corrected alongside.

abhijithanandan:
Closes #556

Found while landing #554. It is the mismatch #539 fixed for consumption, still present on the site transfer path, so this follows what #539 did rather than inventing a second approach.

## What was wrong

`CurrentStock` is keyed by `(material, project, storage location)`, and a null storage location is a row of its own rather than a rollup over the located rows. `InventoryService.findUnlocatedStock` already says so:

> This is the row a movement with no storage location reads and writes. It is not the project total: stock booked into a storage location lives in its own row and is not reachable from here.

`InventoryEventListener.handleSiteTransferCreated` did not follow that. With no sending storage location it read the opening figure from `getCurrentStock`, the project total across every location, and then debited the unlocated row through `updateCurrentStock`. Two different quantities: the `TRANSFER_OUT` entry stated an opening and a closing figure for a balance the movement never touched. The receiving side had the same shape, and so did `handleGrnCreated`.

`SiteTransferService` chose its check off the same branch, calling `validateSufficientStockForMultipleItems`, which sums every row on the project. A project holding 400 in a storage location and nothing outside one passed a check for 30, and the unlocated row it then debited went to -30.

## Where it had already happened

Row 16 of `inventory_transaction` on `echno.in`:

```
 id | transaction_type | material_id | project_id | storage_location_id | opening_stock | quantity_changed | closing_stock | reference_number
 16 | USE              |           8 |           7 |                NULL |           400 |              -30 |           370 | MC-1
```

Material 8 at project 7 held 400 at storage location 2 and nothing outside a location. The entry claims 400 down to 370; the row it actually debited went 0 to -30, and that is `current_stock` id 15, the negative balance reported as item 4 of #529 with no cause identified. This is the cause. It came through consumption, which #539 has since fixed, so nothing needs correcting beyond that one historical row.

No transfer row is wrong yet, only because no transfer has been created without a sending storage location. The web form makes both locations mandatory, which is a property of the current frontend and not a guard. The API accepts a null on either side.

## The change

- Both sides of `handleSiteTransferCreated`, and `handleGrnCreated`, read their opening figure from `findUnlocatedStock(...).orElse(0.0)` when the side names no location, matching what `handleMaterialConsumed` does since #539. A GRN only ever adds so it could not drive a balance negative, but the figure it wrote was still not the row it credited.
- `InventoryService.validateSufficientUnlocatedStockForMultipleItems` is the multiple-item form of `validateSufficientUnlocatedStock`, which existed for exactly this and had no plural version, which is why the site transfer path had nothing correct to call. It collects every shortfall rather than stopping at the first, like the two checks either side of it, and its message carries the same "name the location to draw from" hint as the single-item version.
- `SiteTransferService` calls it on the no-sending-location branch.

`validateSufficientStockForMultipleItems` and `getCurrentStockForMaterials` are deleted. Both are left with no caller by this change, and both encode the project-total scope that is the bug, so leaving them is leaving the next caller a way to reintroduce it. `validateSufficientStock` is also dead, but it was already dead before this change, so it belongs to whatever sweep picks up #473's leftovers rather than here.

The frontend half is echno-web #324, which is deliberately not started: two other agents are in that repo and it is a form-level change.

## Tests

Every new test was run against the fix removed.

| Test | Fix removed | Result without it |
|---|---|---|
| `InventoryEventListenerSiteTransferTest.aTransferOutWithNoSendingLocationOpensFromTheSendingProjectsUnlocatedBalance` | listener opening figures back to `getCurrentStock` | **FAILED** |
| `InventoryEventListenerSiteTransferTest.aTransferInWithNoReceivingLocationOpensFromTheReceivingProjectsUnlocatedBalance` | same | **FAILED** |
| `InventoryEventListenerSiteTransferTest.aProjectWithNoUnlocatedRowAtAllOpensFromZeroRatherThanTheProjectTotal` | same | **FAILED** |
| `InventoryEventListenerSiteTransferTest.aGrnWithNoStorageLocationOpensFromTheProjectsUnlocatedBalance` | same | **FAILED** |
| `SiteTransferStockScopeTest.aTransferWithNoSendingLocationIsRefusedWhenTheProjectHoldsItsStockInsideLocations` | service back to `validateSufficientStockForMultipleItems`, method restored | **FAILED** |
| `SiteTransferStockScopeTest.aTransferWithNoSendingLocationIsRefusedWhenTheUnlocatedBalanceIsShort` | same | **FAILED** |
| `SiteTransferStockScopeTest.aTransferWithNoSendingLocationIsAllowedAgainstTheUnlocatedBalance` | same | **FAILED** |
| `SiteTransferStockScopeTest.everyShortMaterialIsNamedRatherThanOnlyTheFirst` | same | **FAILED** |
| `SiteTransferServiceTest.create_aggregatesQuantitiesOfRepeatedMaterial_forStockCheck` | same | **FAILED** |

Two are controls and pass under the matching revert, which is what they are for: `InventoryEventListenerSiteTransferTest.aTransferBetweenTwoLocationsStillOpensFromEachLocationsOwnBalance` and `SiteTransferStockScopeTest.aTransferFromANamedLocationStillReadsThatLocationsBalance` pin the located branch, which this change does not touch. `SiteTransferServiceTest.create_withSendingLocation_usesLocationScopedStockCheck` also passes under the revert, because its `never()` on the renamed method is vacuously true there.

`SiteTransferStockScopeTest` runs the service over a real `InventoryService` and a mocked `CurrentStockRepository`, following `MaterialConsumptionStockScopeTest` from #539, so the balance row the check reads is the thing under test rather than a mock's say-so.

No changelog in this one. Full suite green on the lab box.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/events/listeners/InventoryEventListener.java
+++ b/src/main/java/org/tornotron/echno_backend/common/events/listeners/InventoryEventListener.java
@@ -44,13 +44,16 @@ public class InventoryEventListener {
         logger.info("Handling GRN created event for GRN number: {}", grn.getGrnNumber());
 
         for (GrnItem item : grn.getItems()) {
+            // Read the balance this receipt actually credits. With no storage location that is
+            // the project's unlocated row, not the project total.
             Double openingStock;
             if (grn.getStorageLocation() != null) {
                 openingStock = inventoryService.getStockAtLocation(
                         item.getMaterial().getId(), grn.getProject().getId(),
                         grn.getStorageLocation().getId());
             } else {
-                openingStock = inventoryService.getCurrentStock(item.getMaterial().getId(), grn.getProject().getId());
+                openingStock = inventoryService.findUnlocatedStock(
+                        item.getMaterial().getId(), grn.getProject().getId()).orElse(0.0);
             }
             Double quantityChanged = item.getReceivedQuantity().doubleValue();
             Double closingStock = openingStock + quantityChanged;
@@ -136,15 +139,20 @@ public class InventoryEventListener {
         logger.info("Handling site transfer created event for transfer number: {}", transfer.getTransferNumber());
 
         for (SiteTransferItem item : transfer.getItems()) {
-            // TRANSFER_OUT: Stock decreases at sending project
+            // TRANSFER_OUT: Stock decreases at sending project.
+            // Read the balance this movement actually debits. With no sending storage location
+            // that is the sending project's unlocated row, not the project total: the total
+            // would put an opening and closing figure on the ledger entry that the row it
+            // moves never held, and it is a different quantity from the one updateCurrentStock
+            // goes on to write.
             Double sendingOpeningStock;
             if (transfer.getSendingStorageLocation() != null) {
                 sendingOpeningStock = inventoryService.getStockAtLocation(
                         item.getMaterial().getId(), transfer.getSendingProject().getId(),
                         transfer.getSendingStorageLocation().getId());
             } else {
-                sendingOpeningStock = inventoryService.getCurrentStock(
-                        item.getMaterial().getId(), transfer.getSendingProject().getId());
+                sendingOpeningStock = inventoryService.findUnlocatedStock(
+                        item.getMaterial().getId(), transfer.getSendingProject().getId()).orElse(0.0);
             }
             Double sendingQuantityChanged = -item.getSentQuantity().doubleValue();
             Double sendingClosingStock = sendingOpeningStock + sendingQuantityChanged;
@@ -181,14 +189,16 @@ public class InventoryEventListener {
                     item.getMaterial().getId(), transfer.getSendingProject().getId(), sendingQuantityChanged, sendingClosingStock);
 
             // TRANSFER_IN: Stock increases at receiving project (carries avg cost from sender)
+            // Same on the receiving side: the row credited with no location is the receiving
+            // project's unlocated row.
             Double receivingOpeningStock;
             if (transfer.getReceivingStorageLocation() != null) {
                 receivingOpeningStock = inventoryService.getStockAtLocation(
                         item.getMaterial().getId(), transfer.getReceivingProject().getId(),
                         transfer.getReceivingStorageLocation().getId());
             } else {
-                receivingOpeningStock = inventoryService.getCurrentStock(
-                        item.getMaterial().getId(), transfer.getReceivingProject().getId());
+                receivingOpeningStock = inventoryService.findUnlocatedStock(
+                        item.getMaterial().getId(), transfer.getReceivingProject().getId()).orElse(0.0);
             }
             Double receivingQuantityChanged = item.getSentQuantity().doubleValue();
             Double receivingClosingStock = receivingOpeningStock + receivingQuantityChanged;
--- a/src/main/java/org/tornotron/echno_backend/inventoryTransaction/InventoryService.java
+++ b/src/main/java/org/tornotron/echno_backend/inventoryTransaction/InventoryService.java
@@ -124,22 +124,6 @@ public class InventoryService {
     }
 
     /**
-     * Returns current stock for several materials at a project in one call.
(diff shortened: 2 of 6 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

Issue #541, opened 2026-08-28
Client-supplied status on sixteen create endpoints, including approving a project and closing an issue without the authority to


abhijithanandan:
Found while fixing #531. That issue was about purchase orders being creatable already `APPROVED`; #540 closes it by typing `status` as the enum and refusing anything but `DRAFT` on create. The same weakness is on sixteen other create endpoints, so this is the sweep.

The shape is always the same: `status` is on the creation DTO, the service copies it (or `valueOf`s it) onto the entity, and nothing checks whether the value is a legitimate *initial* state or whether the caller holds the authority the value implies.

## Survey

| Entity | DTO field | Type | Written at | Restricted? | Approval-bearing value reachable? | Separate status endpoint |
|---|---|---|---|---|---|---|
| Project | `project/dto/ProjectCreationDto.java:60` | String → `ProjectCreationStatus` | `ProjectService.java:113` | no | **yes, `approved`** | no, only `PATCH /{id}` |
| Issue | `issue/dto/IssueCreationDto.java:39` | String → `IssueStatus` | `IssueService.java:87` | no | **yes, `resolved`, `closed`** | no, `PATCH /issues/{id}` |
| SiteTransfer | `siteTransfer/dto/SiteTransferCreationDto.java:46` | String → `SiteTransferStatus` | `SiteTransferService.java:174` | no | **yes, `COMPLETED`** | yes, `PATCH /{id}/status` |
| StockAdjustment | `stockAdjustment/dto/StockAdjustmentCreationDto.java:22` | free String | `StockAdjustmentService.java:124` | no | yes, any text | no |
| Expense | `expense/dto/ExpenseCreationDto.java:22` | free String | `ExpenseService.java:127` | no | yes, any text | no |
| Receipt | `receipt/dto/ReceiptCreationDto.java:19` | free String | `ReceiptService.java:134` | no | yes, any text | no |
| SubContract | `subcontract/dto/SubContractCreationDto.java:57` | free String | `SubContractService.java:122` | no | yes, any text | no |
| Asset | `asset/dto/AssetCreationDto.java:30` | free String | `AssetService.java:95` | no | yes, any text | no |
| Task | `task/dto/TaskCreationDto.java:66` | String → `TaskStatus` | `TaskService.java:101` | no | yes, `completed` | no |
[...]

PR #559, opened 2026-08-29
Require someone else to approve what you raised

### LINKED ISSUE #548: Whoever drafts a stock adjustment can approve it: no segregation of duties on the only path that sets a balance
## What is wrong

Nothing stops the person who drafts a stock adjustment from approving it themselves. Approval is the step that writes to the inventory ledger and moves a balance, so one person can set stock to any figure they like without a second pair of eyes.

Came out of reviewing #539.

## What I verified

`StockAdjustmentService.approve(Long id)` on `development` loads the document, refuses a second posting, validates the lines, posts the ledger entries, and then does:

```java
Long approver = userContextService.getCurrentUserId();
stockAdjustment.setApprovedBy(approver);
stockAdjustment.setApprovedAt(postedAt);
stockAdjustment.setProcessedBy(approver);
stockAdjustment.setProcessedAt(postedAt);
```

It never compares the approver against whoever created or submitted the document. The entity has a `submittedBy` field, which the creation DTO fills in, so the information to compare against is already there and simply unused.

`StockAdjustmentControllerWeb` gates create, update, delete and `POST /{id}/approve` with the identical expression:

```java
@PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')")
```

So any project manager can create a document and approve it in the next request.

## Why it matters

An adjustment is the only path that can set a balance to an arbitrary number, and the module's own documentation says so. Segregation of duties is the control that makes that safe: someone raises the correction, someone else agrees to it. Without it the audit trail records who approved but not that anyone independent reviewed, and inventory shrinkage becomes writable by one account.

It was not caught in review because it is consistent with everything around it. The only other approval steps in the codebase are `ConstructionInvoiceControllerWeb.approve` and the two attendance controllers, and the invoice one carries the same `hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')` gate with no self-approval check either. (Purchase orders and site transfers have no approval step at all.) That consistency is the reason to decide the rule once rather than patch this one service.

## What a fix would involve

Two decisions before any code:

1. Is a self-approval refused outright, or allowed only for a higher role (say `system-admin` approving their own draft when no one else is available)? Refusing outright is the stricter reading and the easier one to explain.
2. Does the same rule apply to construction invoice approval and to attendance approval? If yes it belongs in one shared check rather than three copies.

The implementation once decided is small: capture the creating user on the document (the entity already has `submittedBy`, but confirm it is set from the session rather than trusted from the request body), then in `approve` compare it against `userContextService.getCurrentUserId()` and throw an `InvalidRequestException` on a match. Add a test alongside `StockAdjustmentApprovalTest`.

Worth noting the second half of decision 2 also needs the create path to stop trusting a client-supplied submitter, which overlaps with #541.

abhijithanandan:
Closes #548.

## The rule I chose, and why

**Refuse a self-approval, with one recorded exception for `system-admin`.** Not a hard four-eyes rule, and not a per-organisation setting.

The three options were: refuse outright, refuse by default with a per-organisation opt-out, or allow it and stamp it for the audit. I ruled out each of the other two for a concrete reason.

Refusing outright is unsafe here in a way it would not be elsewhere. A stock adjustment is the only path that can correct a balance, which is why #539 built it: staging held a `-30` row nothing could fix. A tenant with one person who has the authority to correct stock would then have a wrong figure and no route back, which is the exact failure the module exists to remove.

A per-organisation setting would work, and the codebase already has the shape for it (`FinanceSettings.approvalThreshold` is a per-tenant knob that relaxes an approval control). I did not take it because it buys nothing the role does not already give. The relaxation has to be a deliberate act by someone with authority over the tenant either way, and `system-admin` is that someone. A setting adds a table, an endpoint, a migration and a value that can be switched on once and forgotten, in exchange for the same outcome.

Recording rather than refusing is too weak on its own: the issue is a missing control, and a note in an audit trail is not a control. So the recording is folded into the refusal instead. When a system administrator does approve their own document it is logged at WARN, the document shows the raiser and the approver side by side (both are already on the DTO), and every ledger entry the adjustment writes carries `(self-approved: raised and approved by the same person)` in its remarks. The ledger is what a stock figure is explained from, so that is where a movement nobody independent agreed to should say so.

## What the check needed first

The rule is worth nothing while the raiser is whatever the request body claims. `StockAdjustmentService.applyHeaderFields` copied `submittedBy` straight off the creation DTO, so a drafter could name somebody else and clear their own way to approve. The document now records the raiser (and `submittedAt`) from the session on create, and `update` leaves the field alone. `StockAdjustmentCreationDto.submittedBy` is documented and marked read-only rather than deleted, so the web client keeps compiling; the value sent is ignored.

That is the same weakness #541 surveys for `status`. This PR fixes only `submittedBy` on this one entity, because it is a precondition for the check; `status` is untouched and stays #541's.

Rows written before this carry no raiser. There is nothing to compare, so they are allowed through rather than becoming unapprovable. Documented on the policy.

## Construction invoices

Same rule, same object, applied to `ConstructionInvoiceService.approve`. The issue is right that two approval paths behaving differently for no stated reason would be worse than either, so the check lives in one `common/approval/SelfApprovalPolicy` that both call. Invoices already record `submittedBy` from the session in `submit`, so nothing else was needed there.

**One deliberate exemption, stated because it is a real difference:** the auto-approval inside `submit` is not subject to the rule. A threshold set on the organisation or the project is the tenant's own standing decision that invoices below that figure do not need a second person, and passing the submitter's own invoice through it is exactly what it was configured for. Subjecting it to the rule would make the threshold unreachable for its most common case. Above the threshold, and where no threshold is set, the invoice goes to `approve`, which requires a different person.

Attendance approval is out of scope. It is a different risk class (approving a timesheet does not post to a ledger) and its two controllers take the approver as a request parameter rather than from the session, so the same check would need that path fixed first. Worth its own issue.

## What the live data shows

Staging has one real self-approval. `CINV-2027-000002` was submitted by user 1 and approved by user 1, `submitted_at 08:19:15`, `approved_at 08:19:48`, 33 seconds apart. `finance_settings.approval_threshold` is NULL, so no auto-approval was involved: it was a manual self-approval through `POST /{id}/approve`. This PR would have refused it unless that user held `system-admin`.

Stock adjustments on staging: one row, still a draft, never approved. Nothing to migrate and no existing posting to reinterpret.

## Failing-test table

Each new test run against a reverted baseline (`checkSelfApproval` stubbed to `return false`, `submittedBy` restored to the request body), then against the fix.

|---|---|---|
| `SelfApprovalPolicyTest.approvingYourOwnDocumentIsRefused` | FAILED | PASSED |
| `SelfApprovalPolicyTest.aSystemAdministratorMayApproveTheirOwnDocumentAndItComesBackAsASelfApproval` | FAILED | PASSED |
| `StockAdjustmentApprovalTest.theSamePersonCannotRaiseAndApproveTheDocumentAndNothingIsPosted` | FAILED | PASSED |
| `StockAdjustmentApprovalTest.aSystemAdministratorApprovingTheirOwnDocumentPostsAndTheLedgerEntrySaysSo` | FAILED | PASSED |
| `StockAdjustmentServiceTest.create_recordsTheSessionUserAsTheRaiserAndIgnoresTheOneInTheBody` | FAILED | PASSED |
| `StockAdjustmentServiceTest.update_leavesTheRecordedRaiserAloneWhateverTheBodySays` | FAILED | PASSED |
| `ConstructionInvoiceSelfApprovalTest.theSubmitterCannotApproveTheirOwnInvoiceAndNoEntryIsPosted` | FAILED | PASSED |
| `ConstructionInvoiceSelfApprovalTest.aSystemAdministratorMayApproveTheirOwnInvoice` | FAILED | PASSED |

Six further new tests are negative controls and pass either way by design, since they pin that the rule does **not** fire where it should not: a different approver, a document with no recorded raiser, an unresolved approver, and the invoice equivalent.

Full suite on `lab-node-116-f`: **BUILD SUCCESSFUL**, 1127 tests, 0 failures, rebased onto `origin/development` at `2385513` (after #555). No Liquibase changeset: `approved_by` and `submitted_by` already exist on both tables, and self-approval is derivable from them, so a redundant boolean column that can disagree with the two id columns was not worth adding.

The three construction-invoice ITs use sliced `@DataJpaTest` contexts and needed `SelfApprovalPolicy` and `OrganizationSecurityService` added to their `@Import` lists. No new contexts are created, so no change to the test-heap position.

## If you would rather have it another way

If you want the per-organisation setting instead, the shape is a boolean on `finance_settings` (or its own `approval_settings` row on the same get-or-create idiom) read inside `SelfApprovalPolicy`. Only that one class changes; neither service and none of the tests would move. If you want it stricter, drop the `system-admin` exception: delete the role check, and `checkSelfApproval` becomes an unconditional refusal. Either way the decision stays in one place.

## Summary by CodeRabbit

* **New Features**
  * Added segregation-of-duties controls for invoice and stock adjustment approvals.
  * Submitters cannot approve their own documents unless they have the system-admin break-glass role.
  * Approval records now capture the authenticated submitter, ignoring client-provided values.
  * Break-glass self-approvals are clearly marked in ledger remarks.

* **Documentation**
  * Updated API documentation to explain approval restrictions, exceptions, and validation errors.

* **Tests**
  * Added coverage for standard, self-approval, break-glass, auto-approval, and legacy-record scenarios.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/approval/SelfApprovalPolicy.java
+++ b/src/main/java/org/tornotron/echno_backend/common/approval/SelfApprovalPolicy.java
@@ -1,0 +1,69 @@
+package org.tornotron.echno_backend.common.approval;
+
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+import org.tornotron.echno_backend.common.exception.InvalidRequestException;
+import org.tornotron.echno_backend.common.service.OrganizationSecurityService;
+
+/**
+ * Segregation of duties on the approvals that post an entry: whoever raised a document is not
+ * the person who approves it.
+ *
+ * <p>An approval is the second pair of eyes on the entry it writes. Where the same account both
+ * raises and approves a document, that entry has been reviewed by nobody, and the record shows
+ * an approver without showing that anyone independent agreed. The role gate on the endpoint does
+ * not supply this: every approver may also draft, so a single role holder can do both halves in
+ * two requests.
+ *
+ * <p>The rule applied here is a refusal with one recorded exception. An ordinary approver is
+ * refused outright. A {@value #BREAK_GLASS_ROLE} may approve their own document, because a small
+ * tenant can genuinely have one person with the authority to correct anything and a rule that
+ * cannot be satisfied would leave a wrong figure standing with no route back. That exception is
+ * logged and stays visible on the document, where the raiser and the approver are both recorded
+ * and can be compared.
+ *
+ * <p>A document that names nobody as its raiser cannot be compared against anything, so it is
+ * allowed through. That is the case only for rows written before the raiser was stamped from the
+ * session, since every path that creates one now records who did it.
+ */
+@Slf4j
+@Component
+public class SelfApprovalPolicy {
+
+    /** The one role that may approve its own document, and only as a recorded exception. */
+    public static final String BREAK_GLASS_ROLE = "system-admin";
+
+    private final OrganizationSecurityService orgSecurity;
+
+    public SelfApprovalPolicy(OrganizationSecurityService orgSecurity) {
+        this.orgSecurity = orgSecurity;
+    }
+
+    /**
+     * Applies the rule to one approval, and reports whether it went through as a self-approval.
+     *
+     * @param raisedBy The user who raised the document, or null when the document does not say.
+     * @param approver The user approving it now.
+     * @param document How to name the document in the refusal, for example "Stock adjustment with ID 5".
+     * @return true when the approver is approving their own document under the break-glass role,
+     *         so the caller can record the approval as one; false when the two are different
+     *         people, which is the ordinary case.
+     * @throws InvalidRequestException if the approver raised the document and does not hold the
+     *         break-glass role.
+     */
+    public boolean checkSelfApproval(Long raisedBy, Long approver, String document) {
+        if (raisedBy == null || approver == null || !raisedBy.equals(approver)) {
+            return false;
+        }
+        if (!orgSecurity.hasAnyOrgRoleForCurrentTenant(BREAK_GLASS_ROLE)) {
+            throw new InvalidRequestException(document + " was raised by the same person who is now "
+                    + "approving it. An approval is the second pair of eyes on the entry it posts, so it "
+                    + "has to come from someone other than whoever raised the document. Ask another "
+                    + "approver to review it, or have a system administrator approve it, which is "
+                    + "recorded as a self-approval.");
+        }
+        log.warn("Self-approval: {} was raised and approved by user {}, allowed under the {} role",
+                document, approver, BREAK_GLASS_ROLE);
+        return true;
+    }
+}
--- a/src/main/java/org/tornotron/echno_backend/finance/construction/service/ConstructionInvoiceService.java
+++ b/src/main/java/org/tornotron/echno_backend/finance/construction/service/ConstructionInvoiceService.java
@@ -6,6 +6,7 @@ import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
+import org.tornotron.echno_backend.common.approval.SelfApprovalPolicy;
 import org.tornotron.echno_backend.common.configuration.MoneyUtils;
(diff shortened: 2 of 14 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

Issue #532, opened 2026-08-28
s3Client has no path-style addressing, so every server-side object-store call fails with UnknownHostException on echno-objects.echno-minio


abhijithanandan:
## Reported

Anand, QA pass on `echno.in`, 28 Aug 2026. ClickUp `86d45v7nt`.

Attaching a file while editing an Issue fails with "Unexpected error occurred", the detail mentioning an `UnknownException` while interacting with a service. The Issue itself saves, then a second toast says "Issue updated, some files failed".

## Root cause

Confirmed from the staging backend log rather than inferred. Three occurrences on 28 Aug from 09:02 UTC, all identical:

```
software.amazon.awssdk.core.exception.SdkClientException: Received an UnknownHostException
when attempting to interact with a service. ...
Caused by: java.net.UnknownHostException: echno-objects.echno-minio
```

`echno-objects` is the bucket. `echno-minio` is the MinIO container. The SDK is putting the bucket in the **host**, virtual-host style, and Docker DNS has no such name. Only `echno-minio` resolves.

The reason is in `common/configuration/SpacesConfig.java`. There are two S3 beans and only one of them was given path-style addressing.

`s3Presigner()`, lines 50-66, is correct, and its comment even says why:

```java
// Path-style addressing is required for MinIO, which puts the bucket in the path rather than the host.
.serviceConfiguration(
        S3Configuration.builder()
                .pathStyleAccessEnabled(true)
                .build()
)
```

`s3Client()`, lines 30-44, has no `serviceConfiguration` at all:

```java
return S3Client.builder()
        .endpointOverride(URI.create(spaces))
        .credentialsProvider(...)
        .region(Region.US_EAST_1)
        .build();
```

So every server-side object-store call, which is the path the Issue edit form still takes, builds `http://echno-objects.echno-minio` and fails to resolve. The presigned browser path works, because the presigner is configured properly. That is why this looks intermittent depending on which screen you are on.

Staging environment, for the record:

```
DIGITAL_OCEAN_SPACES_URI=http://echno-minio:9000
DIGITAL_OCEAN_SPACES_BUCKET_NAME=echno-objects
OBJECT_STORE_PUBLIC_ENDPOINT=https://storage.echno.in
```

The bucket name has no dots, so this is not a TLS or SNI problem. It is purely addressing style.

## Not a duplicate of the object-store issues already open

Both were checked before filing:

[...]

PR #560, opened 2026-08-29
Give an inspection somewhere to keep its evidence

### LINKED ISSUE #535: Inspections cannot carry evidence attachments
## Requested

Anand, 28 Aug 2026. ClickUp `86d45v86y`.

Inspections have nowhere to attach evidence. For a Compliance Inspection that stands in for a permit, a certificate or a statutory approval, the record is not usable as an audit trail without the document itself.

Feature request, not a bug. Filing it with the current state, because this one is genuinely small and should not be estimated as if it were the asset-documents work in #533.

## Current state

The `attachment` table is already generic. It is keyed on `entity_type` and `entity_id`, with dedicated FK columns only for the entities that were wired up (`issue_id`, `project_id`, `material_id`, `indent_item_id`). `entity_type` is a free-form `String`, not an enum, and it is supplied by the caller as a request parameter, for example `IssueControllerWeb.java:175`:

```java
@RequestParam(value = "entityType", required = false, defaultValue = "ISSUE_ATTACHMENTS") String entityType
```

Live values on staging:

```
CLOCK_EVENT_CHECK_IN, ISSUE_ATTACHMENTS, ORGANIZATION, ORGANIZATION_LOGO,
PROJECT_ATTACHMENTS, TASK, TASK_ATTACHMENTS
```

So the storage layer needs nothing new. `INSPECTION_ATTACHMENTS` follows the existing pattern, and the Issue attachment endpoints are the template to copy.

Note the dependency: attachment uploads on the compose staging are currently failing outright, for a reason unrelated to inspections. See #532. There is no point testing this feature until that is in.

## Rough shape of the work

1. Inspection attachment endpoints on the inspection controller, mirroring the Issue ones, scoped by tenant the same way.
2. Attachment list and upload on the inspection detail page in `echno-web`, below the checklist, as the ticket suggests.
3. Checklist-item level evidence, if we want point 6. That needs an `entity_id` pointing at the checklist item rather than the inspection, which the generic table already supports.

Two parts of the ticket are decisions rather than work, and I would settle them before starting:

- **Evidence-required compliance types.** Anand is explicit that attachments should not be mandatory in general, but that specific compliance types can be configured to require them. That is a per-type flag we do not currently have.
- **Immutability after completion.** Attachments on a passed inspection are part of the audit record. Deciding whether they can be replaced, and whether replacement leaves a trail, is the same question we owe the asset documents, so worth answering once for both.

Anand's last point is worth carrying into the UI as written: an AI-suggested inspection is a recommendation, and it must not read as proof of compliance. The evidence section should make the distinction visible rather than leaving it implied.

abhijithanandan:
Closes #535

Anand's 28 Aug QA pass (ClickUp `86d45v86y`) asks for evidence on inspections. The issue expected this to be nearly free, on the strength of what #545 found. Most of it was. One thing was not, and it is the reason inspections had no attachments in the first place.

## What #545 already gave, and what it could not

Checked before assuming, as the issue asked.

**Free, exactly as expected.** `attachment.entity_type` is a free-form `String`, the upload, presign, register, list and delete endpoints all take the type and the id as path variables, and the storage folder is derived as the part of the type before the first underscore. `INSPECTION_EVIDENCE` therefore lands files under `inspection/` with **no change to the storage layer at all**.

**Also free: what a file is and when it lapses.** #545 put `document_type`, `issued_on` and `expires_on` on the **shared** attachment table specifically because inspections wanted them, and `PATCH /api/v1/attachment/web/attachmentId/{id}/document` already records all three for any entity through a tenant-scoped finder. That endpoint serves inspection evidence unchanged; this PR adds nothing to it. The derived `expired` and `daysUntilExpiry` come along too, so a permit that lapsed last week reads as lapsed rather than sitting in the database marked valid.

**Not free, and not mentioned anywhere in the issue: the key.** `attachment.entity_id` is `BIGINT`. `Inspection`, `InspectionCheckItem`, `InspectionDefect` and `Ncr` are all `@GeneratedValue(strategy = UUID)`. A UUID does not fit a BIGINT, so an inspection could not be named as an attachment's owner by any means. That is the actual reason this module has never had attachments, and it is the only schema work here.

## The one migration

`068-add-attachment-entity-uuid.xml`. `entity_uuid UUID` goes on the **shared** table, for the reason #545 widened it rather than giving assets a documents table: the plumbing is already generic, so `Ncr`, checklist templates and every future UUID-keyed record inherit the fix instead of reimplementing it.

`entity_id` therefore stops being `NOT NULL`, and a check constraint takes over from it:

```sql
CHECK ((entity_id IS NOT NULL) != (entity_uuid IS NOT NULL))
```

A row names **exactly one** of the two. Without that, a file could arrive naming neither and be invisible to every read path while still occupying storage, which is strictly worse than the `NOT NULL` it replaces. Every existing row satisfies it on day one, because every existing row has a numeric id and no UUID. The changeset carries a rollback that puts the `NOT NULL` back.

`AttachmentOwner` states the same rule in Java, in a compact constructor, so a path that gets it wrong fails at the call rather than at the insert. It is also the only thing the four upload and read paths now take, so the key type stops being visible past that boundary and the four methods are not duplicated once per key.

## Where evidence attaches, and why not lower

**Against the inspection, and only against the inspection.**

`InspectionService.update` clears the check items and the defects and rebuilds both from the payload, with `orphanRemoval`, so **every child row is deleted and reinserted under a fresh UUID each time an inspector presses save**. A file keyed on a check item id or a defect id would be orphaned by the next edit. This is the trap #543 hit with photo annotations and solved by keying on the photograph rather than the defect row, and the `ncrs` changelog already records it as the reason an NCR keeps a scalar `defectId` rather than a foreign key.

`evidenceFiledAgainstTheInspectionSurvivesAnUpdateThatReissuesEveryChildId` pins it rather than asserting it in a comment: it files one attachment against the inspection and one against a check point, edits the inspection, and shows the check point's evidence left naming a row that no longer exists while the inspection's is untouched. Mutating that test to file the inspection's own evidence against the check point (M9 below) fails it, which is the design decision being tested rather than described.

The domain agrees with the constraint, which is why this is not a workaround. The case the ticket exists for is a compliance inspection standing in for a statutory approval, and its permit is evidence of **the whole inspection**, not of one line on its checklist. Per-check-point and per-defect photographs already have a home in the `photos` element collections, which a rebuild carries through by value and which #543 marks up on top. What did not exist anywhere was a place for the inspection's own paperwork with a document type and an expiry on it.

**Point 6 of the ticket (per-check-point evidence) is therefore not here.** Doing it properly needs the check item to survive a save under a stable identity, which means changing the inspection update contract so the client can name the check points it is re-sending. That is a contract decision, not an implementation detail, and guessing it would be the expensive mistake this PR exists to avoid.

## The assumption about evidence after a pass

Stated rather than blocked on, as asked. **Add always, remove never once judged.**

- **Adding is allowed in every state, including passed.** The certificate a compliance inspection stands in for routinely arrives days after the work was signed off. That is the normal shape of a statutory approval, and freezing additions at the verdict would make the record unfinishable, which defeats the point of the feature.
- **Removal is refused once the inspection is `passed`, `passed-with-remarks` or `failed`.** Those files were part of the record when somebody reached that conclusion; deleting one afterwards leaves a verdict whose basis is gone with nothing to say it ever existed. A file that turns out to be wrong is superseded by uploading the correct one, in the same spirit as #529's stock adjustment corrected by a further adjustment rather than by an edit.
- **`completed` and `cancelled` are deliberately not frozen.** Completed means the work is done but no verdict has been given, and a cancelled inspection concluded nothing, so neither has files that are the basis of anything.

The refusal names the alternative rather than just saying no, matching the convention #545 and #529 settled on.

The other decision the issue raised, **evidence-required compliance types**, is left out: it needs a per-type flag we do not have, and Anand is explicit that attachments are not mandatory in general. Nothing here blocks adding it.

## Endpoints

On the inspection controller, not the generic attachment one, because that controller keys on a numeric entity id and because the removal rule above is an inspection rule rather than an attachment rule.

| | |
|---|---|
| `GET /inspections/web/{id}/evidence` | read, `canRead()` |
| `POST /inspections/web/{id}/evidence` | multipart, `canManageInspections()` |
| `POST /inspections/web/{id}/evidence/presign` | step one of direct-to-storage |
| `POST /inspections/web/{id}/evidence/register` | step two, storage verified before recording |
| `DELETE /inspections/web/{id}/evidence/{attachmentId}` | frozen after a verdict |

Presign and register are included rather than left for later: the deployed frontend already uploads through that path (`use-direct-attachment-upload.ts`, #181), and a scanned approval or a site photograph routinely exceeds what the streaming path allows.

Every one resolves the inspection through `findByIdScoped` first, so a caller in one organization gets a 404 rather than acting on another tenant's record.

**One hole closed while passing through.** `AttachmentService.deleteAttachment(Long)` resolves by numeric id alone, which the file has carried a `NOTE (phase0 authz lockdown)` about since the lockdown: a member of org A can delete org B's attachment by numeric id. The evidence path does not use it. `deleteAttachmentOf(owner, id)` goes through the tenant-scoped finder #545 added and then checks the file is actually the named record's, so the id in the path has to agree with the record in the path. The old method is left alone rather than changed underneath its existing callers.

## Not in this PR

**Web work.** The evidence list and upload on the inspection detail page, and the document type and expiry fields on it. Anand's last point is worth carrying into that as written: an AI-suggested inspection is a recommendation and must not read as proof of compliance, so an empty evidence list on a `suggested` inspection is the normal state and the section should say so rather than leave it implied. The `GET` endpoint's OpenAPI description says as much.

**#532.** Attachment uploads on the compose staging are failing for an unrelated reason, so this cannot be exercised on `echno.in` until that lands. Nothing here depends on it.

## Tests, and that each one fails without its fix

Every new test is plain JUnit and Mockito with no Spring context, except the two added to the **existing** `InspectionServiceIT` so they reuse its cached context rather than building another. That matters in this repo: the test JVM is capped at 1 GB with no `forkEvery`, so a new distinct test configuration is a cached Spring context that makes an unrelated test fail on heap. `AttachmentRepository` is autowired into that class without touching the `@Import` list, because `@DataJpaTest` enables every Spring Data repository already, so the context cache key is unchanged and the four inspection ITs still share one context.

Nine mutations, each applied and reverted around one run on `.116`. Every one is killed.

| # | The fix removed | Test that then fails |
|---|---|---|
| M1 | the "exactly one key" rule on `AttachmentOwner` | `aRecordIsKeyedByItsNumericIdOrItsUuid_neverBoth`, `aRecordThatNamesNeitherKeyIsRefused` |
| M2 | deriving the storage folder from the type before the first underscore | `theStorageFolderIsTheTypeBeforeTheFirstUnderscore`, `evidenceIsFiledAgainstTheInspectionsUuid`, `presigningEvidenceGoesToTheInspectionFolderUnderTheInspectionsUuid` |
| M3 | the refusal to remove evidence from a judged inspection | `evidenceCannotBeRemovedOnceTheInspectionHasBeenJudged`, all three verdicts |
| M4 | treating completed and cancelled as unjudged | `evidenceCanBeRemovedBeforeTheInspectionIsJudged` for `COMPLETED` and `CANCELLED` |
| M5 | allowing evidence to be added after a pass | `evidenceCanStillBeAddedOnceTheInspectionHasPassed` |
| M6 | resolving the inspection through the tenant-scoped finder before touching files | `anInspectionOfAnotherTenantHasNoEvidenceToRead`, `evidenceCannotBeAttachedToAnInspectionOfAnotherTenant` |
| M7 | the owner check on the scoped delete | `refusesAFileFiledAgainstADifferentRecord`, `refusesAFileFiledUnderADifferentEntityType` |
| M8 | the check constraint in changeset 068 | `anAttachmentMustBeFiledAgainstExactlyOneOfTheTwoKeys` |
| M9 | filing evidence against the inspection rather than the check point | `evidenceFiledAgainstTheInspectionSurvivesAnUpdateThatReissuesEveryChildId` |

Full suite green on `.116`. `EndpointAuthorizationTest`, `MultipartPayloadValidationTest`, `UnboundedRepositoryReadTest` and #555's new `PaginationParameterBoundTest` all pass: the evidence list is a bare array like the asset documents one, with no paging parameters of its own, so it declares neither `pageNo` nor `pageSize`.

## Migration number

`068`. `061` was #540, `062` is claimed by #542 (still open), `063` to `067` went in with #545 and #543. Rechecked against `origin/development` and against every open PR's diff immediately before pushing.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/dto/AttachmentOwner.java
+++ b/src/main/java/org/tornotron/echno_backend/common/dto/AttachmentOwner.java
@@ -1,0 +1,59 @@
+package org.tornotron.echno_backend.common.dto;
+
+import java.util.UUID;
+
+/**
+ * The record a file is filed against: its attachment type, and its key.
+ *
+ * <p>Most of the application keys its rows with a {@code Long}, and the attachment table was
+ * built for exactly that. The inspection module keys everything with a {@code UUID}, and a UUID
+ * does not fit a BIGINT, so an inspection had nowhere to put a file. Rather than duplicating the
+ * upload, presign, register and list paths once per key type, each takes one of these and the
+ * key type stops being visible past the boundary.
+ *
+ * <p>Exactly one of the two keys is set, which is the same rule the database enforces with the
+ * {@code ck_attachment_entity_key} check constraint. The two factory methods are the only way to
+ * build one, so the rule holds by construction.
+ *
+ * @param entityType The attachment entity type, for example ISSUE_ATTACHMENTS or INSPECTION_EVIDENCE
+ * @param entityId   The record's numeric id, or null when it is keyed by UUID
+ * @param entityUuid The record's UUID, or null when it is keyed by a numeric id
+ */
+public record AttachmentOwner(String entityType, Long entityId, UUID entityUuid) {
+
+    public AttachmentOwner {
+        if (entityType == null || entityType.isBlank()) {
+            throw new IllegalArgumentException("An attachment must state the type of record it belongs to");
+        }
+        if ((entityId == null) == (entityUuid == null)) {
+            throw new IllegalArgumentException(
+                    "An attachment must be filed against exactly one of a numeric id and a UUID");
+        }
+    }
+
+    /** A record keyed by a numeric id. */
+    public static AttachmentOwner of(String entityType, Long entityId) {
+        return new AttachmentOwner(entityType, entityId, null);
+    }
+
+    /** A record keyed by a UUID. */
+    public static AttachmentOwner of(String entityType, UUID entityUuid) {
+        return new AttachmentOwner(entityType, null, entityUuid);
+    }
+
+    /**
+     * The storage folder files for this record go in: the part of the entity type before the
+     * first underscore, lower-cased. {@code INSPECTION_EVIDENCE} therefore lands under
+     * {@code inspection/} with nothing in the storage layer needing to know it exists.
+     *
+     * @return The folder name
+     */
+    public String folder() {
+        return entityType.split("_", 2)[0].toLowerCase();
+    }
+
+    /** The key as it reads in a message to the caller. */
+    public String keyAsText() {
+        return entityId != null ? entityId.toString() : entityUuid.toString();
+    }
+}
--- a/src/main/java/org/tornotron/echno_backend/common/entity/Attachment.java
+++ b/src/main/java/org/tornotron/echno_backend/common/entity/Attachment.java
@@ -16,6 +16,7 @@ import org.tornotron.echno_backend.user.User;
 
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import java.util.UUID;
 
 /**
  * Generic attachment entity for storing file references across different modules.
@@ -30,6 +31,7 @@ import java.time.LocalDateTime;
 @NoArgsConstructor
 @Table(name = "attachment", indexes = {
         @Index(name = "idx_attachment_entity", columnList = "entity_type, entity_id"),
+        @Index(name = "idx_attachment_entity_uuid", columnList = "organization_id, entity_type, entity_uuid"),
         @Index(name = "idx_attachment_expiry", columnList = "organization_id, entity_type, expires_on")
 })
 @Filter(name = "orgFilter", condition = "organization_id = :organizationId")
@@ -46,10 +48,27 @@ public class Attachment implements TenantScopedEntity {
     private String entityType;
 
     /**
-     * The ID of the entity this attachment belongs to.
+     * The numeric id of the entity this attachment belongs to, for the modules keyed that way.
+     *
+     * <p>Null on a file filed against a UUID-keyed record; {@link #entityUuid} carries the key
+     * instead. A database check constraint requires exactly one of the two, so a file cannot
+     * arrive naming neither and become invisible to every read while still occupying storage.
      */
-    @Column(name = "entity_id", nullable = false)
+    @Column(name = "entity_id")
     private Long entityId;
+
+    /**
+     * The UUID of the entity this attachment belongs to, for the modules keyed that way.
+     *
+     * <p>The inspection module is keyed by UUID throughout, and a UUID does not fit the BIGINT
+     * {@link #entityId}, so before this column an inspection had nowhere to file evidence at all.
+     * It sits on the shared table rather than inspections getting an evidence table of their own,
+     * for the reason the document columns do: the upload, presign, register, list and delete
+     * plumbing is already generic, so every UUID-keyed module inherits the fix rather than
+     * reimplementing it.
+     */
+    @Column(name = "entity_uuid")
+    private UUID entityUuid;
 
     /**
      * The unique key/path of the file in the storage bucket.
--- a/src/main/java/org/tornotron/echno_backend/common/repository/AttachmentRepository.java
+++ b/src/main/java/org/tornotron/echno_backend/common/repository/AttachmentRepository.java
@@ -9,6 +9,7 @@ import org.tornotron.echno_backend.common.entity.AttachmentDto;
 import java.time.LocalDate;
 import java.util.List;
 import java.util.Optional;
+import java.util.UUID;
 
 /**
  * Repository for managing Attachment entities.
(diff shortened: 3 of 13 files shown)


abhijithanandan:
@coderabbitai review

PR #561, opened 2026-08-29
Ban database I/O from mapping hooks, and enforce it

abhijithanandan:
Item A of #522, and only item A.

The audit on that issue concluded the converter component does not need reworking: the hand-rolled-or-mapper question was settled by the MapStruct migration, there is no per-request reflection anywhere on the conversion path, and finance and inspection, which the issue names as the place to start, are the two cleanest modules in the repo. What is real is three mappers doing per-row database I/O inside an `@AfterMapping` hook, which is invisible at the call site because the mapper looks like a pure conversion. That is what this changes.

## What it cost

`MaterialMapper.fillStock` asked `InventoryService` for two aggregates keyed on the material id. The generated mapper calls the hook once per mapped material, so the cost followed the row count and not the request, and a service mapping a page in a stream read as free:

```java
return materialRepository.findAll(pageable)
        .map(material -> materialMapper.toDto(material));
```

`IndentItemDto` carries a full `MaterialDto` and `IndentItemMapper` declares `uses = MaterialMapper.class`, so an indent paid the same cost once per line, and a page of indents multiplied that by the page size. `StorageLocationMapper` did the same thing with a `COUNT DISTINCT`, once per location, on four listing paths.

`hibernate.default_batch_fetch_size` does not soften any of it. It batches lazy association loads; these were explicit repository calls it cannot see. #537 capped `GET /materials` at 500 rows, which bounds the row count but not the per-row cost.

## What it costs now

Counted at `CurrentStockRepository`, so the number is statements issued, not service calls:

| Path | Before | After |
|---|---|---|
| one material | 2 | 1 |
| page of 50 materials | 100 | 1 |
| ten-line indent | 20 | 1 |
| page of 50 storage locations | 50 | 1 |

Two grouped reads replace the per-row ones: one totalling quantity and value per material, one counting distinct materials per location. The result is handed to the mapper as a MapStruct `@Context`, which is the shape `MaterialMapper.toWithStockDto` always had and which three call sites in `MaterialService` already used. Because it is a context, `IndentItemMapper` and `IndentMapper` pass the same instance down rather than each fetching their own, so DTO depth stops multiplying query count.

A material or a location that holds nothing produces no group. The lookup supplies the zero the per-row `COALESCE(SUM(...), 0)` used to return, so that case is unchanged too.

`countDistinctMaterialsByStorageLocationIdAndOrganizationId` had no caller left and is gone with it. That overlaps #550 by one method; happy to drop it from here if it is easier to take there.

## No response changed shape

This touches DTO-producing code that a lot of endpoints run through, and `echno-core` maintains its contract by hand with no codegen, so a shape change would not fail to compile anywhere. I checked it rather than reasoning about it: compiling `origin/development` and this branch and diffing all 52 generated mapper implementations, the entire difference is the threaded parameter and where three values come from.

```
-        fillStock( material, materialDto );
+        materialDto.setCurrentStock( stock.currentStockOf(material.getId()) );
+        materialDto.setStockValue( stock.stockValueOf(material.getId()) );

-        fillItemsCount( storageLocation, storageLocationDto );
+        storageLocationDto.setStorageItemsCount( itemCounts.itemCountOf(storageLocation.getId()) );
```

Every other line is the parameter being passed along. No field was added, removed, renamed or retyped, and every field is still set from the same source in the same order. `CurrentStockBatchReadIT` then checks against a real CockroachDB that the grouped reads agree with the per-row reads they replaced, for the stocked case and the unstocked one.

## The guardrail

Neither mapper had anything in the repo saying it should not do this, which is why the pattern survived both the MapStruct migration and the unbounded-read sweep. `MapperDatabaseAccessTest` makes it a build failure, sharing the existing `@AnalyzeClasses` import with the other architecture tests so no second class graph sits in the capped test heap.

Two checks. The first is cheap and blunt: a `@Mapper` class may not depend on a Spring Data repository at all, which fails at the field rather than at the hook. The second walks the method-call graph out of every `@BeforeMapping` and `@AfterMapping` and fails where a database access is reachable, reporting the shortest path. Transitive rather than direct, because neither offender named a repository in the hook: `MaterialMapper` went through `InventoryService`, so a direct-call rule would have passed the code it exists to prevent. Reachability follows method calls only and only through this codebase's classes, which keeps it precise; `AttachmentMapper` signs a storage URL from its hook and is correctly not reported, because signing is not a database read.

Run against the code as it stood before the fix, both fail:

```
Field <StorageLocationMapper.currentStockRepository> has type <CurrentStockRepository>

Expecting empty but was:
  MaterialMapper.fillStock
      -> InventoryService.getAggregateStock
      -> CurrentStockRepository.sumCurrentQuantityByMaterial
  StorageLocationMapper.fillItemsCount
      -> CurrentStockRepository.countDistinctMaterialsByStorageLocationIdAndOrganizationId
```

## Draft #538

Folded in and superseded. Its commit is the first of the three here, so the measurement lands as it was written and the next commit flips the assertions, which is what that PR said should happen when item A landed. #538 can be closed unmerged.

## Still open on #522

Items B, C and D are not in here and the issue stays open for them: list projections for `OrganizationDto`, `ProjectDto` and `IndentDto`; the seven remaining hand-written mappers; and the 28 hand-parsed request bodies, which is the one that pays off soonest because it is the half the published spec cannot describe today.

`AttachmentMapper` is also untouched. It presigns once per attachment and is a `uses` dependency of six mappers, so it fans out further than anything here, but it costs CPU rather than queries and moving it means changing six mappers and their whole call chain. It is not item A and it is not what the guardrail bans.

## Verification

Full suite on lab `.116`: 1138 tests, 0 failures, including the Testcontainers integration tests against CockroachDB v26.2.4.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/indent/IndentService.java
+++ b/src/main/java/org/tornotron/echno_backend/indent/IndentService.java
@@ -27,13 +27,17 @@ import org.tornotron.echno_backend.indentItem.dto.IndentItemUpdateDto;
 import org.tornotron.echno_backend.indent.dto.IndentCreationDto;
 import org.tornotron.echno_backend.indent.dto.IndentDto;
 import org.tornotron.echno_backend.indent.enums.IndentStatus;
+import org.tornotron.echno_backend.inventoryTransaction.InventoryService;
+import org.tornotron.echno_backend.inventoryTransaction.MaterialStockLookup;
 import org.tornotron.echno_backend.material.Material;
 import org.tornotron.echno_backend.material.MaterialRepository;
 import org.tornotron.echno_backend.project.Project;
 import org.tornotron.echno_backend.project.ProjectRepository;
 
+import java.util.Collection;
 import java.util.List;
-import java.util.stream.Collectors;
+import java.util.Objects;
+import java.util.stream.Stream;
 
 /**
  * CRUD for material indents and their line items.
@@ -56,13 +60,14 @@ public class IndentService {
     private final IndentItemMapper indentItemMapper;
     private final DocumentNumberAllocator documentNumberAllocator;
     private final TransactionRetryTemplate retryTemplate;
+    private final InventoryService inventoryService;
 
     public IndentService(IndentRepository indentRepository, IndentItemRepository indentItemRepository,
                          MaterialRepository materialRepository, IndentMapper indentMapper,
                          TenantEntityHelper tenantEntityHelper, EmployeeRepository employeeRepository,
                          ProjectRepository projectRepository, IndentItemMapper indentItemMapper,
                          DocumentNumberAllocator documentNumberAllocator,
-                         TransactionRetryTemplate retryTemplate) {
+                         TransactionRetryTemplate retryTemplate, InventoryService inventoryService) {
         this.indentRepository = indentRepository;
         this.indentItemRepository = indentItemRepository;
         this.materialRepository = materialRepository;
@@ -73,6 +78,7 @@ public class IndentService {
         this.indentItemMapper = indentItemMapper;
         this.documentNumberAllocator = documentNumberAllocator;
         this.retryTemplate = retryTemplate;
+        this.inventoryService = inventoryService;
     }
 
     // ==================== Indent CRUD ====================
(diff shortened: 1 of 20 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #564, opened 2026-08-29
Make the compliance AI timeouts configuration like everything around them

### LINKED ISSUE #552: Compliance AI connect and read timeouts are hardcoded while everything around them is configuration
## What is wrong

`OpenAiCompatibleComplianceService.requestFactory()` hardcodes its connect and read timeouts, while every other knob in the same method and the same class comes from `ComplianceAiProperties`.

Noticed while reviewing #542. Small, and not urgent.

## What I verified (branch `feat/482-async-compliance-generation`, 29 Aug 2026)

```java
private SimpleClientHttpRequestFactory requestFactory() {
    SimpleClientHttpRequestFactory factory = new SimpleClientHttpRequestFactory();
    factory.setConnectTimeout(Duration.ofSeconds(10));
    factory.setReadTimeout(Duration.ofSeconds(60));
    String proxyHost = props.getProxyHost();
    ...
}
```

Both timeouts are literals. Everything around them is configuration: `props.getProxyHost()` and `props.getProxyPort()` two lines down, and `getBaseUrl()`, `getApiKey()`, `getModel()`, `getMaxTokens()`, `getTemperature()` and `getBatchSize()` in the callers. `application.yml` exposes each of those as an environment override; there is no `COMPLIANCE_AI_*_TIMEOUT`.

## Why it matters

Less than it did. #542 batches the rule catalogue into groups of `batch-size` (default 10, `COMPLIANCE_AI_BATCH_SIZE`), so a single model call now carries at most ten rules and 60 seconds is comfortable. Before batching, a large catalogue in one call could plausibly have run past it.

So this is consistency, not a live risk: the one dimension we cannot tune without a rebuild is the one that decides whether a slow provider fails the whole job. Two things could make it bite again. Raising `batch-size` for a provider that rate-limits per key puts more rules back into one call. And a different endpoint (we already switch between DO Gradient and the lab-proxied route) may simply be slower than the one it was measured against.

## What a fix would involve

Add `connect-timeout-seconds` and `read-timeout-seconds` to `ComplianceAiProperties` with the present values (10 and 60) as defaults, expose them in `application.yml` as `${COMPLIANCE_AI_CONNECT_TIMEOUT_SECONDS:10}` and `${COMPLIANCE_AI_READ_TIMEOUT_SECONDS:60}`, and read them in `requestFactory()`. Behaviour is unchanged unless someone sets them.

Do this after #542 merges, not on that branch. It is not worth widening an open PR for.

abhijithanandan:
Closes #552.

## What changed

`requestFactory()` had the connect and read timeouts as literals while every other knob in the same method came from `ComplianceAiProperties`, `props.getProxyHost()` and `props.getProxyPort()` two lines below them. Both are now properties with the values they already had, exposed as `COMPLIANCE_AI_CONNECT_TIMEOUT_SECONDS` and `COMPLIANCE_AI_READ_TIMEOUT_SECONDS`. Behaviour is unchanged unless someone sets them.

## Why the read timeout is the one worth moving

It is sized against `batch-size`, not against the catalogue. At roughly 1.7 seconds a rule the default ten-rule batch takes about 20 seconds and sits at a third of the 60-second budget, which is exactly why the value can stay still while the catalogue grows. The two numbers have to move together, though: twenty rules in a batch is about 34 seconds, over half the budget, and past thirty a batch no longer fits at all. A slower endpoint moves the same line without anyone touching the batch size, and we already switch between DO Gradient and the lab-proxied route. Neither case should need a rebuild.

The connect timeout is less interesting but has the same shape: where egress goes through a forward proxy the connect is to the proxy, and a loaded proxy can be slower to accept than the endpoint behind it.

## Test

`readTimeoutComesFromConfiguration` is behavioural rather than a getter assertion, deliberately. A property nothing consumes would satisfy a getter test and change nothing about the call, so the test has to observe the value reaching the request factory. It points the client at a loopback listener that accepts the connection and then never writes a byte, sets the read timeout to one second, and bounds the elapsed time.

|---|---|---|
| `readTimeoutComesFromConfiguration` | FAILED, 78 s build (the call sat on the hardcoded 60 s, elapsed bound of 15 s missed) | PASSED, ~1 s |
| `timeoutsDefaultToThePreviouslyHardcodedValues` | n/a (property did not exist) | PASSED |

Verified on `lab-node-116-f` by reverting only the `setReadTimeout` line in the service and re-running.

No changelog in this PR.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/compliance/ai/ComplianceAiProperties.java
+++ b/src/main/java/org/tornotron/echno_backend/compliance/ai/ComplianceAiProperties.java
@@ -76,4 +76,28 @@ public class ComplianceAiProperties {
 
     /** Proxy port used only when {@code proxyHost} is set. */
     private int proxyPort = 3128;
+
+    /**
+     * How long to wait for the TCP connection to the inference endpoint, in seconds.
+     *
+     * <p>Ten seconds is generous for a connect: either the endpoint answers the handshake
+     * quickly or something is wrong with the route. It is configuration rather than a
+     * constant because the route is not always direct. Where egress goes through a forward
+     * proxy the connect is to the proxy, and a loaded proxy can be slower to accept than the
+     * endpoint behind it.
+     */
+    private int connectTimeoutSeconds = 10;
+
+    /**
+     * How long to wait for the endpoint's response after the request is sent, in seconds.
+     *
+     * <p>Sixty is sized against {@code batchSize}, not against the catalogue: a run costs
+     * about 1.7 seconds a rule, so the default ten-rule batch takes roughly 20 seconds and
+     * sits at a third of this. That margin is the whole reason this value can stay still
+     * while the catalogue grows, and it is also why the two numbers have to move together.
+     * Raising {@code batchSize} to twenty puts a batch at about 34 seconds, over half the
+     * budget; past thirty rules a batch no longer fits at all. A slower endpoint moves the
+     * same line without anyone touching the batch size.
+     */
+    private int readTimeoutSeconds = 60;
 }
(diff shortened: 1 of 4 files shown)


abhijithanandan:
@coderabbitai review

PR #567, opened 2026-08-29
Take the regularization approver from the session, not the query string

### LINKED ISSUE #562: Attendance regularization takes the approver as a request parameter, so a caller records whoever they like as having approved it
Found while landing #559 (self-approval on stock adjustments and construction invoices), and deliberately left out of that PR because fixing it properly is a contract change, not a guard.

## What is wrong

Both regularization controllers take the approver from the query string:

```java
// AttendanceRegularizationControllerWeb.java and AttendanceRegularizationController.java
public ResponseEntity<AttendanceRegularizationDto> process(
        @PathVariable Long id,
        ...
        @RequestParam String approvedBy,
        @RequestParam(required = false) Long approvedById) {
    return ResponseEntity.ok(regularizationService.processRegularization(id, dto, approvedBy, approvedById));
}
```

Nothing ties `approvedBy` or `approvedById` to the authenticated caller. Whoever can reach the endpoint records **any name and any id** as the approver, and the stored record then says someone else approved it.

Verified on `development` at the time of writing: both the `/web` controller and its twin, identical signatures, value passed straight through to the service.

## Why it matters

The attendance record is what a regularization changes, and the approver is the accountability on that change. An audit asks who approved a corrected attendance entry, and the answer is currently whatever the caller typed. This is weaker than the equivalent paths elsewhere in the codebase:

- **Stock adjustments** now stamp the raiser from the session on create and refuse self-approval (#559, merged). `submittedBy` used to be copied from the request body in the same way this does, and that was fixed as a precondition of the self-approval rule.
- **Construction invoices** resolve the approver from the session and go through the same shared `common/approval/SelfApprovalPolicy`.

So attendance is now the odd one out, and it is the only remaining approval path where the approver is caller-supplied.

## Why #559 did not fix it

Two reasons, both stated in that PR:

1. **It does not post to a ledger**, so it was outside that PR's scope, which was the paths that set a balance.
2. **The self-approval check cannot be applied until this is fixed.** A rule comparing raiser against approver is meaningless when the approver is whatever the caller claims, so the parameter has to come from the session first.

## What a fix involves

1. Resolve the approver from the authenticated session rather than the request, on both controllers.
2. Remove `approvedBy` and `approvedById` from the signature. **This is a breaking API change**: check whether `echno-web` sends them today and sequence accordingly. Unknown properties are ignored on deserialization for request bodies, but these are query parameters, so the usual "the deployed frontend keeps working" argument does not apply automatically. Verify rather than assume.
3. Then apply `SelfApprovalPolicy` the way #559 does, so the same person cannot raise and approve a regularization, with the same recorded `system-admin` exception.
4. Consider whether the two controllers should share one implementation. They are byte-identical here, which is the `/web` twin duplication #522's analysis noted more generally.

## Related

- #559 (the self-approval rule and the shared policy), #548 (the issue behind it)
- #541 tracks client-supplied `status` on create endpoints, which is the same class of defect (trusting the caller for a field the server should own) on a different field

abhijithanandan:
Closes #562.

## What changed

Both regularization controllers took the approver off the query string and passed it straight to the service, so the record said whoever the caller typed had approved a correction to an attendance record. The approver now comes from the authenticated session on both twins, and `approvedBy` / `approvedById` are gone from the signature.

The caller is resolved to their employee record in the current tenant, which is the same resolution `AttendanceService.approveAttendance` already uses next door, so `approvedById` keeps meaning what it means on an attendance record and the web client's "requested by" link still resolves. A caller with no employee record here, a bootstrap administrator for instance, is recorded by their authenticated username instead: still not something they choose per request. Only a caller with no identity at all falls back to `system`, which the endpoint's authorization already rules out.

## The requester had to move too

The rule the issue asks for compares the raiser against the approver, and on this entity the raiser was as forgeable as the approver. Worse: **the web client never sent `requestedById` at all**, so every row written through the UI carries a null requester id and a rule comparing the two would never have fired once. It would have been a control in name only.

So `requestedBy` and `requestedById` are stamped from the session on submit as well, which is the same precondition #559 had to satisfy on `StockAdjustment.submittedBy`. The monthly cap already counted by `requestedBy`, and the value the web client sends for it today is the signed-in employee's name, the same string the session now supplies, so the cap keeps counting the same rows and needs no migration.

## Then the rule, from the shared policy

`common/approval/SelfApprovalPolicy` unchanged, called the way `StockAdjustmentService.approve` calls it: refused outright for an ordinary approver, allowed and logged for `system-admin`. A break-glass self-approval writes its note onto every corrected clock event (`Self-regularized (self-approved: raised and approved by the same person)`). That is this module's equivalent of #559 putting the note in the ledger remarks: the ledger is what a stock figure is explained from, and the clock event is what a corrected day is read from.

Two paths are outside the rule on purpose:

- **Rejection.** It writes nothing to the attendance record, and refusing a self-rejection would leave an employee unable to withdraw a request they filed by mistake.
- **The auto-approve path** inside `submitRequest`, where `regularizationApprovalRequired` is off. A project configured not to require an approval is the tenant's own standing decision that these corrections do not need a second person, which is exactly the case that setting exists for. Same exemption, and same reasoning, as #559 gave the invoice approval threshold.

Requests stored before this name no requester, so there is nothing to compare and they are let through rather than becoming unapprovable.

## Does `echno-web` send these today, and is a coordinated deploy needed?

**It sends them, and no coordinated deploy is needed.** Verified in the clone rather than assumed.

`echno-core`'s `attendanceRegularizationService` puts the identity on the query string on both calls (`src/services/attendance-regularization-service.ts`): `request(dto, requestedBy)` sends `{ requestedBy }`, `process(id, status, approvedBy, rejectionReason)` sends `{ approvedBy }`. Both call sites in `echno-web` (`features/attendance/components/attendance-regularization-card.tsx` and `regularization-management.tsx`) pass `currentEmployee?.name ?? currentEmployee?.employeeId ?? ''`. Neither ever sends `requestedById` or `approvedById`.

The direction of the break matters. A query parameter no handler declares is dropped by Spring, so the deployed client keeps working and the value it sends is simply ignored. What would have broken is the reverse case, a client that stopped sending `approvedBy` while the parameter was still `@RequestParam` and required: that is a 400. Backend first is therefore the safe order, and the frontend can drop the arguments whenever it likes.

This is pinned rather than asserted: `AttendanceRegularizationApproverParamTest` drives both twins through MockMvc with `?approvedBy=Somebody Else&approvedById=99` still attached and expects the same 200 the updated client gets, and the same for `requestedBy` on submit. The one case in that file that fails against the old code is a caller sending neither, which the old contract answered with a 400.

## Two byte-identical controllers: what I chose

**Kept both, and did not merge them.** The shared implementation already exists and it is the service: after this change the controllers hold no logic at all beyond delegation, so the rule lands on both twins whether or not the pair is ever collapsed. Collapsing 32 controller-twin pairs is #522's analysis and is decision-gated on the `@PreAuthorize` question, since the two twins can carry different guards and a shared base class has to settle which one wins. Doing it here, in a security fix, would put that question in the diff that has to be reviewed for something else. The duplication that remains is two annotation blocks, and it is now duplication of documentation rather than of behaviour.

## What the live data shows

Read-only against the compose staging on `.116`. One regularization exists:

| id | requested_by | requested_by_id | approved_by | approved_by_id | requested_at | approved_at |
|---|---|---|---|---|---|---|
| 1 | Echno Admin | 4 | Echno Admin | 4 | 2026-06-02 12:59:07 | 2026-06-02 13:03:18 |

So the one row on staging is a self-approval, four minutes apart: employee 4 raised it against their own attendance record (`attendance.employee_id = 4`) and approved it. There is no orphan approver: id 4 resolves to a real employee, "Echno Admin", of organization 2, and every other regularization column is consistent.

Whether this PR would have refused it comes down to the break-glass role. `employee_org_roles` gives employee 4 `SYSTEM_ADMIN`, so on that evidence it would have gone through as a recorded self-approval rather than being refused. Worth stating precisely: the gate `SelfApprovalPolicy` reads is the Keycloak org group (`ORG_2_ROLE_system-admin`), not that table, so the table is strong evidence and not proof. Either way there is nothing to migrate and no existing record to reinterpret.

## Failing-test table

Each new test run against a reverted baseline (the three main-source files restored to `origin/development`, the tests' call sites given back the query parameters the old contract accepted, sending a forged `"Somebody Else"` / `99L` exactly as the endpoint allowed), then against the fix.

|---|---|---|
| `submitRequest_recordsTheSignedInEmployeeAsTheRequester` | FAILED | PASSED |
| `submitRequest_recordsTheAuthenticatedUsernameWhenTheCallerIsNotAnEmployeeHere` | FAILED | PASSED |
| `processRegularization_recordsTheSignedInEmployeeAsTheApprover` | FAILED | PASSED |
| `processRegularization_theSamePersonCannotRaiseAndApproveAndNothingIsCorrected` | FAILED | PASSED |
| `processRegularization_aBreakGlassSelfApprovalIsAppliedAndTheClockEventsSaySo` | FAILED | PASSED |
| `processRegularization_anApprovalBySomeoneElseIsNotMarkedAsSelfApproved` | FAILED | PASSED |
| `processRegularization_aRequestWithNoRecordedRequesterIdIsStillApprovable` | FAILED | PASSED |
| `AttendanceRegularizationApproverParamTest.aCallerThatSendsNeitherIsServedTheSame` | FAILED | PASSED |

Two further new tests pass either way by design, and are there to pin what must not change:

- `processRegularization_rejectingYourOwnRequestIsNotSubjectToTheRule` (the rule must not fire on a rejection)
- the three `...IgnoresA{Requested,Approved}ByTheCallerStillSends` cases, which are the compatibility controls: they have to pass on both sides, because that is the whole claim about the deployed client.

Full suite on `lab-node-116-f`: **BUILD SUCCESSFUL**, 1243 tests, 0 failures, 0 errors, branched from `origin/development` at `ef15dad`.

No Liquibase changeset. `requested_by`, `requested_by_id`, `approved_by` and `approved_by_id` all already exist on `attendance_regularization`, and this changes only where their values come from. No new Spring context either: the controller test is `MockMvcBuilders.standaloneSetup`, and the service test is plain Mockito, so the cached-context position in the test JVM is unchanged.

## If you would rather have it another way

If self-rejection should also be refused, move the policy call above the `status` check in `processRegularization` and drop the `APPROVED` condition; one line, and one test flips. If the auto-approve path should be subject to the rule, that belongs in `submitRequest` rather than here, and it would mean a project that auto-approves can no longer auto-approve anything, which is why I did not.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/attendance/AttendanceRegularizationController.java
+++ b/src/main/java/org/tornotron/echno_backend/attendance/AttendanceRegularizationController.java
@@ -41,7 +41,9 @@ public class AttendanceRegularizationController {
     @Operation(
             summary = "Submit a regularization request",
             description = "Files a request to correct an attendance record, naming the missing clock "
-                    + "events and, optionally, the events that should be added in their place."
+                    + "events and, optionally, the events that should be added in their place. The "
+                    + "authenticated caller is recorded as the requester, and cannot then approve the "
+                    + "request themselves unless they hold the system-admin role."
     )
     @ApiResponses({
             @ApiResponse(responseCode = "201", description = "Regularization request submitted"),
@@ -50,11 +52,9 @@ public class AttendanceRegularizationController {
             @ApiResponse(responseCode = "404", description = "No attendance record with the given attendanceId")
     })
     public ResponseEntity<AttendanceRegularizationDto> submitRequest(
-            @Valid @RequestBody RegularizationRequestDto dto,
-            @RequestParam String requestedBy,
-            @RequestParam(required = false) Long requestedById) {
+            @Valid @RequestBody RegularizationRequestDto dto) {
         return ResponseEntity.status(HttpStatus.CREATED)
-                .body(regularizationService.submitRequest(dto, requestedBy, requestedById));
+                .body(regularizationService.submitRequest(dto));
     }
 
     @PostMapping("/{id}/process")
@@ -62,20 +62,23 @@ public class AttendanceRegularizationController {
     @Operation(
             summary = "Approve or reject a regularization request",
             description = "Sets the status of a pending regularization request. A rejection should carry "
-                    + "a rejectionReason."
+                    + "a rejectionReason. The authenticated caller is recorded as the approver. Whoever "
+                    + "raised the request cannot approve it: an approval is the second pair of eyes on a "
+                    + "change to an attendance record, so it has to come from someone else. A system "
+                    + "administrator is the one exception, and their self-approval is recorded as one on "
+                    + "the corrected clock events. Rejecting your own request is allowed, since it writes "
+                    + "nothing to the attendance record."
     )
     @ApiResponses({
             @ApiResponse(responseCode = "200", description = "Regularization request processed"),
-            @ApiResponse(responseCode = "400", description = "status is missing or invalid"),
+            @ApiResponse(responseCode = "400", description = "status is missing or invalid, the request has already been actioned, or it is being approved by whoever raised it without the system-admin role"),
             @ApiResponse(responseCode = "403", description = "Caller lacks permission to manage attendance records"),
             @ApiResponse(responseCode = "404", description = "No regularization request with the given id")
     })
     public ResponseEntity<AttendanceRegularizationDto> process(
             @PathVariable Long id,
-            @Valid @RequestBody RegularizationActionDto dto,
-            @RequestParam String approvedBy,
-            @RequestParam(required = false) Long approvedById) {
-        return ResponseEntity.ok(regularizationService.processRegularization(id, dto, approvedBy, approvedById));
+            @Valid @RequestBody RegularizationActionDto dto) {
+        return ResponseEntity.ok(regularizationService.processRegularization(id, dto));
     }
 
     @GetMapping("/pending")
--- a/src/main/java/org/tornotron/echno_backend/attendance/AttendanceRegularizationControllerWeb.java
+++ b/src/main/java/org/tornotron/echno_backend/attendance/AttendanceRegularizationControllerWeb.java
@@ -41,7 +41,9 @@ public class AttendanceRegularizationControllerWeb {
     @Operation(
             summary = "Submit a regularization request",
             description = "Files a request to correct an attendance record, naming the missing clock "
-                    + "events and, optionally, the events that should be added in their place."
+                    + "events and, optionally, the events that should be added in their place. The "
+                    + "authenticated caller is recorded as the requester, and cannot then approve the "
+                    + "request themselves unless they hold the system-admin role."
     )
     @ApiResponses({
             @ApiResponse(responseCode = "201", description = "Regularization request submitted"),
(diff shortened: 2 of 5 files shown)


abhijithanandan:
@coderabbitai review

PR #568, opened 2026-08-29
Refuse a status a create could not have transitioned into

abhijithanandan:
Continues #541. #544 closed the two entries that were not hygiene; this closes the rest of the survey, under the same rule and with no second mechanism: a create may not reach a state whose transition carries a check or an event, the field stays on the payload typed as its enum and optional, and the schema's `allowableValues` names what is real.

## Refused

| Entity | What create takes now | What is behind the state it can no longer reach |
|---|---|---|
| StockAdjustment | `draft` only, on create **and** update | `approve` refuses a self-approval (#559), writes an `InventoryTransaction` per line, moves the balance, and stamps the approver and the posting time |
| SiteTransfer | `PENDING` only, and defaults to it | `PATCH /{id}/status` records what the receiving site took delivery of, and a different authority holds it |

**StockAdjustment (rank 3).** Its status was a free string copied straight onto the entity by both create and update. `"APPROVED"` on create is the case the issue names, and #559 makes it more than untidy: the self-approval check it added guards the one transition, so a body that wrote the status itself sidestepped the control by not using the transition. The posted state is worse still, because `POSTED_STATUS` is what every list and detail view reads as dealt with, while `processedAt` (which is what actually freezes the document) stayed null: a document reading as posted with no approver on record, no ledger entries and the balance untouched.

Applied to `update` as well as `create`, because both write the same header through `applyHeaderFields` and a document can be edited right up until it is posted. Gating only create would have left the same door one call further along.

The status stays a `String` rather than becoming an enum. The column deliberately holds the frontend's own lower-case vocabulary (documented on the entity, alongside `write_off` and `physical_count`), the module has exactly two producers of the value, and the issue's suggestion of an enum plus a status endpoint is a build rather than a guard. `draft` is matched case-insensitively.

**SiteTransfer (rank 4).** `PARTIALLY_TRANSFERRED` and `COMPLETED` both say the far end has received the materials, which is recorded through `PATCH /site-transfers/{id}/status`. On the token-authority controller that endpoint wants `site-transfer:update` while create wants `site-transfer:create`, so a caller who may only raise transfers could record their own receipt of one. Stock moves on creation either way, through `SiteTransferCreatedEvent`, so what a transfer issued as `COMPLETED` left behind was a movement standing as confirmed by nobody.

The field is now typed as `SiteTransferStatus` and optional, defaulting to `PENDING`, which is what the web form already sends and what `echno-core` documents as the typical initial value.

## Deleted

**`EmployeeCreationDto.status` (rank 7).** Verified still correct: `EmployeeService.EmployeeObjectMapper` does not read it, and nothing else does. It was `@NotBlank`, so a caller was made to send a value nothing read and a caller who left it out was refused for it, which is also why `echno-core`'s `createEmployeeToJson` (which sends no status) would have been rejected by this endpoint. An employee's status is written through the employee patch path, which is the one place that sets it. Deleting relaxes the contract in both directions.

## Left alone, deliberately

Not narrowing a field just because it can be narrowed is the other half of #544's rule, so these are stated rather than quietly skipped.

| Entity | Why it stays as it is |
|---|---|
| Expense, Receipt, SubContract, Asset (rank 5) | Free strings with no enum, no state machine, no status endpoint and no event. Written by create and update under the same role, and read only as a list filter. Refusing values would take away frontend-defined vocabulary for no safety gained. The day a workflow goes behind one of them, its create has to be gated in the same change. |
| `ConstructionPaymentService.update` (the adjacent item) | `ConstructionPaymentVoucherStatus` states carry no journal entry on any transition (stated on the enum itself), the only guard is that `CANCELLED` is terminal, and `update` already enforces that. There is no status endpoint, so update is the only way a payment can be recorded as completed at all: refusing the field there would make `COMPLETED` unreachable rather than gated. `create` already forces `PENDING`. |
| Task, WBS, Indent, Vendor, Labour, Employee join-org, InviteCode (rank 6) | Checked again: plain `setStatus` on create and on update, the same role gate on both, and the only status-related events in the codebase are `ProjectApprovedEvent` (closed by #544) and `SiteTransferCreatedEvent` (which fires on create whatever the status). Nothing sits behind any of their terminal values. |

## Deploy ordering

None of this needs a coordinated deploy, for the same reason #544 gave. A refused value fails loudly with a message naming the endpoint to use instead, rather than being silently written or silently dropped. Beyond that:

- Stock adjustments have no client at all: the module appears nowhere in `echno-core` or `echno-web`, so nothing sends a status to refuse. Staging holds one row, a draft.
- The site-transfer create page already hardcodes `SiteTransferStatus.pending`, so the tightened endpoint refuses nothing the client sends. tornotron/echno-web#320 needs no follow-up for this, and neither does anything else in the app; the status control on the transfer form is on the detail page, which goes through the status endpoint.
- Deleting `EmployeeCreationDto.status` cannot break a client that still sends it: the application's mapper ignores an undeclared property, pinned by a test.

## Failing-test table

Each new test run against a reverted baseline (the guards removed and the status copied from the body again, `EmployeeCreationDto.status` restored), then against the fix.

|---|---|---|
| `SiteTransferCreateStatusTest.create_refusesATransferIssuedAlreadyCompleted` | FAILED | PASSED |
| `SiteTransferCreateStatusTest.create_refusesATransferIssuedAlreadyPartiallyTransferred` | FAILED | PASSED |
| `SiteTransferCreateStatusTest.create_refusesEveryStartingStateButPending` | FAILED | PASSED |
| `SiteTransferCreateStatusTest.create_withNoStatusInThePayload_startsPending` | FAILED | PASSED |
| `StockAdjustmentCreateStatusTest.create_refusesADocumentRaisedAlreadyApproved` | FAILED | PASSED |
| `StockAdjustmentCreateStatusTest.create_refusesADocumentRaisedInThePostedState` | FAILED | PASSED |
| `StockAdjustmentCreateStatusTest.create_withNoStatusInTheBody_raisesADraft` | FAILED | PASSED |
| `StockAdjustmentCreateStatusTest.create_withDraftInTheBody_isAcceptedWhateverItsCase` | FAILED | PASSED |
| `StockAdjustmentCreateStatusTest.update_refusesToMoveADocumentIntoThePostedState` | FAILED | PASSED |
| `EmployeeCreationDtoTest.aPayloadThatNamesNoStatusIsValid` | FAILED | PASSED |

Two further new tests are negative controls and pass either way by design: `SiteTransferCreateStatusTest.create_withPendingInThePayload_isAccepted` (the legitimate value still goes through and still publishes the stock event) and `EmployeeCreationDtoTest.aPayloadThatStillSendsAStatusIsStillAccepted` (an old client is not broken).

Full suite on `lab-node-116-f`: **BUILD SUCCESSFUL**, 1242 tests, 0 failures, on `ef15dad`. Plain JUnit and Mockito throughout, no new Spring context, so no change to the test-heap position. No Liquibase changeset: nothing about the stored data changes.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/employee/dto/EmployeeCreationDto.java
+++ b/src/main/java/org/tornotron/echno_backend/employee/dto/EmployeeCreationDto.java
@@ -3,8 +3,6 @@ package org.tornotron.echno_backend.employee.dto;
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.OptBoolean;
 import io.swagger.v3.oas.annotations.media.Schema;
-import jakarta.persistence.EnumType;
-import jakarta.persistence.Enumerated;
 import jakarta.validation.constraints.*;
 import lombok.Data;
 
@@ -40,11 +38,10 @@ public class EmployeeCreationDto {
     @Schema(description = "Id of the structured shift timing to assign to the employee.", example = "5")
     private Long shiftTimingId;
 
-    @Schema(description = "Employment status.", example = "active")
-    @NotBlank(message = "status is required")
-    @Size(min = 3, max = 50, message = "status must be between 3 and 50 characters")
-    @Enumerated(EnumType.STRING)
-    private String status;
+    // No status field. It was declared here and required, then discarded by the mapper that
+    // builds the employee, so a caller was made to send a value nothing ever read and a caller
+    // that left it out was refused for it. An employee's status is set through the employee
+    // update endpoint, which is the one place that writes it.
 
     @Schema(description = "Full name of the employee.", example = "Ravi Kumar")
     @NotBlank(message = "employeeName is required")
--- a/src/main/java/org/tornotron/echno_backend/siteTransfer/SiteTransferService.java
+++ b/src/main/java/org/tornotron/echno_backend/siteTransfer/SiteTransferService.java
@@ -64,6 +64,13 @@ import java.util.stream.Collectors;
 @Service
 public class SiteTransferService {
 
+    /**
+     * The state a transfer starts in, and the only one create accepts. It is what the web
+     * client's transfer form already sends, and what {@code echno-core} documents as the
+     * typical initial value.
+     */
+    private static final SiteTransferStatus CREATION_STATUS = SiteTransferStatus.PENDING;
+
     private final SiteTransferRepository siteTransferRepository;
     private final SiteTransferItemRepository siteTransferItemRepository;
     private final UserRepository userRepository;
@@ -109,6 +116,9 @@ public class SiteTransferService {
     /**
      * Creates a site transfer with its items after checking sending-side stock.
      *
+     * <p>The transfer starts {@link SiteTransferStatus#PENDING}, whether the payload says so or
+     * says nothing, and any other starting value is refused: see {@link #requireCreatableStatus}.
+     *
      * <p>Resolves the sending person, both projects and both optional storage locations,
      * refuses a pair of sides that resolve to the same balance row, then totals the requested
      * quantities per material and validates them against the sending side. The transfer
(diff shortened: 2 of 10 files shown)


abhijithanandan:
@coderabbitai review

PR #570, opened 2026-08-29
Close the API docs by default instead of at one edge

### LINKED ISSUE #569: Swagger and the OpenAPI document are permitAll by default and only the edge hides them
Found while auditing the `permitAll` list for #547. Not a tenant-isolation problem: springdoc reads Spring's own request-mapping metadata and `OpenApiConfig` builds a static document, so nothing here reaches the database. It is a default that leans the wrong way and depends on something outside this repo to be safe.

## What is true today

`application.yml` ships `springdoc.swagger-ui.public-access: true`, and `SecurityConfig` reads it as `${springdoc.swagger-ui.public-access:false}` to decide whether to `permitAll` `/swagger-ui/**`, `/swagger-ui.html` and `/v3/api-docs/**`. There is no profile block and no environment override anywhere in the repo, so every environment built from this file serves the whole endpoint surface, every DTO schema and every path, to an unauthenticated caller on the container port.

What stops that being public is `echno-deployment`, `ansible/roles/edge/templates/vhost.conf.j2`, which returns 404 for `actuator|swagger-ui|swagger-resources|webjars|v3/api-docs` on the backend vhost. That is the right call at the edge and it is well commented there. It is also the only thing standing between the default and a public API catalogue, and it lives in a different repository.

Note the code default is already `false`: the fallback in the `@Value` is the safe one. It is the shipped `application.yml` that turns it on.

## Why it is worth a ticket rather than a shrug

- The safe state depends on an edge rule, so anything that reaches the backend without passing that vhost is exposed. Worth confirming for the k8s ingress on `ui-k8s.echno.in`, which does not use the compose edge template.
- A developer running the app directly, or any future deployment shape, inherits the permissive default silently.
- Defence in depth is cheap here: flip the shipped value to `false` and set it `true` only where the docs are wanted.

## Suggested shape

Set `springdoc.swagger-ui.public-access: false` in `application.yml` and turn it on per environment (an env var on the local/dev compose file). Keep the edge 404 either way. While there, `management.endpoint.health.show-details: always` deserves the same look: it is unauthenticated on `/actuator` for the same reason and hidden by the same edge rule.

## Related

- #547 (the audit), #527 (the tenant work that prompted it)

abhijithanandan:
Closes #569.

## What was wrong

`application.yml` shipped `springdoc.swagger-ui.public-access: true`. `SecurityConfig` reads it as `${springdoc.swagger-ui.public-access:false}`, so the code default was already the safe one; the shipped value is what turned the docs on everywhere.

What kept them private was a rule in `echno-deployment`, `ansible/roles/edge/templates/vhost.conf.j2`, which returns 404 for `actuator|swagger-ui|swagger-resources|webjars|v3/api-docs` on the backend vhost. That rule is correct and stays. It covers one ingress. The k8s ingress carries no equivalent, so the exposure the issue predicted is not latent, it is live.

Probed unauthenticated, from outside the network, before this change:

| URL | `backend.echno.in` (compose) | `backend-k8s.echno.in` (k8s) |
|---|---|---|
| `/v3/api-docs` | 404 | **200, 1,466,410 bytes** |
| `/swagger-ui/index.html` | 404 | **200** |
| `/actuator/health` | 404 | **200** |

The complete API specification is a public 1.4 MB download today. Nothing here reaches the database (springdoc reads Spring's own request-mapping metadata), so this is not a data exposure. It is the whole endpoint map, which is the reconnaissance step for finding the endpoint whose authorization is weakest.

## What this changes

- `springdoc.swagger-ui.public-access: ${SWAGGER_PUBLIC_ACCESS:false}`. Closed by default, switchable per environment.
- `management.endpoint.health.show-details: ${MANAGEMENT_ENDPOINT_HEALTH_SHOW_DETAILS:when_authorized}`, same shape. See the correction below: this one is not currently broken anywhere.
- `ShippedConfigurationDefaultTest` evaluates both values the way an environment setting nothing would, and fails the build if either goes back to being permissive. Verified both directions: it passes on the new values and fails on the old ones. It parses the file rather than booting a context, so it costs nothing in the 1 GB test JVM.
- README says the docs are closed by default and how to open them.

## Correction to the issue: health details do not reproduce

The issue also flagged `management.endpoint.health.show-details: always`. It is not in effect on either host. `backend-k8s.echno.in/actuator/health` returns

```json
{"status":"UP","groups":["liveness","readiness"]}
```

with no component detail, because both deployments already set `MANAGEMENT_ENDPOINT_HEALTH_SHOW_DETAILS=when_authorized` by environment variable (`ansible/roles/apps/templates/backend.env.j2` and `k8s/scripts/create-app-secrets.sh`). So nothing is being fixed there. The change is alignment: the shipped default now says what both environments already say, which makes it a literal no-op in deployment and closes the same gap for a developer run and for any future deployment shape. If you would rather this PR touched only springdoc, say so and I will drop that hunk.

## Does this take away something the team uses

On compose staging the docs have always been 404 at the edge, so nothing changes there. On k8s staging they were reachable and will now answer 401. I have no evidence anyone was reading them there (Apidog is the team's collection), but it is a real capability being withdrawn, so it should be a decision rather than a side effect.

To turn them back on for an environment, set `SWAGGER_PUBLIC_ACCESS=true`:

- k8s: one more entry in the backend `env` list in `k8s/helm/echno/templates/backend.yaml`, or in `echno-backend-env`.
- compose: one line in `ansible/roles/apps/templates/backend.env.j2`. Note the edge rule would still 404 them, so the vhost rule would have to be relaxed too.
- locally: `SWAGGER_PUBLIC_ACCESS=true ./gradlew bootRun`.

## Two decisions, recorded rather than assumed

**The k8s ingress does not get the equivalent block.** The compose parallel is tempting, but ingress-nginx 4.11 ships `allow-snippet-annotations: false` (default since controller 1.9), so `server-snippet` is the only way to express `return 404` there and it would mean flipping that flag, plus the annotations risk level, on the shared controller. That lets anything able to create an Ingress in any namespace inject nginx configuration: a cluster-wide relaxation bought to close a host-specific gap that this PR now closes at the application, on every ingress at once, present and future. The application default is the durable control here; a third place to remember is what produced this issue in the first place.

**`/actuator/**` is a separate problem and is not in this PR.** It is unconditionally `permitAll` in `SecurityConfig` and cannot simply be authenticated: the kubelet probes and the compose healthcheck reach it with no credential. On k8s that leaves `/actuator` (the link listing), `/actuator/info` and `/actuator/prometheus` (470 lines of metrics, including per-URI request timings) publicly readable. The disclosure is small but real, and neither lever used here reaches it: the application cannot close it without breaking the probes, and the ingress cannot without the snippet relaxation above. The clean fix is `management.server.port`, moving actuator to its own port that the ingress does not route to while kubelet and the PodMonitor keep scraping the pod directly. That touches probes, the PodMonitor, the compose healthcheck and the blackbox check, so it belongs in its own change. Filed as a follow-up.

## Verification

Before and after probes on both hosts are in the issue thread once this reaches staging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,14 @@ Once the service is running, the live, generated API reference is available at `
 (OpenAPI JSON at `/v3/api-docs`). Controllers and DTOs carry OpenAPI annotations, so the Swagger UI
 groups endpoints by module and documents request and response fields.
 
+Those paths are closed to unauthenticated callers by default, because the OpenAPI document is the
+whole endpoint surface in one file. Set `SWAGGER_PUBLIC_ACCESS=true` to open them for an environment
+where the docs are wanted, such as your own run:
+
+```bash
+SWAGGER_PUBLIC_ACCESS=true ./gradlew bootRun
+```
+
 ## Getting Started
 
 ### Prerequisites
--- a/src/main/java/org/tornotron/echno_backend/common/configuration/SecurityConfig.java
+++ b/src/main/java/org/tornotron/echno_backend/common/configuration/SecurityConfig.java
@@ -37,6 +37,13 @@ public class SecurityConfig {
     @Value("${keycloak.frontend.web-origin}")
     private List<String> allowedOrigins;
 
+    /**
+     * Whether the API docs are served to an unauthenticated caller. Closed unless an
+     * environment opts in (SWAGGER_PUBLIC_ACCESS=true), because the OpenAPI document is
+     * the entire endpoint surface in one download. Both the fallback here and the value
+     * shipped in application.yml are false, so an ingress that reaches this application
+     * without passing an edge rule still gets 401 on the docs paths. See issue #569.
+     */
     @Value("${springdoc.swagger-ui.public-access:false}")
     private boolean swaggerPublicAccess;
 
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -21,7 +21,15 @@ management:
     prometheus:
       access: unrestricted
     health:
-      show-details: always
+      # /actuator is permitAll (kubelet probes and the compose healthcheck reach it
+      # without a credential), so whatever this value is, it is served to whoever can
+      # reach the port. "always" would put every component's detail there: database
+      # url, disk paths, broker addresses. "when_authorized" keeps the aggregate
+      # status the probes need and withholds the breakdown from an anonymous caller.
+      # Both deployed environments already set this to when_authorized by env var; the
+      # shipped default now matches, so a developer run and any future deployment
+      # shape inherit it too.
+      show-details: ${MANAGEMENT_ENDPOINT_HEALTH_SHOW_DETAILS:when_authorized}
   endpoints:
     web:
       exposure:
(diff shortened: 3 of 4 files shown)


abhijithanandan:
@coderabbitai review

Issue #563, opened 2026-08-29
Correct the -30 unlocated balance on staging, now that its cause is known and an adjustment path exists


abhijithanandan:
Staging holds a negative stock balance that should not exist. It was reported as item 4 of #529 with no cause identified; #556 found the cause.

## The row

`current_stock` id 15: material 8, project 7, no storage location, quantity **-30**.

## How it got there

Movement row 16, on the consumption path:

```
 id | type | material | project | location | opening | changed | closing | ref
 16 | USE  |        8 |       7 |     NULL |     400 |     -30 |     370 | MC-1
```

Material 8 at project 7 held 400 at storage location 2 and nothing outside a location. The entry claims 400 → 370, but the row it actually debited was the project's unlocated row, 0 → -30. The opening figure came from the project total while the write went to the unlocated row.

That mismatch is fixed: #529 corrected the validation side, and #556 corrected the movement-recording side in the three remaining places (`TRANSFER_OUT`, `TRANSFER_IN`, `handleGrnCreated`). So this cannot recur. What remains is the row itself.

## What to do

Two separate questions, and they should be answered in order.

**1. What is the true figure?** Material 8 at project 7 apparently has 400 at location 2 and -30 unlocated, when the real consumption of 30 was almost certainly meant to come out of location 2. So the honest end state is likely 370 at location 2 and 0 unlocated. **Confirm that against the source document (MC-1) before adjusting anything**, rather than assuming.

**2. Correct it through the product, not the database.** #529 added the stock adjustment path precisely so a wrong balance has a route back, and #559 gave it segregation of duties. Using it here is also a real exercise of that path against a genuine case, which is worth more than the correction itself. A direct SQL update would fix the number and leave the ledger unable to explain it.

Note the adjustment will carry a reason, and the reason should name this issue so the history stays readable.

## Scope

Staging only. Whether any client environment carries an equivalent row is unknown and worth a check when one exists, since the defect predates the fix.

## Related

#529 (where it was reported and the validation fix), #556 (the cause and the recording fix), #559 (self-approval rules now on the adjustment path)

abhijithanandan:
Re-verified read-only against the compose staging on `.116`, 30 Aug 2026. The row is unchanged, the true figure is confirmed against MC-1, and the correction is blocked. Not by the thing I expected.

## Still in the state described

```
 id | mat | proj | loc  |  qty | stock_value
 15 |   8 |    7 | NULL |  -30 |        0.00
[...]

PR #572, opened 2026-08-29
Let a stock adjustment correct a balance on another project's location

abhijithanandan:
Closes the blocker on #563.

## The problem

`StockAdjustmentService.approve` scopes every line through
`StorageLocationScope.requireUsableFromProject`, which refuses a line naming a storage
location whose `project_id` differs from the adjustment's project. That rule is right for a
**new** movement, which is what it was written for. An adjustment correcting an existing
balance is the one case where the wrong pairing is the thing being fixed, so refusing it
removes the only tool.

Twelve of the twenty `current_stock` rows on staging sit at exactly that pairing, listed in
#563. #529 built the adjustment path as the route back for balances no goods receipt can
explain, and the scope rule closed it against the legacy rows that most need it. Both landed
in the same commit, `5973899`, so it has never worked for these.

## The narrowed rule

On the adjustment path, a location is accepted when a balance row already exists at that
`(material, project, location)`. That is the precise condition and it is deliberately narrow:
it permits correcting a pairing that is already there and still refuses inventing a new one.
It is not "any location in the org", and the check is not disabled.

## How the relaxed path is made explicit

A second entry point on `StorageLocationScope`, not a flag on the existing one:

```java
public static void requireUsableForBalanceCorrection(StorageLocation location, Long projectId,
                                                     BooleanSupplier balanceRowExists)
```

- The strict `requireUsableFromProject` stays untouched and stays the default. Consumption,
  goods receipt and site transfer still call it.
- Relaxing has to be asked for by name at the call site, so no future caller can opt into it
  by passing a boolean into the method it already uses.
- The third argument is a fact about the data, not a mode switch, and it is a `BooleanSupplier`
  consulted only when the strict rule would refuse, so the common path pays nothing for it.
  `approve` already reads the balance, so it passes `existingBalance::isPresent` and there is
  no extra query.
- The refusal message says why: the location holds no balance for this material and project,
  so there is no figure here to correct, and an adjustment cannot create one.

## Tests

Both directions, each proved to fail without the fix.

| Test | What it pins | Without the fix |
|------|--------------|-----------------|
| `StockAdjustmentApprovalTest.aBalanceAlreadySittingOnAnotherProjectsLocationCanBeCorrected` | The #563 shape: 400 at a location owned by another project, counted at 370, posts a -30 `ADJUST` against that location | FAILS: `InvalidRequestException: Storage location with ID 14 belongs to project with ID 9 and cannot be used from project with ID 7` |
| `StockAdjustmentApprovalTest.aLineOnAnotherProjectsLocationHoldingNoBalanceIsRefused` | An adjustment still cannot invent a cross-project pairing; nothing is posted | FAILS: refused by the strict rule with the wrong message, so it is not exercising the new path |
| `StorageLocationScopeTest.aBalanceAlreadySittingOnAnotherProjectsLocationCanBeCorrected` | The rule itself, in one place | FAILS to compile: method does not exist |
| `StorageLocationScopeTest.anAdjustmentCannotInventABalanceOnAnotherProjectsLocation` | The refusal and its message | FAILS to compile: method does not exist |
| `StorageLocationScopeTest.aLocationTheStrictRuleAlreadyAllowsIsNotLookedUpAtAll` | The lookup is not run when the strict rule already passes | FAILS to compile: method does not exist |
| `MaterialConsumptionStockScopeTest.aStorageLocationBelongingToAnotherProjectIsRefusedEvenWhenItAlreadyHoldsABalance` | Leak guard: consumption is still strict with a balance row present | Passes before and after, by design. It fails only if the relaxation is later moved onto the shared rule |
| `SiteTransferStockScopeTest.aSendingLocationBelongingToAnotherProjectIsRefusedEvenWhenItAlreadyHoldsABalance` | Leak guard: transfer is still strict with a balance row present | Same |

The two leak guards are regression guards rather than red-then-green tests, and they are
stated that way on purpose: they hold the strict rule down on the paths that record a new
movement, and both would have passed on the old code because the old code was strict
everywhere.

Full `./gradlew test` green on the lab build box.

## Note for the web client

`StorageLocationScope`'s doc points at `features/material-consumptions/storage-location-scope.ts`,
which applies the strict rule to the location dropdown. That is not changed here and does not
need to be for correctness: it narrows what can be picked, not what can be posted. It does mean
the adjustment screen will not offer such a location until it is widened. Worth pairing with the
separate gap #563 records, that `echno-web` and `@tornotron/echno-core` carry no call to
`POST /stock-adjustments/web/{id}/approve` at all, so a draft can be raised in the UI and never
posted.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/stockAdjustment/StockAdjustmentService.java
+++ b/src/main/java/org/tornotron/echno_backend/stockAdjustment/StockAdjustmentService.java
@@ -32,6 +32,7 @@ import org.tornotron.echno_backend.user.UserContextService;
 import java.math.BigDecimal;
 import java.time.LocalDateTime;
 import java.util.List;
+import java.util.Optional;
 
 /**
  * Stock-adjustment documents: draft, browse, edit, and post to the stock ledger.
@@ -199,6 +200,13 @@ public class StockAdjustmentService {
      * document, unless they hold the break-glass role, in which case the posting is allowed and
      * the ledger entries say they were self-approved. See {@link SelfApprovalPolicy}.
      *
+     * <p>The location on a line is scoped through
+     * {@link StorageLocationScope#requireUsableForBalanceCorrection}, not the strict rule the
+     * other stock paths use. A location belonging to another project is still refused, unless a
+     * balance row already sits at that (material, project, location), which is the one pairing an
+     * adjustment exists to correct: the strict rule is written for a new movement, and applying it
+     * here would leave a wrongly located balance with no route back at all.
+     *
      * <p>Every posted line writes an {@code ADJUST} {@link InventoryTransaction} whose
      * remarks carry the reason, so the balance stays explainable, and only then moves the
      * balance. Both happen in this transaction: an approval that recorded no movement is the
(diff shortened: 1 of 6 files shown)


abhijithanandan:
@coderabbitai review

PR #574, opened 2026-08-29
Give compliance rules a change timestamp, and a sweep that reads it

### LINKED ISSUE #553: No nightly sweep for already-approved projects, and ComplianceRule has no change timestamp to drive one
## What is missing

Compliance inspections are generated when a project is approved, and only then. A project approved before its applicable rules existed never gets them, and nothing goes back for it. There is no sweep.

Came out of #542. This one needs a schema change before it can be built, so it is being filed rather than picked up.

## What I verified (29 Aug 2026)

- Generation is triggered from approval, through `ComplianceGenerationListener` into `ComplianceGenerationService`. On branch `feat/482-async-compliance-generation` the only `@Scheduled` in the compliance package is `ComplianceGenerationJobDispatcher` at line 120, which drains the job queue every `compliance.job.poll-interval-millis` (default 2000). It picks up queued jobs; it does not look for projects that should have one.
- `ComplianceGenerationService.gatherInputs` aborts when `IndianStateResolver.forProject(...)` returns null, or when `findByStateIgnoreCaseAndProjectTypeAndActiveTrue(state, projectType)` finds no rules. In both cases the project stays approved with no compliances and nothing records that it was tried.
- **`ComplianceRule` carries no timestamp of any kind.** Its fields are `id`, `state`, `projectType`, `phase`, `code`, `name`, `description`, `defaultRiskLevel`, `resolutionOptions`, `authority`, `active`. No `createdAt`, no `updatedAt`, and it does not extend an auditable base. So there is no way today to ask "which rules changed since the last sweep".
- On staging right now there are 21 active rules and exactly one approved project (id 27, "Demo Project"), which has zero compliance inspections. See tornotron/echno-backend#549.

## Why it matters

This is the shape of the problem the moment Anand's curated rule catalogue lands. Today the catalogue is a small seed and every project of consequence is still `upcoming`, so nothing is visibly wrong. When the real catalogue is loaded, every project already approved by then is permanently missing the compliances it should have, with no signal that anything is absent. Someone finds out by noticing an empty list, which is the worst way to find out.

A sweep also covers the two abort paths above, which are silent by design: a project approved while its state was unset gets nothing, and if the state is filled in later nothing retries.

## What a fix would involve

Order matters here.

1. **First, the schema.** Add `created_at` and `updated_at` to `compliance_rules` (Liquibase changeset, alongside the existing `037-create-and-seed-compliance-rules.xml`), and set them on write. Without a change timestamp a sweep can only be "reconsider every approved project every night", which re-runs the model against unchanged inputs and costs real money.
2. **Then the sweep.** A nightly `@Scheduled` that finds approved projects whose compliance set predates the newest applicable rule change, and enqueues a generation job through the existing `ComplianceGenerationJobDispatcher` rather than calling the service directly, so batching, retry and progress reporting all still apply.
3. **Guard the cost.** Cap how many projects one pass enqueues, and make sure the existing dedupe (the unique constraint behind `060-compliance-inspection-rule-ref-unique.xml`) means a re-run adds only the genuinely new compliances rather than duplicating what is there.

Open question for whoever picks this up: should the sweep also retry the projects that aborted for an unresolvable state, or should those surface as a warning on the project instead? Retrying them nightly forever is not obviously right.

abhijithanandan:
Closes #553.

Compliance is generated when a project is approved and only then. A project approved before the rules that cover it existed never gets them, and nothing goes back for it. The moment Anand's curated catalogue lands, every project approved before it is permanently missing compliances that nobody knows are missing, and someone finds out by noticing an empty list.

Two things here: the timestamp, which had to come first, and the sweep that reads it.

## The timestamp, and why it is three columns rather than two

The issue asked for `created_at` and `updated_at`. Building it that way would have produced a sweep that regenerates the world, so this adds a third.

`created_at` and `updated_at` are ordinary audit columns written by Hibernate. Nothing schedules off them. `effective_from` is the curated one and the only input to the sweep.

They are not the same fact, and the difference is the whole design. `updated_at` moves on **any** write: a corrected typo in `description`, a widened `authority` string, and every row of a bulk re-seed whose content did not change at all. A sweep keyed on it would read a comma fix across the catalogue as the whole catalogue changing, and spend a model call per approved project per tenant to re-derive answers that are identical. That is expensive and it is also simply wrong, because nothing about the answer changed.

So `effective_from` is deliberately inert. It is set once, to the insert time, by a `@PrePersist` that fires only when the caller has not supplied one, and the application never touches it again. Editing a rule does not move it. Re-dating it forward is a curator's deliberate statement that the rule is materially different and projects should be assessed against it again, and that act is the only thing that makes an already-assessed project stale.

Two cases that are easy to get wrong, stated in the entity javadoc so the next person does not have to rediscover them. Flipping `active` back to true does not by itself re-date a rule, so a rule switched off and on again reaches only projects that were never assessed; move `effective_from` too if the intent is that everyone re-assesses. And a rule loaded with a back-dated `effective_from` reaches nobody already assessed, which is right for recording history and wrong for a rule that is new to us.

### The backfill is load bearing

Changelog `069` dates the existing seeded rules to when the seed **actually ran**, read out of Liquibase's own `databasechangelog` (2026-08-21 on IITM staging), rather than to `now()`.

This is the difference between the first pass doing nothing and the first pass queueing a job for every approved project in every tenant. The 21 seeded rules have been in force since `037` applied. Dating them to `now()` would make every one newer than every assessment ever done, so every approved project would read as stale, spend a model call and create nothing, because those projects were already assessed against exactly those rules. The coalesce chain degrades to the changeset id, then the filename, then `current_timestamp`; a row that lands on the fallback costs one wasted call, not a wrong answer.

## The sweep

`ComplianceRuleSweep` compares, per approved project, the newest `effective_from` among the active rules for its jurisdiction against the last time a run for that project actually finished. Never assessed is stale by definition, which is the backlog the issue is about.

**It queues, it does not generate.** Every stale project goes through `ComplianceGenerationJobService.submit`, the same entry point approval and the manual endpoint use. Calling generation inline from a scheduler thread would put a model call per project outside any lease, with no record it was attempted, no progress for anyone watching, and nothing to recover it when the replica restarts halfway through, which is the whole of what #542 built the queue to fix. Going through the queue also means the one-active-job-per-project partial index does the obvious right thing: a project with a run already in flight is handed back that run instead of getting a second one, and that outcome does not consume one of the pass's slots.

### What happens to a project already carrying compliances from an older catalogue

Nothing is duplicated and nothing is contradicted. Generation is already idempotent per `(projectId, complianceRuleRef)`: `ComplianceGenerationService.persist` skips a rule that already has an inspection on the project, and the partial unique index on `(organization_id, project_id, compliance_rule_ref)` from #510 / changelog `060` is what makes that an invariant rather than an unlocked read. A swept project therefore gains only the compliances it was missing, and existing rows are untouched, including any a user has since scheduled, added check items to, or raised an NCR against.

One consequence worth being explicit about, because it is a real limit rather than an oversight: generation only ever **creates**. A rule whose text was corrected produces no change on a project that already has that rule's inspection, even if the sweep re-runs it. That is another reason `effective_from` is the right trigger and `updated_at` is not, since re-sweeping for a text correction would cost a model call and change nothing at all.

### Pacing

The thundering herd is prevented by the queue rather than by anything in the sweep. Enqueueing is an insert. `compliance.job.max-in-flight` (default 2 per replica) decides how many runs ever reach the inference endpoint at once, so a pass that queues 25 jobs produces a queue that drains two at a time, not 25 simultaneous calls against a rate-limited endpoint. This is exactly why the sweep must not bypass the queue.

On top of that the pass bounds both what it reads and what it spends:

- `scan-limit` (500) bounds the cross-tenant scan. An unbounded scan of every approved project is fine at today's 29 and is not fine later.
- `max-projects-per-run` (25) bounds the spend. At the measured ~1.7 s and ~57 completion tokens per rule, 25 projects over a 21-rule catalogue is roughly nine minutes of wall clock at two in flight.

Candidates come back oldest-assessed first with nulls first, so a capped pass puts the real backlog at the front, makes deterministic progress, and the next pass continues behind it.

## What is left configurable, and off by default

`compliance.sweep.enabled` is **false**, and the bean does not exist unless it is turned on. Saying that plainly: the mechanism is settled, the policy is not, and three of the open questions are product decisions this code cannot make.

- Which catalogue changes justify re-assessment. This PR's answer is "whatever a curator re-dates", which is a mechanism, not a policy.
- How often a pass should run. `cron` defaults to 02:00 `UTC`.
- Whether a tenant should be able to decline it, since the spend is real and lands on whoever owns the inference key. Not built; there is no per-tenant switch here.

The issue's own open question I did answer, and it is worth flagging for disagreement. A project whose state still cannot be resolved is skipped quietly every pass rather than retried against the model, because there is nothing to retry: the precondition fails before any call is made. Retrying it nightly for ever costs nothing but achieves nothing. Skipping it silently for ever would be its own version of this bug, so the pass counts them and logs the count. Surfacing that on the project itself, where somebody would actually see it, is a UI change and is left open.

## Tests

Every test below was proven to fail by mutating the single behaviour it guards, on `lab-node-116-f`, then restoring.

| Test | Mutation applied | Result without it |
|---|---|---|
| `leavesAProjectAssessedAfterTheNewestRuleChangeAlone` | up-to-date check removed | FAILED (`NeverWantedButInvoked`) |
| `treatsAnAssessmentAtTheSameInstantAsUpToDate` | up-to-date check removed | FAILED |
| `queuesAProjectThatWasNeverAssessed` | null last-assessed treated as up to date | FAILED |
| `stopsAtThePerRunCap` | per-run cap removed | FAILED |
| `doesNotCountAProjectAlreadyRunningAgainstTheCap` | already-running counted against the cap | FAILED |
| `skipsAProjectWhoseStateCannotBeResolved` | unresolvable-state guard removed | FAILED |
| `carriesOnAfterOneProjectIsRefused` | expected-exception catch removed | FAILED |
| `matchesTheJurisdictionCaseInsensitively` | case-sensitive jurisdiction key | FAILED |
| `doesNothingWhenNoRulesAreActive` | empty-catalogue short circuit removed | FAILED |
| `isOffByDefault` | `enabled` defaulted to true | FAILED |
| `aNewRuleBecomesEffectiveWhenItIsCreated` | `@PrePersist` body removed | FAILED |
| `anExplicitEffectiveFromIsNotOverwritten` | `@PrePersist` overwrites unconditionally | FAILED |
| `reportsWhenEachJurisdictionLastChanged` (IT) | `069` backfill leaves `effective_from` null | FAILED |
| `findsAnApprovedProjectThatWasNeverAssessed` (IT) | new query, no prior behaviour | n/a |
| `doesNotCountAFailedRunAsAnAssessment` (IT) | `FAILED` added to the assessed statuses | FAILED |
| `countsARunThatFoundNothingApplicableAsAnAssessment` (IT) | `NOTHING_TO_REPORT` dropped from the assessed statuses | FAILED |
| `countsASucceededRunAsAnAssessment` (IT) | new query, no prior behaviour | n/a |

The integration tests went into the existing `ComplianceGenerationServiceIT` rather than a new class, deliberately: they need the real CockroachDB and the migration-seeded catalogue, and reusing that class's context means no additional Spring context is cached in a JVM that already runs at `maxHeapSize=1024m` with no `forkEvery`.

Both new queries are exercised against a real database, which is the only place they can be. `findSweepCandidates` is native SQL over a table its repository does not own, with a left join whose null is load bearing, and `findNewestEffectiveFromByJurisdiction` is the only assertion that the backfill actually dated the seeded rows. Full `./gradlew build` green on `.116`.

## Changelog

`069`, checked against `origin/development` (max `068`) and against every open PR immediately before pushing; no other PR touches a v4.0 changelog.

## Not done here

No `effective_from` editing surface. Re-dating a rule is a database operation today, which is fine while the catalogue is curated by hand, and is the obvious next thing if the sweep is turned on and used in anger.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/compliance/domain/ComplianceRule.java
+++ b/src/main/java/org/tornotron/echno_backend/compliance/domain/ComplianceRule.java
@@ -4,9 +4,12 @@ import jakarta.persistence.*;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.UpdateTimestamp;
 import org.tornotron.echno_backend.compliance.CompliancePhase;
 import org.tornotron.echno_backend.inspection.ComplianceRiskLevel;
 
+import java.time.LocalDateTime;
 import java.util.UUID;
 
 /**
@@ -68,4 +71,55 @@ public class ComplianceRule {
 
     @Column(name = "active", nullable = false)
     private boolean active = true;
+
+    /** When this row first appeared. Audit only; nothing schedules off it. */
+    @CreationTimestamp
+    @Column(name = "created_at", nullable = false, updatable = false)
+    private LocalDateTime createdAt;
+
+    /** When this row was last written, for any reason at all. Audit only. */
+    @UpdateTimestamp
+    @Column(name = "updated_at", nullable = false)
+    private LocalDateTime updatedAt;
+
+    /**
+     * When the rule came into force, as the catalogue's curator states it. This is the only
+     * timestamp the nightly sweep reads, and the distinction from {@link #updatedAt} is the
+     * whole reason it exists.
+     *
+     * <p>{@code updatedAt} moves on every write, including a corrected typo in
+     * {@link #description} and including a bulk re-seed that rewrites rows whose content did
+     * not change. A sweep keyed on it would treat a comma fix across the catalogue as the
+     * whole catalogue changing, and re-assess every approved project in every tenant against
+     * rules whose meaning is identical. That is expensive and it is also wrong: nothing about
+     * the answer changed.
+     *
+     * <p>So this one is deliberately inert. It is set once, to the insert time, by
+     * {@link #defaultEffectiveFrom()} when the caller has not set it, and the application never
+     * touches it again. Re-dating it forward is a curator's decision that says "this rule is
+     * materially different, assess projects against it again", and it is the only thing that
+     * makes an already-assessed project stale.
+     *
+     * <p>Two cases worth stating because they are easy to get wrong. Reactivating a rule by
+     * flipping {@link #active} back to true does not by itself re-date it, so a rule switched
+     * off and on again reaches only projects that were never assessed; move this forward as
+     * well if the intent is that everyone re-assesses. And a rule loaded with a back-dated
+     * {@code effectiveFrom} reaches nobody who has already been assessed, which is the right
+     * behaviour for recording history and the wrong one for a rule that is new to us.
+     */
+    @Column(name = "effective_from", nullable = false)
+    private LocalDateTime effectiveFrom;
+
+    /**
+     * Gives a new rule an {@code effectiveFrom} of now unless one was supplied. A rule the
+     * curator says nothing about is in force from when we learned of it, which is the only
+     * defensible default: dating it earlier would hide it from every project already assessed,
+     * and that is the exact failure this column exists to prevent.
+     */
+    @PrePersist
+    void defaultEffectiveFrom() {
+        if (effectiveFrom == null) {
+            effectiveFrom = LocalDateTime.now();
+        }
+    }
 }
--- a/src/main/java/org/tornotron/echno_backend/compliance/job/ComplianceGenerationJobRepository.java
+++ b/src/main/java/org/tornotron/echno_backend/compliance/job/ComplianceGenerationJobRepository.java
@@ -179,6 +179,65 @@ public interface ComplianceGenerationJobRepository extends JpaRepository<Complia
             + "FROM compliance_generation_jobs WHERE id = :id", nativeQuery = true)
     Optional<DispatchRow> findDispatchRow(@Param("id") UUID id);
 
+    /**
+     * Approved projects across every tenant, each with the last time compliance was actually
+     * assessed for it. The nightly sweep's one scan.
+     *
+     * <p>Cross-tenant and scalars only, for the same reason the dispatcher queries above are:
+     * the caller has no tenant, because its whole job is to work out which tenants have
+     * something to do. It establishes one per project before anything is enqueued.
+     *
+     * <p>"Assessed" means a run that reached the model and came back, which is {@code SUCCEEDED}
+     * or {@code NOTHING_TO_REPORT}. The second of those matters: a run that assessed every rule
+     * and found none applicable did the work, and treating it as never-assessed would put the
+     * project back in the queue every night for ever. {@code FAILED} deliberately does not
+     * count, because a failed run assessed nothing, and {@code QUEUED} and {@code RUNNING} do
+     * not count because they have not finished; a project with one in flight is filtered out
+     * later by the job table's own one-active-job-per-project index, not here.
+     *
+     * <p>A null {@code lastAssessedAt} is a project that has never had a completed run at all,
+     * which is exactly the backlog this exists for: everything approved before generation was
+     * wired up, and everything whose approval-time run failed silently. Ordering nulls first
+     * puts that backlog at the front of a capped pass.
+     *
+     * <p>The project's jurisdiction is returned raw rather than matched here. Resolving a state
+     * from a free-text address is Java, so the comparison against the rule catalogue happens in
+     * the caller; this query's job is only to bound what the caller looks at.
+     */
+    @Query(value = "SELECT p.id AS \"projectId\", "
+            + "p.organization_id AS \"organizationId\", "
+            + "p.project_state AS \"projectState\", "
+            + "p.project_address AS \"projectAddress\", "
+            + "p.project_type AS \"projectType\", "
+            + "j.last_finished_at AS \"lastAssessedAt\" "
+            + "FROM project p "
+            + "LEFT JOIN (SELECT project_id, organization_id, max(finished_at) AS last_finished_at "
+            + "           FROM compliance_generation_jobs "
+            + "           WHERE status IN ('SUCCEEDED', 'NOTHING_TO_REPORT') "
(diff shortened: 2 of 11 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Reviewed this myself, since CodeRabbit was rate limited on it and it carries a migration.

**Both calls beyond the issue scope are right, and I would not have wanted it built the other way.**

The third column is the one that matters. `updated_at` moves when someone fixes a typo, and keying the sweep on it would spend a model call per approved project per tenant to re-derive an identical answer. Splitting `effective_from` out as the curated, deliberately-set field is the distinction the issue was missing rather than an addition to it.

**Verified the backfill against the live changelog on staging rather than reading it.** The first branch of the coalesce chain hits exactly:

```
037-seed-compliance-rules | db/changelog/v4.0/037-create-and-seed-compliance-rules.xml | 2026-08-21 09:30:01.399208
```

That is the load-bearing detail. If that id had not matched, the chain would have fallen through to `current_timestamp`, every seeded rule would have read as newer than every assessment, and the first pass would have queued a job for every approved project in every tenant to create nothing. Dating to the seed is what makes the first pass a no-op, so it was worth checking rather than assuming.

Also confirmed off by default in both places, and gated so the bean does not exist rather than existing and declining to run: `@ConditionalOnProperty(havingValue = "true")`, `enabled = false`, `COMPLIANCE_SWEEP_ENABLED:false`.

The three open product questions are correctly left open. Which catalogue changes justify re-assessment, how often, and whether a tenant may decline the spend are Anand's calls once the curated rule set exists, and they are all configuration now rather than a build.

One limit worth carrying forward: generation only creates, so a corrected rule never updates an inspection that already exists. That is the right behaviour for now, and it is also the thing to revisit first if re-dating `effective_from` ever becomes routine.

Merging on green CI, the mutation-proven test table, and this review.

PR #575, opened 2026-08-29
Say what the data part of a write actually carries

abhijithanandan:
Item D of #522, first half: the nineteen endpoints whose JSON payload travels as a `data` string.

## What was wrong

A multipart endpoint cannot take a `@RequestBody`, because the body is the multipart envelope, so its payload travels as a `data` part and the endpoint reads it with `JsonPartBinder`. The declared Java type of that parameter is `String`, springdoc documents what is declared, and so every one of these endpoints published `type: string` for its payload. The document named no field, no type and no constraint for any of them. Attendance check-in and clock-event are worse than undescribed: they take the payload as a `@RequestParam`, so the document advertised a string in the query string.

That is the gap tornotron/echno-core#49 ran into. It wants a contract test between the published spec and the hand-written TypeScript client, and for these endpoints there was nothing to check a field name against: `plannedStartDate` and `startDate` were equally valid, because the schema was "a string".

## Why the Java type could not simply change

The obvious fix is to declare the parameter as the payload type and let Spring bind it. It would break every client.

Both clients append the part with `formData.append('data', JSON.stringify(payload))`, which is a plain string form field carrying no `application/json` content type. Spring answers a typed `@RequestPart` with **415** in that case. Checked against Spring Boot 3.4 rather than assumed:

| part as sent | `@RequestPart String data` | `@RequestPart Payload data` |
|---|---|---|
| plain string field, no content type (what the clients send) | 200 | **415** |
| same JSON, part typed `application/json` | 200 | 200 |

So the runtime keeps the `String` and the schema is declared beside it with `@Parameter(schema = @Schema(implementation = ...))`, which springdoc merges into the multipart schema in place of the string. Nothing about how a request is parsed, validated or answered changes in this PR. Only what the document says about it.

## What is here

**Nine create endpoints and the chat send** name the payload DTO they already deserialize into: `TaskCreationDto`, `IssueCreationDto`, `OrganizationCreationDto`, `ProjectCreationDto`, `SendMessageDto`, and `AttendanceCheckInDto` / `AttendanceClockEventDto` for the four attendance endpoints, which stay documented as query parameters because that is genuinely where the client puts them.

**Six partial updates** name a new `*UpdateFieldsDto` listing the keys their service actually applies. Those endpoints keep the map at runtime on purpose: a partial update has to tell a field that was left out from one explicitly sent as null, a bean cannot (both arrive as a null property), and several of these fields are cleared by sending null. The fields class supplies the half the map cannot: it is the published schema and is never deserialized into.

**Two tests hold it.** A schema nothing enforces is a schema nothing keeps true, and a documentation class that has drifted from the code is worse than none, because it reads as authoritative.

- `PartialUpdateSchemaContractTest` reads the `case "..."` labels out of each service's update method and fails when they and the schema's fields have moved apart in either direction. Source rather than bytecode, because a string switch compiles to a hash cascade. Plain JUnit, no Spring context.
- `RequestPayloadSchemaTest` is the ratchet, in the shape of `MultipartPayloadValidationTest` beside it: a `data` payload declared as a `String` and left undescribed fails the build. Fixing the nineteen that exist does nothing about the twentieth, which is what this is for.

Both were checked against their own absence: removing one annotation fails the ratchet at the exact parameter, and deleting one field from `TaskUpdateFieldsDto` fails the contract test for that surface alone.

## Verification

Booted the real application against a Testcontainers CockroachDB and read `/v3/api-docs`. All nineteen `data` payloads resolve to a DTO schema; none is left as a bare string.

```
TYPED data payloads: 19
UNTYPED data payloads: 0
```

Nothing published consumed these as strings, so nothing downstream can break on the change: `echno-core` has no codegen, no committed spec, and no OpenAPI-derived types (`api:snapshot` diffs its own exported symbols, not any spec), and the only place it treats a body as a string beyond `JSON.stringify` is a `"salary"` float fix-up on the plain-JSON PATCH path, which none of these endpoints emits.

## What is left

The other half of item D, coming as a second PR: nineteen request bodies still publish as a free-form map, counted from this same document. Seven raw `@RequestBody Map` endpoints, six `*PatchDto.updates` batch bodies, and six on the leave and keycloak paths that this pass did not cover.

Refs #522, tornotron/echno-core#49

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/attendance/AttendanceController.java
+++ b/src/main/java/org/tornotron/echno_backend/attendance/AttendanceController.java
@@ -2,6 +2,8 @@ package org.tornotron.echno_backend.attendance;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import org.springframework.security.access.prepost.PreAuthorize;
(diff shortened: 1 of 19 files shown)


abhijithanandan:
@coderabbitai review

PR #579, opened 2026-08-29
Name the keys a map-shaped request body actually accepts

abhijithanandan:
Item D of #522, second half. Follows #575, which landed the `data` string parts.

(This replaces PR #578, which GitHub auto-closed when its stacked base branch was deleted on merge. Same branch, same commit, now based on `development`.)

## What was wrong

A request body declared as a `Map` publishes as `additionalProperties`. The document names no key, holds no type, and no key can ever fail against it. Nineteen bodies were in that state, counted from the document itself rather than from a grep.

It is not only a documentation problem. The services behind these bodies switch over a fixed key set with **no `default`**, so a key nobody recognises is a 200 and a silent no-op, and the published contract offered no way to notice. That is how #512 happened, and it is the half of tornotron/echno-core#49 a client-side test cannot work around.

## What is here

**Seven raw `@RequestBody Map` partial updates** and **seven batch bodies** whose `*PatchDto.updates` carries the same map now publish an `*UpdateFieldsDto` naming exactly the keys their switch applies. New here: employee, leave policy, leave request. Reused from #575: task, issue, project, organization, user.

They keep the map at runtime, and that is deliberate rather than a shortcut. A partial update has to tell a field that was left out from one explicitly sent as null, a bean cannot (both arrive as a null property), and several of these fields are cleared by sending null. The batch endpoints route each element through the same private method as the single update, so they name the same class.

**Four leave endpoints get a real DTO, not a description of one.** `cancelRequest` reads one key out of a `Map<String, String>` and `calculateDays` reads four; neither is a partial update, so nothing was gained by the map. The hand parsing was also doing badly what a bean does well:

| request | before | after |
|---|---|---|
| `startDate` omitted | `LocalDate.parse(null)` → NPE → **500** | `@NotNull` → **400** |
| `startDate` malformed | `DateTimeParseException` → **500** | Jackson → **400** |
| `startHalfDayType` not a valid enum | `IllegalArgumentException` → 400 | Jackson → 400 |

Both endpoints have documented 400 for exactly the first two cases since they were written. This is what makes the documented answer the real one. Nothing that was accepted before is refused now: the reason on a cancellation stays optional, and unknown keys are still ignored (`FAIL_ON_UNKNOWN_PROPERTIES` is off).

**One stays a map.** `KeycloakManagementController.addRealmRoles` takes a map of role name to description, where the keys are the roles the caller is creating. No schema could name them without being wrong, so it is allowlisted with that reason rather than forced into a shape it does not have. It is the only one of the nineteen where the map is the payload rather than a stand-in for a type nobody wrote.

## Guards

`RequestPayloadSchemaTest` grows the matching rule: an undescribed map body fails the build the way an undescribed `data` part already does. The allowlist is checked for staleness beside it, in the shape `UnboundedRepositoryReadTest` already uses, so an exemption cannot outlive what it excuses.

`PartialUpdateSchemaContractTest` extends to the three new surfaces, reading the `case` labels out of `EmployeeService.partialUpdateAnEmployee`, `LeavePolicyService.updatePolicy` and `LeaveRequestService.updateRequest`.

Both rules were checked against their own absence: removing one `@RequestBody` schema fails the first at the exact parameter, and changing `addRealmRoles` to stop taking a map fails the staleness check.

## Verification

Booted the real application against a Testcontainers CockroachDB and read `/v3/api-docs`, before and after:

```
free-form map request bodies: 19   ->   1
batch updates now typed:       0   ->   7
```

The one remaining is `POST /api/v1/keycloak/roles`, the allowlisted entry. Full `./gradlew build` green.

`echno-core` calls the leave policy and leave request updates with fixed, named key sets (`updateLeavePolicyToJson`, `updateLeaveRequestToJson`), all of which the new schemas name. It does not call any of the seven batch endpoints, or the plain `/user/{id}` and `/employee/{id}` variants, at all.

Refs #522, tornotron/echno-core#49

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/employee/EmployeeController.java
+++ b/src/main/java/org/tornotron/echno_backend/employee/EmployeeController.java
@@ -1,6 +1,8 @@
 package org.tornotron.echno_backend.employee;
 
 import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import jakarta.validation.Valid;
@@ -18,6 +20,7 @@ import org.tornotron.echno_backend.employee.dto.EmployeePatchDto;
 import java.util.List;
 import java.util.Map;
 import org.tornotron.echno_backend.common.pagination.UnpagedResultCap;
+import org.tornotron.echno_backend.employee.dto.EmployeeUpdateFieldsDto;
 
 /**
  * REST controller for managing employees.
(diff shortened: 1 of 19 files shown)


abhijithanandan:
@coderabbitai review

Issue #573, opened 2026-08-29
A stock adjustment can be approved but never rejected: the columns exist and nothing writes them


abhijithanandan:
## What is missing

`StockAdjustment` carries `rejected_by`, `rejected_at` and `rejection_reason`, and `StockAdjustmentDto` exposes all three with schema descriptions ("Reason given for rejecting the adjustment", example "Variance not supported by the count sheet"). Nothing writes them.

`StockAdjustmentControllerWeb` exposes list, paginated list, get by id, create, update, `POST /{id}/approve` and delete. There is no reject mapping, and `StockAdjustmentService` has no reject method. Grepping every `@PostMapping`, `@PutMapping` and `@PatchMapping` in the repository for a reject path returns only `NcrControllerWeb`, `LeaveApprovalControllerWeb` and `LeaveApprovalController`.

## Why it matters

An approver who looks at a draft and decides the variance is not supported by the count sheet has two options: approve it, which posts the movement, or delete it, which erases the fact that it was ever raised and why it was turned down. There is no third one. On a document whose whole purpose is to make a balance explainable, the record of a refused correction is worth as much as the record of an accepted one, and the columns to hold it are already there.

This surfaced building the web client for the approval path (echno-web #330, PR #333). The client ships approve only. The detail screen displays `rejectedBy`, `rejectedAt` and `rejectionReason` and can therefore show a rejection the moment something writes one, but it offers no reject action, because there is nothing to call.

## What it needs

- `POST /api/v1/stock-adjustments/web/{id}/reject`, taking the reason in the body. `RegularizationActionDto` is the closest existing shape for a reason-carrying action payload.
- `StockAdjustmentService.reject`, setting `rejectedBy` from the session the way `approve` sets the approver, `rejectedAt`, `rejectionReason` and the status, and refusing a document that is already posted the way `requireNotPosted` does for edit, delete and a second approval.
- The same role gate as approve, `system-admin` or `project-manager`.
- Worth deciding at the same time: whether a rejected document can be edited and resubmitted, or whether rejection is terminal and the raiser starts again. `approve` freezes the document; rejection probably should not, since the point is to send it back.

Once it exists, the web client needs the reject half added to `services/stock-adjustments-service.ts` and the detail screen, with the reason required in the form.

PR #580, opened 2026-08-29
Publish the API contract as a file instead of a running server

abhijithanandan:
Closes the last thing standing between tornotron/echno-core#49 and a contract test.

## The problem

The OpenAPI document had no artifact. It existed only in the memory of a booted application, on `GET /v3/api-docs`. No `openapi.json` was checked in, no Gradle task emitted one, no CI job produced one. Getting a copy meant running a server and curling it.

That is what blocked the contract test in `echno-core`. The client is hand-maintained with no code generation anywhere, so a renamed backend field produces no compile error there and nothing fails loudly. The test exists to catch exactly that, and it cannot check against a document that is not published. Pinning a hand-curled copy as a fixture would have been worse than no check: nothing would notice when the fixture stopped describing the backend, and the suite would go on reporting green against a contract that no longer existed.

PRs #575 and #579 made the document say something worth checking, by typing 37 of 38 previously undescribed write endpoints. This publishes it.

## What this does

`OpenApiSnapshotTest` boots the application, reads the document, writes it in a canonical form, and fails when the copy committed at `docs/openapi.json` differs.

```
./gradlew openApiSnapshot                          verify
./gradlew openApiSnapshot -PupdateOpenApiSnapshot  regenerate
```

CI runs the verify on every pull request, and uploads the served document on failure so a drift can be diffed without rerunning anything.

## Committed, not only published from CI

Both, in effect: the file is committed and CI verifies it. The argument for committing rather than publishing an artifact from a workflow:

- A renamed field arrives in review as a diff on the contract, in the same pull request as the code that renamed it. On a client with no generated types that diff is the only warning anyone gets before a user does.
- A consumer can read the contract at a git ref with no running server, which is what `echno-core` needs.

The argument for the CI check on top of that is the fixture-rot failure above: without it, a committed document is a fixture that silently stops being true.

The file is 3 MB. That is the size of the API, and it compresses to a fraction of that in the object store. It is a generated file, and the diffs on it are what make it worth having.

## Determinism

springdoc assembles paths and schemas from hash-ordered maps, so the same code can serve the same document with its members in a different order. Committing that raw would produce a diff on every regeneration and bury the real ones, so the document is re-serialised with its object keys sorted, plus the two arrays springdoc fills from a set (`required`, `tags`).

Measured, not assumed: two runs in separate JVMs against separate database containers produced byte-identical files (`md5 76f6530a7f6624bf9f254a453ffa6bd2` both times).

## The failure reads as a rename

An equality assertion on two 3 MB strings produces a failure nobody reads. Both sides are canonical and one key per line, so the lines are compared as multisets and the failure names what moved. Verified by renaming `CategoryCreationDto.icon` to `iconUrl` and running the check:

```
The committed OpenAPI document no longer matches what this code serves.

  committed: docs/openapi.json
  served:    build/openapi.json
...
Only in the committed document (removed or renamed) (1):
  "icon": {
Only in what this code serves (added or renamed) (1):
  "iconUrl": {
```

Reverted after the check, so the branch carries no mutation.

## Its own Gradle task, not `test` and not `check`

It boots a web context. The main suite runs in a 1 GB JVM with `spring.test.context.cache.maxSize = 8` and no `forkEvery`, where an extra context is not free and OOMs an unrelated test rather than itself. A separate task means a separate forked JVM and the cost lands nowhere near the suite; `test` excludes the class explicitly.

It is not wired into `check` either, so a local `./gradlew build` does not pay for a second context boot and a second database container for something CI verifies on every pull request.

## The property that would have made this answer 401

`springdoc.swagger-ui.public-access` ships as false since #569, so `/v3/api-docs` answers 401 to an unauthenticated caller. springdoc still builds the document; Spring Security just refuses to hand it over. The test sets the property true for its own context, which changes no deployed default and no ingress rule.

## Also here

`CategoryPatchDto` is deleted. It has no reference anywhere in the codebase and appears in no operation, so it is absent from the document it looks like it describes. A reader would take it for the category patch contract; it is dead code. This is the note handed over from the #522 item D work, alongside the observation that the remaining typed `*PatchDto` classes now split cleanly into the ones that name their fields and the single one that genuinely cannot (`KeycloakManagementController.addRealmRoles`, allowlisted with its reason in `RequestPayloadSchemaTest`).

## Verification

Full suite, coverage gate and the snapshot check on the lab box:

```
./gradlew test jacocoTestCoverageVerification openApiSnapshot
BUILD SUCCESSFUL in 8m
```

## What this does not do

Nothing here changes what any deployment serves, or who can read the document from one. That is still `SWAGGER_PUBLIC_ACCESS`, still false by default, still a per-environment decision.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on the published OpenAPI contract, snapshot generation, and authentication requirements for API documentation endpoints.

* **Chores**
  * Added automated checks to detect differences between the published and served OpenAPI specifications.
  * Failed builds now provide the served OpenAPI document for easier comparison.

* **Refactor**
  * Removed an obsolete category batch-update data structure.


abhijithanandan:
@coderabbitai review

abhijithanandan:
Reviewed this myself, since CodeRabbit was rate limited on it and on the core side both.

Three design calls I want on the record because they are the difference between a check that works and one that decays:

**Committed and CI-verified is the right pairing.** A committed document turns a renamed field into a diff sitting beside the code that renamed it, which on a hand-maintained client with no generated types is the only warning anyone gets. A committed-but-unchecked document would be a fixture that silently stops being true, which is worse than no check at all. Both halves are needed.

**Its own Gradle task rather than `test` or `check`** correctly sidesteps the heap trap: a separate task is a separate forked JVM, so the extra web context never lands in the suite JVM. Worth noting #481 merged today and raised that cap to 2048m with a gate at 80%, so this is now belt and braces rather than load-bearing.

**Not `springdoc-openapi-gradle-plugin`**, because it forks `bootRun` and needs a live database plus the full Keycloak and Spaces environment in CI. The Testcontainers path already works and Docker is already required by `test`. Right call, and the reasoning is the useful part.

The determinism work was not optional and I am glad it was not skipped: springdoc serialises from hash-ordered maps, so without canonicalisation this file would have produced spurious diffs on every regeneration and been ignored within a fortnight. Two runs in separate JVMs against separate containers giving byte-identical output is the evidence that matters.

That it caught real drift mid-flight, when the #573 reject endpoint landed underneath it and CI failed exactly as designed, is a better proof than any test I could ask for.

Merging first, since core #56's index reads `docs/openapi.json` from `development` and cannot resolve until this lands.

abhijithanandan:
@coderabbitai review

PR #581, opened 2026-08-29
Give the test run room, and a voice when it runs out

### LINKED ISSUE #481: The test suite runs in one JVM with no forkEvery, so it is close to its heap cap and fails misleadingly

abhijithanandan:
Closes #481.

## What the suite actually costs

Measured on the lab build box, full suite, 1310 tests. The instrumentation added here is what produced these numbers, so they can be re-taken at any time.

| run | heap cap | context cache | peak live set | occupancy high water | result | wall clock |
|---|---|---|---|---|---|---|
| as merged today | 1024m | 8 | 506 MB (49%) | 618 MB | pass | 7m51s |
| all contexts held | 4096m | 64 | **1150 MB** | 1316 MB | pass | 8m14s |
| all contexts held | 1024m | 32 (Spring default) | over the cap | | **OOM** | 8m41s |
| all contexts held | 2048m | 32 | 1150 MB (56%) | 1164 MB | pass | 8m00s |
| **this branch** | **2048m** | **8** | **485 MB (24%)** | 645 MB | pass | 8m02s |

The suite is **31 distinct Spring test configurations**, so 31 contexts and 31 Hibernate `SessionFactory` instances over the 98 entities, all in one JVM because there is no `forkEvery`. Holding all of them costs 1150 MB, which is more than the 1024m cap. The run fits today only because #497 pinned the context cache at 8.

That is the real finding. The cap was not close to binding in the configuration we ship, but it was below the suite's own working set, so the cache limit was the only thing between the run and the failure, and nothing said so.

## Reproduced, then absorbed

Same branch, same tests, cache at Spring's default of 32, only the heap cap differs:

- **1024m**: 1297 tests pass, **zero assertion failures**, then `java.lang.OutOfMemoryError: Java heap space` inside `com.tngtech.archunit.core.domain.DomainObjectCreationContext.createJavaMethod`, surfacing as a failure of `EndpointAuthorizationTest`, a test the change never touched. Gradle reports only "There were failing tests. See the report at ...".
- **2048m**: passes, 1150 MB live of the 2048 MB cap.

## Why 2048m and what it costs CI

It is the smallest round number that holds the measured 1150 MB working set with real spare, so the cache cap becomes a performance choice rather than the only thing preventing a misleading failure.

`ubuntu-latest` on a public repository is 4 vCPU and 16 GB, not the 2 vCPU and 7 GB the old comment was sized against. The test JVM shares the runner with the CockroachDB container `AbstractIntegrationTest` already limits to 2 GB, and the Gradle daemon. 2 GB + 2 GB + daemon leaves roughly 11 GB unused. The CI run on this PR reports the runner's memory in its own log, from `/proc/meminfo`, so the assumption is checked rather than asserted.

## The other two levers, measured rather than argued

- **`forkEvery`** bounds accumulation but restarts everything per fork, and `AbstractIntegrationTest` starts its CockroachDB container in a static block, so a new fork means a new container as well as a fresh set of contexts. Measured on this configuration: `forkEvery = 50` takes **28m51s against the 8m02s baseline**, 3.6x, for 20 extra minutes on every push. It does work, holding the peak live set to 387 MB, but it is paying twenty minutes a run for headroom that 1 GB of unused runner memory buys for nothing. Not adopted.
- **Fewer contexts** attacks the cause, and 31 configurations for 1310 tests is not obviously wrong, but it is a large change across 54 `@DataJpaTest` classes and does not belong in this one.

Keeping the cache at 8 costs **one** extra context build over the whole run (37 loads against 36 with the cache at 64) and no measurable wall clock, so it stays.

## Making the ceiling visible

A number in a log nobody reads is not a warning, so there are three layers, in the order they would fire.

1. **Every build prints one line.** `testHeapHeadroom` finalizes `test`, so it reports on failed runs too:

   `Test heap: 485 MB live of a 2048 MB cap (24% used, 1563 MB spare, 645 MB occupied at the fullest), 8 Spring contexts cached of a maximum 8 (37 loaded, 4346 reused), on a machine with 15.6 GB`

2. **It warns at 65% of the cap and fails the build at 80%**, before the heap actually runs out, naming the three levers. Verified firing:

   ```
   > Task :testHeapHeadroom FAILED
   > Test heap: 138 MB live of a 512 MB cap (27% used, ...)
     That is past the 10% mark, which means the next test added here is likely to
     fail in a class it never touched, as a context-loading error rather than an
     assertion. Failing now, while the cause is still legible.
   ```

3. **If it runs out anyway, it says so plainly.** `-XX:+ExitOnOutOfMemoryError` collapses the cascade of unrelated context failures into one exit, `-XX:+HeapDumpOnOutOfMemoryError` leaves the evidence, and the task explains what happened. Same OOM as the reproduction above, on this branch:

   ```
   Test heap: no measurement was written. If the run failed, the test JVM probably
   died before it could finish; look for a heap dump in .../build/reports/test-heap
   and read the failures as a memory problem, not a test one.

   > Process 'Gradle Test Executor 1' finished with non-zero exit value 3
   ```

## How it measures

`TestHeapHeadroomListener` registers through `META-INF/services`, so it applies to every test JVM without any test knowing about it. It samples the **live set**, what survives a collection, every 15 seconds and once after the last test, rather than heap occupancy: occupancy on this suite peaks at 645 MB against a 506 MB live set, because a collector under no pressure has no reason to collect early, and gating on it would have the build permanently declaring an emergency. Occupancy is still reported, next to the number that matters.

`testMaxHeap`, `testContextCacheSize`, `testHeapFailRatio` and `testHeapWarnRatio` are all `-P` overridable, which is how every row in the table above was taken.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,37 @@ ext {
     // measured), then to 25% after the leave-accrual, employee-hierarchy,
     // attendance-regularization and project-invite-code service tests (25.50% measured).
     jacocoLineFloor = 0.25
+
+    // Heap cap for the test JVM, the size of the Spring context cache, and the two thresholds
+    // the testHeapHeadroom gate applies to the first. All four are overridable per run with
+    // -P, which is how the numbers below were measured rather than guessed. Every figure here
+    // is from a full run of the suite (1310 tests) on the lab build box.
+    //
+    // The suite is 31 distinct Spring test configurations, so 31 contexts and 31 Hibernate
+    // SessionFactory instances over the 98 entities, all in one JVM because there is no
+    // forkEvery. Holding all of them costs 1150 MB live. That is more than the 1024m this cap
+    // used to be, which is why a run with the context cache left at Spring's default fails:
+    // 1297 tests pass, then the architecture tests need the whole class graph resident, the
+    // heap has nothing left, and the run ends with an OutOfMemoryError inside ArchUnit and
+    // zero assertion failures anywhere. Capping the cache at 8 is what has been holding that
+    // off, at a peak live set of 485 MB and a cost of exactly one extra context build over
+    // the run (37 loads against 36) and no measurable wall clock.
+    //
+    // The cap is now sized so that the cache cap is a choice rather than the only thing
+    // standing between the suite and that failure: 2048m holds the whole 1150 MB working set
+    // with 40% to spare even if the cache limit were removed, and leaves the run as actually
+    // configured at a quarter of the cap. The CI runner has room for it. ubuntu-latest on a
+    // public repository is 4 vCPU and 16 GB, not the 2 vCPU and 7 GB this was first sized
+    // against, and the test JVM shares it only with the CockroachDB container that
+    // AbstractIntegrationTest already limits to 2 GB, plus the Gradle daemon.
+    testMaxHeap = findProperty('testMaxHeap') ?: '2048m'
+    testContextCacheSize = findProperty('testContextCacheSize') ?: '8'
+    // Above this fraction of the cap the build stops and explains itself; above the lower one
+    // it warns. The gap between them is deliberately wide, because what it is watching moves
+    // in steps: each new test configuration is another whole context, worth roughly a
+    // thirtieth of the full working set, and each new entity grows every context at once.
+    testHeapFailRatio = (findProperty('testHeapFailRatio') ?: '0.80') as double
+    testHeapWarnRatio = (findProperty('testHeapWarnRatio') ?: '0.65') as double
 }
 
 repositories {
@@ -71,7 +102,9 @@ dependencies {
     testImplementation 'com.github.dasniko:testcontainers-keycloak:3.7.0'
     testImplementation 'com.tngtech.archunit:archunit-junit5:1.3.0'
     runtimeOnly 'org.postgresql:postgresql'
-    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
+    // Compile scope rather than runtime only: TestHeapHeadroomListener implements the
+    // launcher's TestExecutionListener, so the interface has to be on the test compile path.
+    testImplementation 'org.junit.platform:junit-platform-launcher'
 }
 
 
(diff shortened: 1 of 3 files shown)


abhijithanandan:
@coderabbitai review

PR #582, opened 2026-08-29
Give AR invoices a way to be listed, scoped to the tenant asking

### LINKED ISSUE #576: AR invoices cannot be listed: the only read mapping is by id, so the issue and cancel endpoints are unreachable from any screen
Found while building the AR invoice issue/cancel client for echno-web #334, the same way the missing stock-adjustment reject endpoint was found in web #333.

## What is missing

`InvoiceControllerWeb` exposes four mappings and no more:

- `POST /api/v1/finance/invoices/web` create draft
- `GET /api/v1/finance/invoices/web/{id}` fetch one by id
- `POST /api/v1/finance/invoices/web/{id}/issue`
- `POST /api/v1/finance/invoices/web/{id}/cancel?reason`

There is no way to list or search AR invoices. Grepped every `@GetMapping` on that controller across `development` and all 30 remote branches: the only read mapping on any of them is `/{id}`. No other controller returns `InvoiceDto` either, so the whole repo has no `List<InvoiceDto>` or `Page<InvoiceDto>` anywhere.

The queries are already written and unused. `InvoiceRepository` carries `findByCustomerId(UUID, Pageable)`, `findByStatus(InvoiceStatus, Pageable)` and `findOpenInvoicesByCustomer(UUID)`, and nothing on `InvoiceService` or the controller calls any of them.

## Why it blocks the client

`echno-core` ships `financeInvoiceService` with `issue` and `cancel`, and web #334 asks for a screen that calls them. A screen cannot be built, because there is no way for a user to arrive at an AR invoice:

1. No list endpoint, so no index page can exist.
2. `ConstructionInvoiceDto` does carry `arInvoiceId`, but `echno-core`'s `ConstructionInvoice` type drops it, so the construction invoice cannot link across to the receivable it raised.
3. Even if it did link, those are exactly the invoices on which neither action is usable. An AR invoice materialized by a construction invoice is already `ISSUED`, so `issue` refuses it, and `InvoiceService.cancel` explicitly refuses it too: "Invoice was raised for construction invoice X and cannot be cancelled on its own".

So the only invoices `issue` and `cancel` apply to are standalone drafts created through the API, and a standalone draft is unreachable from the product the moment the caller loses its id.

## What it needs

A read endpoint on `InvoiceControllerWeb` in the shape the other finance controllers already use, wired to the repository queries that exist:

```
GET /api/v1/finance/invoices/web/all?pageNo&pageSize[&status][&customerId]
```

returning `Page<InvoiceDto>` (or `List<InvoiceDto>` from the page content, matching `JournalEntryControllerWeb.getAll`), behind the same `@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin', 'project-manager')` gate as the rest of the controller.

Optionally, surface `arInvoiceId` through the `echno-core` `ConstructionInvoice` type so a construction invoice can link to the receivable it raised. That is a core change rather than a backend one, and it is a convenience, not the fix: the list endpoint is what unblocks the screen.

Once this lands, web #334 can be built. It is blocked until then.

abhijithanandan:
Closes #576.

## What was wrong

`InvoiceControllerWeb` had four mappings and no way to enumerate anything: create a draft, `GET /{id}`, issue, cancel. Nothing in the repo returned a `List` or `Page<InvoiceDto>` on `development` or on any remote branch. A screen can only act on an invoice once it can put one in front of a user, so the issue and cancel endpoints were unreachable in practice, which is why the client for them (echno-web #334) was refused rather than built against a void.

`InvoiceRepository` already carried `findByCustomerId`, `findByStatus` and `findOpenInvoicesByCustomer`, all three uncalled. The listing was intended and never wired.

## The contract

```
GET /api/v1/finance/invoices/web
      ?pageNo=&pageSize=          bounded by PageQuery
      [&customerId=<uuid>]
      [&status=DRAFT|ISSUED|PARTIALLY_PAID|PAID|CANCELLED]
      [&openOnly=true]
  -> 200 Page<InvoiceDto>
     400 pageNo/pageSize outside its permitted range
     403 caller lacks the role in the current tenant
```

Bare `GET` on the collection rather than the `/all` the issue sketched: `/all` exists on `JournalEntryControllerWeb` only because a bare `GET` there is taken by its by-id lookup. Nothing takes it here, and the finance and inspection siblings that were written since (`ConstructionInvoiceControllerWeb`, `ConstructionPaymentControllerWeb`, `InspectionControllerWeb`, `NcrControllerWeb`, `ChecklistTemplateControllerWeb`) all list on the bare collection path and return `Page<Dto>`. `echno-core` already unwraps that envelope for construction invoices, so the client side has a parse helper to copy.

Rows are ordered newest invoice date first, with the invoice number breaking a same-day tie. A page with no order is a page in whatever order the storage engine produced, which on a distributed engine is not stable between two requests: a row can be served on two pages while another is served on none.

## Filters

Wired:

| Parameter | Why |
|---|---|
| `customerId` | The receivables view a person actually asks for. Was `findByCustomerId`. |
| `status` | Reaching a draft to issue it, or an issued invoice to cancel it, which is exactly what #334 needs. Was `findByStatus`. |
| `openOnly` | The one filter `status` cannot express: what is still owed spans `ISSUED` and `PARTIALLY_PAID`. Was `findOpenInvoicesByCustomer`. |

`openOnly` and `status` are independent and both apply when both are given, so `openOnly=true&status=PAID` is a legal request matching nothing. That is the honest answer to a contradictory filter and keeps the two parameters from having to know about each other.

Left out until a screen asks: invoice-date and due-date ranges, overdue-only (needs a clock in the query and a policy on what counts as overdue), text search on the invoice number or customer name, amount ranges, and a caller-chosen sort.

The three uncalled repository methods are gone, replaced by `InvoiceSpecifications`. One derived method per filter cannot answer two filters at once without a method for every pair, and `findOpenInvoicesByCustomer` also read an unbounded `List`.

## Tenant isolation

Scoping stays where it already is: the Hibernate `orgFilter` the `Invoice` entity declares and `HibernateFilterConfig` enables on the transaction. `JpaSpecificationExecutor.findAll(Specification, Pageable)` is a criteria query on the entity, so the filter covers the row query and the count query behind `getTotalElements` alike. A member of one tenant can neither enumerate another's invoices nor learn how many they have. Restating the scope as a predicate would give a reader two places to check and one of them could drift.

`InvoiceServiceIT` now seeds two organizations and asks from inside each in turn, asserting on ids rather than counts alone so swapping one row for another cannot pass, and asserting on the total so a leaked count is caught even when no foreign row is returned. A second test asks for the other tenant's customer by id and gets nothing, which closes the way through the filters that the unfiltered case would not show.

## Every new test was made to fail

Two mutation runs on `lab-node-116-f`, each reverted afterwards.

| Mutation | Test that failed | Tests that stayed green |
|---|---|---|
| `@Filter(name = "orgFilter")` removed from `Invoice` | `findAll_returnsOnlyTheInvoicesOfTheTenantAsking`, `findAll_filtersByCustomerAndCannotReachAnotherTenantsCustomer` | every other test in both classes, incl. `ConstructionInvoiceServiceIT`'s own tenant-scope test |
| `LIST_ORDER` dropped from the `PageRequest` | `findAll_pagesInAFixedNewestFirstOrder` | |
| `customerId` predicate dropped | `findAll_filtersByCustomerAndCannotReachAnotherTenantsCustomer` | |
| `status` predicate dropped | `findAll_filtersByStatus` | |
| `openOnly` predicate dropped | `findAll_openOnlyReturnsWhatIsStillOwed` | |
| controller ignores `PageQuery`, hardcodes page zero | `theCallersPageBoundsReachTheService` | the other three `InvoiceListingTest` cases |

## Conventions

- Page bounds come from `common/pagination/PageQuery`, so no `pageNo`/`pageSize` parameter is declared on the handler and `PaginationParameterBoundTest` passes. `pageSize` is bounded by the same `UnpagedResultCap.MAX_ROWS` ceiling as every other paginated endpoint.
- `UnboundedRepositoryReadTest` passes: the read takes a `Pageable`, and the one unbounded `List` query on this repository was deleted rather than allowlisted.
- `CappedWebListingTest` is untouched. It covers the bare listings that take no page parameters at all; this one takes them, so it belongs to the paginated family instead.
- `@PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin', 'project-manager')")`, the same gate the four existing mappings carry. Everything on this controller posts to or reverses the ledger, and a listing that showed less than the actions can reach would only be confusing.
- No schema change, so no changelog.

## Not touched

Construction invoices keep their own controller and their own lifecycle. An AR invoice a construction invoice raised is already `ISSUED` and `InvoiceService.cancel` refuses it by name, so this listing makes those rows visible without making either action apply to them. Surfacing `arInvoiceId` through the `echno-core` `ConstructionInvoice` type, which #576 raised as an optional extra, is a core change and is not here.

## Verification

`InvoiceServiceIT` (13 tests, 5 new), `InvoiceListingTest` (4 new), the whole `architecture` package and the whole `common.pagination` package all pass on `lab-node-116-f` against a real CockroachDB 26.2.4.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/finance/invoice/repositories/InvoiceRepository.java
+++ b/src/main/java/org/tornotron/echno_backend/finance/invoice/repositories/InvoiceRepository.java
@@ -1,21 +1,34 @@
 package org.tornotron.echno_backend.finance.invoice.repositories;
 
 import jakarta.persistence.LockModeType;
-import org.springframework.data.domain.Page;
-import org.springframework.data.domain.Pageable;
-import org.springframework.data.jpa.repository.EntityGraph;
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
+import org.springframework.data.jpa.repository.EntityGraph;
 import org.springframework.data.jpa.repository.Lock;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.query.Param;
-import org.tornotron.echno_backend.finance.invoice.InvoiceStatus;
 import org.tornotron.echno_backend.finance.invoice.domain.Invoice;
 
-import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
 
-public interface InvoiceRepository extends JpaRepository<Invoice, UUID> {
+/**
+ * Reads and writes for customer (accounts-receivable) invoices.
+ *
+ * <p>The list read goes through {@link JpaSpecificationExecutor}, whose
+ * {@code findAll(Specification, Pageable)} is a criteria query on the entity and therefore carries
+ * the Hibernate {@code orgFilter} declared on {@link Invoice}. That is what keeps one tenant's
+ * listing from enumerating another's, and it is why the filters are built as predicates rather
+ * than as derived query methods: a derived method per filter combination multiplies the surface
+ * that has to be proven scoped, and it cannot express two filters at once without a method for
+ * every pair.
+ */
+public interface InvoiceRepository extends JpaRepository<Invoice, UUID>, JpaSpecificationExecutor<Invoice> {
+
+    /**
+     * Org-scoped lookup by id with the lines, their revenue accounts and the customer fetched.
+     * Uses JPQL rather than a primary-key {@code find()} so the {@code orgFilter} is applied.
+     */
     @EntityGraph(attributePaths = {"lines", "lines.revenueAccount", "customer"})
     @Query("SELECT i FROM Invoice i WHERE i.id = :id")
     Optional<Invoice> findByIdWithLines(@Param("id") UUID id);
@@ -23,11 +36,4 @@ public interface InvoiceRepository extends JpaRepository<Invoice, UUID> {
     @Lock(LockModeType.PESSIMISTIC_WRITE)
     @Query("SELECT i FROM Invoice i WHERE i.id = :id")
     Optional<Invoice> findByIdForUpdate(@Param("id") UUID id);
-
-    Page<Invoice> findByCustomerId(UUID customerId, Pageable pageable);
-    Page<Invoice> findByStatus(InvoiceStatus status, Pageable pageable);
-
-    @Query("SELECT i FROM Invoice i WHERE i.customer.id = :customerId " +
-            "AND i.status IN ('ISSUED','PARTIALLY_PAID') ORDER BY i.invoiceDate ASC")
-    List<Invoice> findOpenInvoicesByCustomer(@Param("customerId") UUID customerId);
 }
--- a/src/main/java/org/tornotron/echno_backend/finance/invoice/repositories/InvoiceSpecifications.java
+++ b/src/main/java/org/tornotron/echno_backend/finance/invoice/repositories/InvoiceSpecifications.java
@@ -1,0 +1,65 @@
+package org.tornotron.echno_backend.finance.invoice.repositories;
+
+import jakarta.persistence.criteria.Predicate;
+import org.springframework.data.jpa.domain.Specification;
+import org.tornotron.echno_backend.finance.invoice.InvoiceStatus;
+import org.tornotron.echno_backend.finance.invoice.domain.Invoice;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+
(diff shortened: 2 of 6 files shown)


abhijithanandan:
@coderabbitai review

Issue #587, opened 2026-08-30
employee.reporting_manager is an unmapped column that is null on every row


abhijithanandan:
Split out of #520, which asked for a look at this while the employee table was open.

## What it is

`employee.reporting_manager` is a `VARCHAR(255)` that has been in the schema since
`v1.0/009-create-employee.xml` and has been carried through every baseline squash since
(`v2.0`, `v3.0`, `v4.0` all recreate it). It is not the duplicate of `manager_id` it looks
like from the DDL: it is weaker than that.

- **No Java code maps it.** `Employee` has no `reportingManager` field, and the string
  `reporting_manager` appears nowhere in `src/` outside the changelog files. Hibernate never
  reads or writes it, so nothing in the application can put a value there.
- **It is null on every row on staging.** Measured on the IITM staging `echno` database:

  ```
  total | reporting_manager_set | manager_id_set
  ------+-----------------------+---------------
     30 |                     0 |              4
  ```

- `manager_id` is the real relationship. It is the self-referencing foreign key behind
  `Employee.manager`, validated by `EmployeeHierarchyService.validateManager` (same
  organization, not self, no cycle), and read by `findByManager_Id` for the subordinates
  listing.

So there is one working representation of a reporting line and one dead column beside it
that only ever looks like a second one.

## Why it is worth removing

The same reason #520 removed `is_manager`: a column that names a real concept and holds
nothing is an invitation to write code against it. Someone reading the table sees a
`reporting_manager` and a `manager_id` and has to work out which one means anything, and the
answer is not visible from the schema. `is_manager` shows how that ends, with a query nobody
noticed was returning an empty list.

## What it needs

A Liquibase changeset dropping `employee.reporting_manager`, and nothing else: there is no
entity field, DTO field, or query to remove with it, and no data to migrate, since the column
is null everywhere.

Worth confirming the same is true of any client instance before the drop reaches production,
though a column no application code has ever written cannot hold anything the application put
there.

PR #588, opened 2026-08-30
Read an organization's managers from the org role, not a flag nobody set

### LINKED ISSUE #520: employee.is_manager disagrees with Keycloak org roles: it is false on every row while four users hold manager roles
## What I found

Measured on the IITM staging database and realm on 2026-08-28.

`employee.is_manager` is `false` on **all 28 rows**, across both organizations. Meanwhile the Keycloak side has five org-role assignments, four of which are manager roles by `OrgRole.getManagerRoles()`:

| User | Org | Keycloak role | Manager role? |
|---|---|---|---|
| `hrishikesh-ajith` | org-1 | system-admin | yes |
| `abin-johny` | org-2 | system-admin | yes |
| `echno-admin` | org-2 | system-admin | yes |
| `roman-hicks` | org-2 | system-admin | yes |
| `alexander-kelly` | org-2 | project-manager | yes |

So the application table and the identity provider disagree about who manages anything, and the table's answer is "nobody".

## Why it matters

We now have two sources of truth for the same fact and no reconciliation between them:

- `OrgRole.getManagerRoles()` (SYSTEM_ADMIN, ORG_MANAGER, HR_ADMIN, PROJECT_MANAGER) is the authority that `@PreAuthorize` actually checks, derived from the JWT group claim.
- `employee.is_manager` is a plain boolean column, plus there is a separate `employee.manager_id` self-reference for reporting lines.

If any feature branches on `is_manager`, it is dead code today: the branch can never be taken. If nothing branches on it, it is a column that looks meaningful in the schema and is not.

There is a related smell in the same table: `reporting_manager` is a `varchar` sitting next to `manager_id`, which is the foreign key that should carry that relationship.

## What needs deciding

1. Which is authoritative for "is this person a manager", the Keycloak org role or the column?
2. If the role is authoritative, is `is_manager` derived (and should be a view or a computed accessor rather than stored), or should it be dropped?
3. If the column is authoritative for something the role is not (an org-chart reporting line, distinct from a permission), then name that distinction explicitly, because "manager" currently means two different things in one schema.
4. Whatever the answer, the QA seed should produce data consistent with it. Right now the seed creates 28 employees and assigns five org roles, and nothing keeps those two halves in step.

## How to reproduce

```sql
select count(*) filter (where is_manager) as managers, count(*) as total from employee;
```

against the staging `echno` database, and the group membership listing from the Keycloak Postgres:

```sql
select g.name as grp, p.name as parent, u.username
from user_group_membership m
join keycloak_group g on g.id = m.group_id
left join keycloak_group p on p.id = g.parent_group
join user_entity u on u.id = m.user_id
where p.name is not null;
```

abhijithanandan:
Closes #520.

## The bug

`GET /api/v1/employee/web/managers/organizationId/{id}` returns an empty list for every
organization, and always has. Two sibling methods in `EmployeeHierarchyService` answered the
same question from different places:

```java
readAllTheManagers()                   -> findEmployeesByOrgRoles(OrgRole.getManagerRoles())
readAllTheManagersByOrganizationId(id) -> findEmployeesByOrganization_IdAndIsManager(id, true)
```

Nothing has ever written `employee.is_manager`. On IITM staging it is `false` on all 30 rows,
across all four organizations, while five people hold a role in `OrgRole.getManagerRoles()`.
So the second query cannot return a row, for any organization, ever, while the first one a
method above it returns the right people.

## Which side is authoritative

The organization role, and the codebase had already decided that three times over:

- `@PreAuthorize` checks `OrgRole.getManagerRoles()`, derived from the token's group claim.
- `echno-core`'s `isManager(roles)` is a function of the roles array, and `use-authorization.ts`
  consumes it. No client reads the column.
- `EmployeeMapper` deliberately left `isManager` unset, so the response field was null on every
  response the API has ever served.

The scoped read now uses the same source as its sibling, narrowed to one organization.

## What else goes

With that query gone the column has no readers at all. The two remaining queries against it,
`findEmployeesByIsManager` and `findAllManagerIds`, had no callers anywhere in main or test, so
`is_manager` had become a schema field that names a real concept and holds nothing. That is
exactly how the next person writes another query that silently returns nothing, so the column,
those two queries, and the entity field go, in changeset `070-drop-employee-is-manager`.

The two DTO fields go with them. `EmployeeDto.isManager` was mapped to null on every response,
and `EmployeeCreationDto.isManager` bound as `manager` rather than `isManager` and was never read; neither name appears anywhere in
`echno-core` or `echno-web`, and unknown request properties are ignored, so an older client
still sending the field is unaffected. Snapshot regenerated.

Keeping the column and backfilling it from Keycloak was the alternative. It is only worth doing
if something needs to filter managers in SQL without joining the roles, and
`findEmployeesByOrgRoles` shows nothing does.

## Proof

`EmployeeManagerListingIT` seeds the staging shape, role holders and no flag, against a real
CockroachDB. Against `development` all three tests fail with the listing returning `[]`:

```
Expecting actual:
  []
to contain exactly in any order:
  ["Priya Manager", "Sam Admin"]
```

Against this branch all three pass. One of them asserts the scoped and unscoped listings agree,
which is the property that was broken.

## Related

`employee.reporting_manager` came up in the same table and is filed separately as #587: it is an
unmapped `varchar` that no Java code touches and that is null on all 30 staging rows, with
`manager_id` carrying the real relationship. Not changed here.

## One deployment note

The drop ships in the same release as the code that stops reading the column, rather than a
release later. On the compose staging that is safe, because the backend is replaced rather than
rolled. On the k8s deployment a rolling update starts the new pod, which runs Liquibase, while
old pods are still serving, and those old pods still name `is_manager` in every employee select,
so they would fail for the length of the rollout. It is seconds on a staging cluster and there
is no production instance yet, which is why it is one release and not two, but it is a
contract-phase migration and worth knowing before the same pattern reaches a client.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -4714,9 +4714,6 @@
             "format": "date-time",
             "type": "string"
           },
-          "manager": {
-            "type": "boolean"
-          },
           "managerId": {
             "description": "Id of this employee's reporting manager.",
             "example": 5,
@@ -4848,11 +4845,6 @@
             "example": 7,
             "format": "int64",
             "type": "integer"
-          },
-          "isManager": {
-            "description": "Whether the employee is a manager.",
-            "example": false,
-            "type": "boolean"
           },
           "joiningDate": {
             "description": "Date the employee joined.",
@@ -32547,7 +32539,7 @@
     },
     "/api/v1/employee/web/managers": {
       "get": {
-        "description": "Returns every employee flagged as a manager in the current tenant.",
+        "description": "Returns every employee holding a manager organization role in the current tenant.",
         "operationId": "getAllTheManagers",
         "responses": {
           "200": {
@@ -32652,7 +32644,7 @@
     },
     "/api/v1/employee/web/managers/organizationId/{organizationId}": {
       "get": {
-        "description": "Returns every employee flagged as a manager within the given organization.",
+        "description": "Returns every employee holding a manager organization role within the given organization.",
         "operationId": "getAllManagersForAnOrganization",
         "parameters": [
           {
--- a/src/main/java/org/tornotron/echno_backend/employee/Employee.java
+++ b/src/main/java/org/tornotron/echno_backend/employee/Employee.java
@@ -119,9 +119,11 @@ public class Employee implements TenantScopedEntity {
     @Column(name = "date_of_birth",nullable = false)
     private LocalDateTime dateOfBirth;
 
-    @Column(name = "is_manager", nullable = false)
-    private boolean isManager = false;
-
+    /**
+     * The organization roles this employee holds, mirrored from the Keycloak groups on their
+     * token. This is the only record of whether someone manages: there is no separate boolean,
+     * because a second copy of the same fact only ever drifted from this one.
+     */
     @ElementCollection(fetch = FetchType.EAGER)
     @CollectionTable(name = "employee_org_roles", joinColumns = @JoinColumn(name = "employee_id"))
     @Column(name = "role")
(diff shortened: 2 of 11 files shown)


abhijithanandan:
@coderabbitai review

PR #589, opened 2026-08-30
Settle leave authorization on the record, not on the id the caller sent

abhijithanandan:
Found beside the OpenAPI contract check (#580 / echno-core#57), which flagged `POST /leave-requests/web` as guarded only by `isMemberOfCurrentTenant()` while taking an `employeeId`. Reading the family through settled what that allows, and it is worse than the one endpoint.

## What was reachable

**Create.** `POST /leave-requests/web` and `POST /leave-requests` were guarded by tenant membership alone and took the employee as a query parameter. Any member of a tenant could file leave in a colleague's name. The saved request carries the colleague's employee record, holds days against their balance, and enters their approval chain, so it is not a cosmetic attribution: `LeaveRequestService.createRequest` looks the employee up by the id it was handed and never asks who is calling.

**Update, submit, cancel, withdraw.** These were guarded by `isSelfInCurrentTenant(#employeeId)`, which checks the caller against an employee id the caller chose, and never against the request being acted on. `updateRequest(requestId, updates)` and its siblings load by `requestId` alone. So passing your own employee id alongside a colleague's request id passed the guard and then edited, submitted, cancelled or withdrew their request. Cancelling an approved one sets `CANCELLED`, clears the approver and restores the used days.

The UI pins every one of these to the signed-in employee, so nothing is exploitable through the product. The API is another matter, and this is the layer that has to hold.

## The fix

The employee id is an argument, not evidence. Ownership is resolved from the stored record and checked in `LeaveRequestService`, where the record is: the employee the request belongs to, or a `system-admin` / `hr-admin`, and nobody else. Create is the one call with no record to read yet, so it checks the employee it is given, which is exactly the check that was missing.

Putting it in the service rather than the annotation is deliberate. A `@PreAuthorize` on these handlers can only read what the caller sent, which is the hole itself. HR filing or cancelling on somebody's behalf stays possible, and is covered.

## The non-web controller was refusing everything

`LeaveRequestController`'s update, submit, cancel and withdraw carried `isSelfInCurrentTenant(#employeeId)` while declaring no `employeeId` parameter. SpEL resolves an undefined variable to null, so the check ran as `isSelfInCurrentTenant(null)`, found no employee and refused every call. Those four are now the tenant gate the annotation can actually evaluate, with ownership answered in the service alongside the web family's, so they work and are guarded rather than being dead and guarded by accident.

## Tests

`LeaveRequestActorAuthorizationTest`, eight cases: create for a colleague refused before anything is written, create for yourself and create by a leave admin still allowed, and update/submit/cancel/withdraw on somebody else's request refused with nothing saved, plus the owner's own cancel still going through. All eight fail with the service check removed; the five refusals fail by the action succeeding.

Full suite green on `.116`, and `docs/openapi.json` regenerated for the four changed 403 descriptions.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -65489,7 +65489,7 @@
                 }
               }
             },
-            "description": "Caller is not a member of the current tenant"
+            "description": "Caller is neither the employee named nor a holder of the system-admin or hr-admin role"
           },
           "404": {
             "content": {
@@ -66565,7 +66565,7 @@
                 }
               }
             },
-            "description": "Caller is not the employee identified by the id"
+            "description": "Caller is neither the employee the request belongs to nor a holder of the system-admin or hr-admin role"
           },
           "404": {
             "content": {
@@ -66688,7 +66688,7 @@
                 }
               }
             },
-            "description": "Caller is not the employee identified by the id"
+            "description": "Caller is neither the employee the request belongs to nor a holder of the system-admin or hr-admin role"
           },
           "404": {
             "content": {
@@ -66801,7 +66801,7 @@
                 }
               }
             },
-            "description": "Caller is not the employee identified by the id"
+            "description": "Caller is neither the employee the request belongs to nor a holder of the system-admin or hr-admin role"
           },
           "404": {
             "content": {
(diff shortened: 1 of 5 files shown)


abhijithanandan:
@coderabbitai review

PR #591, opened 2026-08-30
Accept the purchase-order edits the web screens actually send

abhijithanandan:
The backend half of the two purchase-order PATCH outages on echno-core#57. The client half is a separate PR on echno-core, and it depends on this one being merged first, because `etc/backend-request-fields.json` is derived from `docs/openapi.json` at `development`.

## The outage this pairs with

echno-core sends `PATCH /purchase-orders/web/{id}` and `PATCH /purchase-order-items/{id}`. Verified against the running backend's own `/v3/api-docs` pulled off `.116`: `/purchase-orders/web/{id}` serves `GET` and nothing else, and `/purchase-order-items/{id}` serves `GET` and `DELETE`. Both real routes are on the collection, with the id in the body. Editing a purchase order's header, its remarks, or any line item has been a 404 behind a "Failed to update" toast for as long as those routes have existed, and nobody flagged it.

Nothing here is needed to end the 404 itself, which is a client-side path fix. What this PR is for is the second half: correcting the paths makes the payloads reach a DTO for the first time, and two of the controls on those screens turn out to be sending fields no DTO has.

## What changes

**`PurchaseOrderUpdateDto.projectId`.** The header card renders a project select in edit mode and posts it. `PurchaseOrder.project` is a real column, `nullable = false`, and `createPurchaseOrder` resolves it; the update DTO simply had no such field. The service now resolves it through `projectRepository.findByIdAndOrganization_Id`, the same tenant-scoped lookup create uses, so an id belonging to another organization is a 404 rather than a reallocation across a tenant boundary.

**`PurchaseOrderItemUpdateDto.materialId`.** The Edit Item row renders a material select and posts it. Same shape: real column on the line, set on create, absent from the update DTO. Accepted now, with one restriction. Once a goods received note has posted against a line, the line is the record of what arrived, so the material may only change while `receivedQuantity` is still zero; past that it is a 400 telling the caller to cancel the line and raise a new one. Sending the material the line already has is always fine, because the web client sends it on every save including saves that only touch the remarks, and rejecting an unchanged value would turn every edit of a part-delivered line into a 400.

**`totalAmount` stays, and stays ignored.** `updatePurchaseOrder` has never applied it, and that is right: the order total is the sum of its lines and `recalculateTotalAmount` runs on every line create, update and delete, so a value written here would survive until the next line edit. It is now documented as ignored on the schema and in the javadoc rather than reading as settable, and echno-core stops sending it. Removing it outright would break the published core that still does, so that is a later step once the web app is off it.

## Tests

`PurchaseOrderUpdateFieldsTest`, seven cases: the project moves, a project from another organization is refused with nothing saved, the total is left to the line items, the material changes on an untouched line, the material is refused on a line that has already received stock, an unchanged material is accepted after a receipt, and a received line stays editable in every other way with its line total and the order total recomputed. Disabling just the two new service branches fails four of them.

Full suite green on `.116`; `docs/openapi.json` regenerated.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -13784,11 +13784,17 @@
         "type": "object"
       },
       "PurchaseOrderItemUpdateDto": {
-        "description": "Payload to update a purchase order line item's quantity, price or remarks.",
+        "description": "Payload to update a purchase order line item's material, quantity, price or remarks.",
         "properties": {
           "id": {
             "description": "Id of the item to update.",
             "example": 512,
+            "format": "int64",
+            "type": "integer"
+          },
+          "materialId": {
+            "description": "Change the material the line is for. Only while nothing has been received against it.",
+            "example": 88,
             "format": "int64",
             "type": "integer"
           },
@@ -13815,7 +13821,7 @@
         "type": "object"
       },
       "PurchaseOrderUpdateDto": {
-        "description": "Payload to update a purchase order's status, delivery date, remarks or total.",
+        "description": "Payload to update a purchase order's status, project, delivery date or remarks.",
         "properties": {
           "expectedDeliveryDate": {
             "description": "Updated expected delivery date.",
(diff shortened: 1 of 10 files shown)


abhijithanandan:
@coderabbitai review

PR #592, opened 2026-08-30
Make the self-approval rule work when the raiser has no employee record

### LINKED ISSUE #584: Self-approval rule is inert when the caller has no employee record in the tenant
Found in the post-merge review of #567.

## The gap

`AttendanceRegularizationService.resolveCurrentActor()` returns `employeeId = null` for a caller with no Employee row in the current organization, and `SelfApprovalPolicy.checkSelfApproval` short-circuits before any comparison:

```java
if (raisedBy == null || approver == null || !raisedBy.equals(approver)) {
    return false;
}
```

So a user holding the org-member group plus `hr-admin` or `project-manager` whose Employee row does not yet exist (Keycloak groups assigned before HR created the record) can submit a regularization and then approve it themselves:

- `requestedById` is stored null, `requestedBy` carries their username.
- At approval both ids are null, the policy returns false, and the segregation-of-duties refusal never fires for a non-system-admin.
- Even for a system-admin, the break-glass warn log and the "(self-approved: ...)" note on the corrected clock events are skipped, so the row is indistinguishable from a legacy pre-#567 row.

The policy javadoc claims a null raiser exists "only for rows written before the raiser was stamped from the session". That is no longer true: every request raised by an employee-less member carries one going forward.

## Why this is not fixed inline

The clean fix needs an identity that exists for every caller. `requestedBy` stores the employee display name when there is an employee and the username when there is not, and the policy is shared with the stock-adjustment path, which passes user ids. Options, roughly in order of preference:

1. Stamp the raiser's **user id** on the request alongside `requestedById`, and have the attendance approve path fall back to comparing user ids when either employee id is null. Exact, collision-free, needs a column.
2. Fall back to comparing the stored `requestedBy` string with the current actor's name only when **both** employee ids are null, in which case both strings are usernames. No schema change; misses the raised-before-employee-created, approved-after case.

Either way the fallback should route through `checkSelfApproval` semantics (refusal for ordinary approvers, recorded break-glass for system-admin) and the javadoc claim should be corrected.

## Severity

Medium. The mechanism is certain, but the account shape is narrow (decision role held while the Employee row is missing), and both fields carry the same username so a human audit can still spot the pair.

abhijithanandan:
Closes #584.

## The gap

`SelfApprovalPolicy.checkSelfApproval` short-circuited on either identity being null, and `AttendanceRegularizationService.resolveCurrentActor()` returns `employeeId = null` for a caller with no Employee row in the current organization. A member holding the org-member group plus `hr-admin` or `project-manager`, whose Employee row does not exist yet because the Keycloak groups were assigned before HR created it, could raise a regularization and approve it themselves with the rule never firing. Even for a system-admin the break-glass warn log and the "(self-approved)" note on the corrected clock events were skipped, so the row looked like a legacy one.

## What changed

**One identity convention across the three callers.** `ApprovalParty(userId, employeeId)` replaces the two bare `Long`s. The finance modules were passing user ids and attendance was passing employee ids into the same two parameters, which is how a comparison could silently be between two different things. A comparison now succeeds on whichever identity the two sides share; both are unique per person, so a match on either is conclusive.

Comparing the stored display names instead (option 2 on the issue) was rejected. `requestedBy` holds an employee name where there is an employee and a username where there is not, two people can share a name, and a person can change theirs. A segregation-of-duties control should not resolve identity by string match.

**Regularizations stamp the raiser's user id** (`requested_by_user_id`, changelog 071), because that is the identity every authenticated caller has. Rows written before this keep a null and are still compared on the employee id.

**The null short-circuit was itself failing open, so it is now three outcomes rather than one silent pass:**

| Case | Before | Now |
|---|---|---|
| Comparable, different people | allowed | allowed, unchanged |
| Comparable, same person | refused, or break-glass | unchanged |
| Approver names nobody at all | allowed silently | **refused** |
| Raiser shares no identity with the approver | allowed silently | allowed, logged at WARN |

The approver refusal is the one place this fails closed, and it is deliberately narrow. It is not a comparison that cannot be made, it is an entry about to be posted by nobody: the approver id stored on the document would be null too, so there would be no accountability on the posting at all. Every path that reaches the policy is a role-gated approve endpoint on an authenticated session.

The raiser side stays open on purpose. Refusing there would strand every document raised before its raiser was stamped, with no second route to approval, which is the same dead end the break-glass exception exists to avoid. It is logged instead, so an approval whose check could not be made is visible rather than indistinguishable from one that passed.

**Fallout worth naming:** the refusal caught `ConstructionInvoicePostingIT` approving invoices with no session identity at all, which means the self-approval rule had never actually run in that suite. It now submits and approves as two people, the way production does.

## No API change

No controller or DTO touched; `docs/openapi.json` verifies unchanged (`./gradlew openApiSnapshot` passes). Nothing needed in `echno-core`.

## Proof

`AttendanceRegularizationSelfApprovalTest` runs the real policy end to end. Against the pre-fix comparison these fail:

| Test | Pre-fix |
|---|---|
| `anEmployeeLessMemberCannotRaiseAndThenApproveTheirOwnRequest` | FAILED, the approval went through |
| `anEmployeeLessSystemAdministratorSelfApprovesAndTheClockEventsSaySo` | FAILED, no "(self-approved)" note |
| `anApproverThatResolvesToNoIdentityAtAllIsRefused` | FAILED, waved through |
| `SelfApprovalPolicyTest.oneSharedIdentityIsEnoughToRecogniseTheSamePersonAcrossTheTwoSpaces` | FAILED |
| `SelfApprovalPolicyTest.anApproverWhoResolvesToNobodyIsRefusedRatherThanWavedThrough` | FAILED |
| `SelfApprovalPolicyTest.anUnidentifiedApproverIsRefusedEvenWhenTheRaiserIsUnknownToo` | FAILED |

Negative controls pass either way: an employee-less approver deciding somebody else's request, an employee approving their own, and a legacy request staying approvable.

Full suite green on `.116`.

## The pending stock adjustment on staging

SA-563 (document id 2) is **not affected**. It carries `submitted_by = 34` (`qa-stock-adjustment-raiser@echno.in`, a real user id), and the approver resolves to a real user id as well, so the comparison runs, finds two different people, and the approval proceeds exactly as it does today. The account has no employee row, which is the shape this PR is about, but on the stock path the user id was always being compared, so nothing there changes.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/attendance/AttendanceRegularization.java
+++ b/src/main/java/org/tornotron/echno_backend/attendance/AttendanceRegularization.java
@@ -45,6 +45,17 @@ public class AttendanceRegularization implements TenantScopedEntity {
     @Column(name = "requested_by_id")
     private Long requestedById;
 
+    /**
+     * The platform user id of whoever raised the request.
+     *
+     * <p>Stamped alongside {@code requestedById} because that one is an employee id, and a caller
+     * can hold a role in the tenant without having an employee record in it yet, which left the
+     * request with no id at all and the self-approval rule with nothing to compare. Every
+     * authenticated caller has a user id, so this is the identity the rule can always fall back to.
+     */
+    @Column(name = "requested_by_user_id")
+    private Long requestedByUserId;
+
     @Column(name = "requested_at")
     private LocalDateTime requestedAt;
 
--- a/src/main/java/org/tornotron/echno_backend/attendance/service/AttendanceRegularizationService.java
+++ b/src/main/java/org/tornotron/echno_backend/attendance/service/AttendanceRegularizationService.java
@@ -8,6 +8,7 @@ import org.tornotron.echno_backend.attendance.dto.*;
 import org.tornotron.echno_backend.attendance.enums.ClockEventType;
 import org.tornotron.echno_backend.attendance.enums.RegularizationStatus;
 import org.tornotron.echno_backend.attendance.mapper.AttendanceRegularizationMapper;
+import org.tornotron.echno_backend.common.approval.ApprovalParty;
 import org.tornotron.echno_backend.common.approval.SelfApprovalPolicy;
 import org.tornotron.echno_backend.common.exception.ResourceNotFoundException;
 import org.tornotron.echno_backend.common.multitenancy.TenantContext;
@@ -76,9 +77,14 @@ public class AttendanceRegularizationService {
         this.selfApprovalPolicy = selfApprovalPolicy;
     }
 
-    /** Who the authenticated caller is, as recorded on a request: a display name and, when the
-     *  caller has an employee record in this tenant, that employee's id. */
-    private record Actor(String name, Long employeeId) {
+    /** Who the authenticated caller is, as recorded on a request: a display name, their platform
+     *  user id, and, when the caller has an employee record in this tenant, that employee's id. */
+    private record Actor(String name, Long userId, Long employeeId) {
+
+        /** The same identity as the self-approval rule compares parties by. */
+        ApprovalParty party() {
+            return new ApprovalParty(userId, employeeId);
+        }
     }
 
     /**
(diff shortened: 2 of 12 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #593, opened 2026-08-30
Let an organization be edited without also replacing its logo

abhijithanandan:
Two defects found beside the OpenAPI contract check (echno-core#57), neither of them on its list of 66.

## The crash

`OrganizationWebController.partialUpdateAnOrganization` declares

```java
@RequestParam(value = "attachments", required = false) List<MultipartFile> attachments
```

and `OrganizationService` iterated it straight away, with no null guard. Spring resolves an absent multipart parameter to null, not to an empty list, so every organization edit that did not also upload a file went through a loop over null. That is nearly all of them: the form has name, address, email, phone and website alongside the logo, and changing any of those without touching the logo took this path. `IssueService` guards exactly this shape correctly, which is what makes the omission look like a copy that lost a line rather than a decision.

This was reasoned rather than executed on echno-core#57, because the staging admin lacks the `system-admin` / `hr-admin` role the endpoint wants and granting one to make a test possible is not a thing to do. It is now settled by test instead, which is the better proof anyway.

**Two tests, because either alone proves nothing.** `OrganizationPatchAttachmentsArgumentTest` is a web slice with the service mocked: it PATCHes a multipart request carrying only the `data` part and captures what the handler was given, asserting it is null. That is the half you cannot read off the service. `OrganizationPatchWithoutAttachmentsTest` builds the service and calls it with null, which fails with a NullPointerException on the unguarded code and passes with the guard. Together they say the crash was live on every logo-less edit.

Worth noting for echno-core#57: this puts the organization `isActive` finding at #5 on that list out of reach entirely, not merely dropped. That PATCH could not complete at all unless a file went with it.

## The typo

`IndentUpdateDto.createdByemployeeId` has a lowercase `e`, against `createdByEmployeeId` on `IndentCreationDto`. `IndentService.updateIndent` reads the misspelled getter, so a caller sending the name the create endpoint uses would have it silently dropped: the indent's creator would not change and the response would say the update succeeded.

Nothing sends either name today. `UpdateIndentRequest` in echno-core has no such field, and echno-web sets `createdByEmployeeId` only on the create page. That is why this has gone unnoticed, and why renaming it costs nothing now: no deployed client is relying on the misspelling, so there is no migration to sequence. Left alone it becomes a live bug the first time somebody adds the field to the client by copying the name from the create DTO, which is exactly what they would do.

Full suite green on `.116`; `docs/openapi.json` regenerated for the renamed property.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -6083,7 +6083,7 @@
       "IndentUpdateDto": {
         "description": "Payload to partially update an indent's number, project, status, expected date or remarks.",
         "properties": {
-          "createdByemployeeId": {
+          "createdByEmployeeId": {
             "description": "Id of the employee who raised the indent.",
             "example": 8,
             "format": "int64",
--- a/src/main/java/org/tornotron/echno_backend/indent/IndentService.java
+++ b/src/main/java/org/tornotron/echno_backend/indent/IndentService.java
@@ -107,9 +107,9 @@ public class IndentService {
             indent.setIndentNumber(indentDto.getIndentNumber());
         }
 
-        if(indentDto.getCreatedByemployeeId() != null) {
-            Employee employee = employeeRepository.findByIdAndOrganizationId(indentDto.getCreatedByemployeeId(), TenantContext.getCurrentOrgId())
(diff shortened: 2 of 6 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #594, opened 2026-08-30
Drop the employee reporting_manager column nothing has ever written

### LINKED ISSUE #587: employee.reporting_manager is an unmapped column that is null on every row

abhijithanandan:
Closes #587.

## What this removes

`employee.reporting_manager`, a `VARCHAR(255)` that has been in the schema since `v1.0/009-create-employee.xml` and was carried through the `v2.0`, `v3.0` and `v4.0` baseline squashes.

## The claim, re-checked rather than inherited

The issue says no Java code maps it. That holds, and it holds a little more widely than the issue states:

- `Employee` has no `reportingManager` field. The reporting line is `manager` / `@JoinColumn(name = "manager_id")`.
- Case-insensitive grep for `reporting_manager` / `reportingManager` over the whole backend repo hits **only** changelog XML: `v1.0/009`, `v1.2/052`, and the three baselines. Nothing in `src/`.
- The same grep over `echno-core` and `echno-web` returns nothing either, so the column has never reached a DTO or the client.

`v1.2/052-refactor-reporting-manager.xml` is the reason it looks like a duplicate: it added `manager_id`, the FK, and the index, and left the string column it replaced sitting beside them. `EmployeeHierarchyService.validateManager` (same organization, not self, no cycle) and `findByManager_Id` are what carry the relationship now; the measurement on the issue was 0 of 30 rows with `reporting_manager` set against 4 with `manager_id`.

## What is here

- `v4.0/072-drop-employee-reporting-manager.xml`, guarded on `columnExists` so re-running against a database that already dropped it marks-ran instead of failing, plus the master include.
- `EmployeeSchemaIT`, a Liquibase-built-schema assertion against the real CockroachDB container: `employee` has `manager_id` and does not have `reporting_manager`.

Nothing else changes: no entity field, DTO field or query to remove with it, no data to migrate, no API surface, so `docs/openapi.json` is untouched.

## Verified

On the lab box, against the real container:

- With the changeset: `EmployeeSchemaIT` passes.
- With the master include removed: it fails on `not to contain ["reporting_manager"] but found ["reporting_manager"]`.

The negative run also confirms `is_manager` is already gone from the built schema, so this sits cleanly on top of #520.

### DIFF
--- a/src/main/resources/db/changelog/db.changelog-master.xml
+++ b/src/main/resources/db/changelog/db.changelog-master.xml
@@ -352,4 +352,7 @@
     <!-- v4.0 - Regularizations: the raiser's user id, so the self-approval rule always has one -->
     <include file="db/changelog/v4.0/072-add-regularization-requested-by-user-id.xml"/>
 
+    <!-- v4.0 - Employees: drop the free-text reporting_manager; manager_id is the reporting line -->
+    <include file="db/changelog/v4.0/074-drop-employee-reporting-manager.xml"/>
+
 </databaseChangeLog>
--- a/src/main/resources/db/changelog/v4.0/074-drop-employee-reporting-manager.xml
+++ b/src/main/resources/db/changelog/v4.0/074-drop-employee-reporting-manager.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <changeSet id="074-01-drop-employee-reporting-manager" author="echno">
+        <preConditions onFail="MARK_RAN">
+            <columnExists tableName="employee" columnName="reporting_manager"/>
+        </preConditions>
+
+        <comment>
+            Removes the employee.reporting_manager string.
+
+            The reporting line is manager_id, the self-referencing foreign key behind
+            Employee.manager: it is validated by EmployeeHierarchyService.validateManager
+            (same organization, not self, no cycle) and read by findByManager_Id for the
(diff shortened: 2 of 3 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
Renumbered the changeset from 072 to 074 on rebase: another branch landed `072-add-regularization-requested-by-user-id.xml` on development in the meantime.

PR #595, opened 2026-08-30
Walk the two hops the public-endpoint guard could not see

### LINKED ISSUE #585: PublicEndpointTenantExposureTest does not walk event listeners or EntityManager queries
Found in the post-merge review of #566. The gate works for what it walks: BFS over `getMethodCallsFromSelf()` from every `permitAll()` handler, repository verdict from the repository's generic declaration, a fail-toward-reporting fallback on the return type, method-level `permitAll` overriding class level. The one permitAll endpoint today (`AuthController` register) is judged correctly. But the guarantee stated next to it in `SecurityConfig` is wider than the graph it actually covers, and a gate with quiet holes buys false confidence.

## Hole 1: the event-publish hop is invisible

The walk follows direct calls into `org.tornotron.echno_backend.*` only. `ApplicationEventPublisher.publishEvent` is a Spring-class leaf, so it is dropped and the listener behind it is never walked. This codebase routes real tenant reads and writes through that hop (`GoodsReceivedNoteService`, `SiteTransferService`, `MaterialConsumptionService`, `ChatService` all publish events).

Concrete miss: a future permitAll handler publishes an event handled by a synchronous `@EventListener` that reads a tenant-scoped repository. The read runs on the request thread, unauthenticated, across every organization, and the build stays green.

## Hole 2: only Spring Data Repository sinks count

`EntityManager` JPQL and native queries are invisible (not ours, not a `Repository`, so a leaf). Three production classes query tenant data this way today: `finance/report/service/ReportService`, `vendor/VendorSummaryService`, `common/documentnumber/DocumentNumberAllocator`. A permitAll endpoint reaching, say, `ReportService` trips nothing in this gate. The runtime load-boundary guard is a separate mechanism and does not make the build-time gate honest about this.

## Minor

- A compound grant (`"permitAll() or hasRole(...)"`) or `isAnonymous()` is treated as protected, under-counting public endpoints.
- Unresolvable call targets (`resolveMember().isEmpty()`) are skipped silently; harmless today because every service is a concrete class, but interface-typed injection would open a quiet gap.

## Suggested follow-up

Either widen the walk (treat `publishEvent` as an edge to every `@EventListener`/`@TransactionalEventListener` whose event type matches, and treat `EntityManager` query methods as sinks), or state the two exclusions in the test's javadoc so the next reader does not take the green as covering them.

abhijithanandan:
Closes #585.

`PublicEndpointTenantExposureTest` worked for what it walked, and the guarantee written beside it in `SecurityConfig` was wider than the graph. Nothing is exploitable today, so this is about the gate being trustworthy rather than about a live exposure. That distinction shaped the fix: both holes turned out to be traversable, so both are traversed, and the one limit that is genuinely intractable statically is written on the test instead of left as a silent gap.

## Hole 1: the event-publish hop

`ApplicationEventPublisher.publishEvent` is a Spring method, so the old walk dropped it at the framework boundary and never reached the listener. Four services move real tenant work across that hop.

The hop is followed by matching the event the publishing method **constructs** against the parameter type of every `@EventListener` and `@TransactionalEventListener`. The call itself is erased to `publishEvent(Object)`, so the constructor call at the publish site is the only thing the bytecode offers, and it is how every publication in this codebase is written. Where no event type can be identified there, **every** listener is followed instead, so a publication assembled elsewhere over-reports rather than disappears.

`AFTER_COMMIT` is not an escape: that listener runs on the request thread inside the same unscoped declaration. An **`@Async`** listener is deliberately not followed, for the same reason `/actuator` is not covered: the thread hand-off leaves the declaration behind, so the load boundary is back on and the read is refused rather than let through. Reporting it would be a false alarm, and false alarms are how a ratchet gets switched off. The two cases are told apart properly rather than by the fallback: listeners are indexed asynchronous ones included, so an event whose only listener is `@Async` is known to have no edge, which is not the same as an event whose type could not be worked out.

## Hole 2: reads that are not Spring Data repositories

`EntityManager` query methods and `JdbcTemplate` statements now count as reads in their own right. Note the issue named `EntityManager` for `DocumentNumberAllocator`; it is actually `JdbcTemplate`, so both are covered.

Which table such a statement touches is a string in the source and not a fact in the bytecode, so unlike a repository these are reported **whatever** they read. That is stated in the class rather than quietly assumed, and it is the direction to fail in: an unauthenticated request that has reached hand-written SQL is worth the argument, and the answer is to key the query on something the caller proved, not to widen the rule. Holding an `EntityManager` is not the same as querying through one, so only the methods that put a statement on the wire count.

## The two minors

- A compound grant (`"permitAll() or hasRole('ADMIN')"`) or `isAnonymous()` read as protected, which shrank what the rule looked at without anyone noticing. Anything mentioning `permitAll` or `isAnonymous` counts now.
- A call typed as an interface resolves to the interface's own declaration, which has no body, so the walk stopped one hop short of the class doing the reading. Every implementation in the imported graph is followed as well.

## The limit that stays, documented

Reachability is over static calls, so anything dispatched by a value rather than by a type is invisible: a lambda handed across a boundary, reflection, a proxy resolved by name. Nothing on a public path does that today, and seeing it would mean turning this into a data-flow analysis. That is now written on the test and referenced from the `SecurityConfig` comment, which is also corrected to say what the gate actually covers.

## Proof the walk is not vacuous

The rule finds nothing in production, which is the answer everyone wants and also the answer a broken rule gives. So `PublicEndpointTenantExposureRuleTest` drives the walk directly over `PublicEndpointReachabilityFixtures`: a planted violation for each hop, shaped like the production code that uses it, each paired with the case it must **not** report (the `@Async` listener, an `EntityManager` call that issues no statement, work that reaches no database).

With the four additions reverted, exactly the five tests for them fail and the negative cases still pass:

```
followsAPublishedEventIntoItsListener()            FAILED
countsAnEntityManagerQueryAsAReadOfTenantData()    FAILED
countsAJdbcTemplateStatementAsAReadOfTenantData()  FAILED
descendsFromAnInterfaceIntoItsImplementation()     FAILED
treatsACompoundOrAnonymousGrantAsPublic()          FAILED
```

The fixtures carry no `@RestController`, no `@Component` and no repository interface of their own, because they sit under the application's package on the test classpath and anything scannable there would be picked up by every `@SpringBootTest` in the suite. The imported graph is held in instance fields rather than a `static final`, so it is collectable as soon as the test using it is done.

Full `./gradlew build` green on the lab box; test heap 25% of the 2048m cap, 8 Spring contexts of a maximum 8 (unchanged, no new context). No production behaviour changed: the only main-source edit is the `SecurityConfig` comment, so `docs/openapi.json` is untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved architectural security checks for unauthenticated endpoints that may access tenant-scoped data.
  * Added coverage for database queries through persistence APIs, JDBC, interfaces, and synchronous event listeners.
  * Improved detection of public access rules, including compound and anonymous-access expressions.
  * Excluded asynchronous event listeners and non-database operations from exposure reports.

* **Tests**
  * Added comprehensive reachability scenarios and clearer reported paths for security violations.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/common/configuration/SecurityConfig.java
+++ b/src/main/java/org/tornotron/echno_backend/common/configuration/SecurityConfig.java
@@ -70,8 +70,14 @@ public class SecurityConfig {
                         // touches tenant-scoped data, it does not belong on this list; scope the
                         // read on something the caller proved and put it behind an authenticated
                         // path instead. PublicEndpointTenantExposureTest walks the handler chain
-                        // of every @PreAuthorize("permitAll()") endpoint and fails the build on
-                        // one that reaches a tenant-scoped repository.
+                        // of every endpoint whose @PreAuthorize lets an unauthenticated caller
+                        // through, and fails the build on one that reaches a tenant-scoped
+                        // repository, an EntityManager query or a JdbcTemplate statement,
+                        // whether it reaches it by a direct call, through an interface, or
+                        // across a published event into a synchronous listener. What it cannot
+                        // see is a call dispatched by a value rather than by a type: a lambda
+                        // handed across a boundary, reflection, a proxy resolved by name. That
+                        // limit is stated on the test itself.
                         //
                         // /actuator is the exception, and only because TenantFilter skips it
                         // outright rather than declaring anything: an entity load from there is
--- a/src/test/java/org/tornotron/echno_backend/architecture/PublicEndpointTenantExposureRuleTest.java
+++ b/src/test/java/org/tornotron/echno_backend/architecture/PublicEndpointTenantExposureRuleTest.java
@@ -74,6 +74,19 @@ class PublicEndpointTenantExposureRuleTest {
                 .isEmpty();
     }
 
+    /**
+     * And the same for a listener that is asynchronous by its class rather than by its method.
+     * Spring applies a type-level {@code @Async} to every method the type declares, so reading
+     * only the method annotation would put this listener back into the reported path.
+     */
+    @Test
+    void doesNotFollowAListenerMadeAsynchronousByItsClass() {
+        assertThat(pathFrom(
+                PublicEndpointReachabilityFixtures.PublishesAClassLevelAsyncEvent.class, "handle"))
+                .as("a type-level @Async is as good as one on the method")
+                .isEmpty();
+    }
+
     // -------------------------------------------------------------------------------
     // Hole 2: reads that are not Spring Data repositories
     // -------------------------------------------------------------------------------
--- a/src/test/java/org/tornotron/echno_backend/architecture/PublicEndpointTenantExposureRuleTest.java
+++ b/src/test/java/org/tornotron/echno_backend/architecture/PublicEndpointTenantExposureRuleTest.java
@@ -1,0 +1,159 @@
+package org.tornotron.echno_backend.architecture;
+
+import com.tngtech.archunit.core.domain.JavaClasses;
+import com.tngtech.archunit.core.domain.JavaMethod;
+import com.tngtech.archunit.core.importer.ClassFileImporter;
+import org.junit.jupiter.api.Test;
+import org.tornotron.echno_backend.architecture.fixtures.PublicEndpointReachabilityFixtures;
+
+import java.util.List;
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for the gate, not for the code the gate guards.
+ *
+ * <p>{@link PublicEndpointTenantExposureTest} finds nothing in the production code, which is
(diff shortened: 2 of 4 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Fixed in 9e885a2. Spring does apply a type-level `@Async` to every method the type declares, so a listener made asynchronous that way was being followed into and would have been reported as a path the load boundary already refuses. `runsOnTheRequestThread` now reads the owner as well as the method.

The fixture you asked for is there too (`ReadsTenantRowsOnAnAsyncClass`, reached through `PublishesAClassLevelAsyncEvent`), and it bites: with the owner check reverted, `doesNotFollowAListenerMadeAsynchronousByItsClass` is the one test of the thirteen that fails.

Issue #549, opened 2026-08-29
Project 27 on staging is approved with zero compliance inspections, and there is no audit trail to say how


abhijithanandan:
## What is wrong

Project id 27, "Demo Project", is sitting on staging in status `approved` with no compliance inspections against it, and no way to tell how it got there. Approval is meant to be the trigger that generates compliance inspections, so an approved project with none is a state the system should not be able to produce.

Found while reviewing #544. This is a decision I need, not a defect report: I am not going to touch a row on staging without agreement.

## What I verified (29 Aug 2026, read-only against the compose staging on `.116`)

```
  id | project_name |  status  | project_state | project_address | project_type |        created_at
-----+--------------+----------+---------------+-----------------+--------------+---------------------------
  27 | Demo Project | approved | NULL          | Chennai         | COMMERCIAL   | 2026-08-25 10:40:43.37743
```

- It has zero inspections of any kind, compliance or otherwise (`SELECT type, origin, count(*) FROM inspections WHERE project_id = 27 GROUP BY type, origin` returns no rows).
- It is the only project on staging in status `approved`. The rest are 20 `upcoming`, 5 `open`, 1 `completed`, 1 `onHold`.
- There are 21 active rows in `compliance_rules`, so the catalogue is not empty.

**Whether it was created approved or patched to approved cannot be determined.** The `project` table has a `created_at` and no `updated_at`, no `created_by` and no `updated_by`, and there is no audit, revision or history table anywhere in the schema (checked `information_schema.tables`). There is nothing left to read.

It also could not have generated compliance by either route. `ComplianceGenerationService` resolves the jurisdiction through `IndianStateResolver.forProject(project.getProjectState(), project.getProjectAddress())` and aborts when that comes back null. `project_state` is NULL and the address is the bare string "Chennai", which the resolver deliberately does not treat as naming a state (its own comment says an address of "Chennai" names no state at all). So even a correctly fired approval would have stopped before creating anything.

## Why it matters

Two separate things, and they should be answered separately.

The row itself is small: one demo project on staging, easy to leave alone or clean up. What it exposes is not small. `project` carries no change timestamp and no audit trail, so for any project on any environment we cannot say when its status last changed or who changed it. The moment compliance generation is a real feature with Anand's rule catalogue behind it, "this project is approved and has no compliances" becomes a question someone will ask about a client's data, and today there is no way to answer it.

## What I need decided

[...]

PR #597, opened 2026-08-30
Record who moved a project's status, and when

### LINKED ISSUE #565: A project records no change history, so no status transition can ever be explained
## What is missing

The `project` table records `created_at` and nothing else about its own history. No `updated_at`, no `created_by`, no `updated_by`, and there is no audit, revision or history table anywhere in the schema.

The consequence is not abstract. A project's status is the field with real behaviour behind it: approval is what draws up the project's compliance inspections. For any project, on any environment, we cannot say when its status last changed or who changed it. When a client asks why their approved project has no compliance work, or asks who approved it, there is nothing to read.

## How it surfaced

Split out of #549, where project 27 on staging was found `approved` with zero compliance inspections. Whether it was created approved or patched there could not be determined, and the reason was this gap. Both routes into that state are now closed (`5073609` refuses a project created already approved, `5a16111` refuses an approval where no state resolves), so the specific bug cannot recur. The inability to diagnose it can, on the next thing.

## What I think we want

Smallest useful step first, rather than a general audit framework:

1. `updated_at` and `updated_by` on `project`, populated on every write. This alone answers "when did this last change, and by whom".
2. Status transitions specifically are worth more than a last-write stamp, because they are what carries behaviour. A small `project_status_history` row per transition (from, to, who, when) would answer the question #549 could not, and would cost one insert on a path that already exists in `ProjectService.onApproval`.

Worth deciding whether other entities with a gated transition want the same treatment. `PurchaseOrderService.createPurchaseOrder` applies the same create-time rule, and `StockAdjustmentService.approve` already records `submittedBy`, `approvedBy` and `processedBy` on the document, which is the shape to copy.

## Related

#549 (where it was found), #553.

abhijithanandan:
Closes #565. The design was posted on the issue first: https://github.com/tornotron/echno-backend/issues/565#issuecomment-5469636996

## The gap, which is two gaps

A project recorded `created_at` and nothing else about its own history. No `updated_at`, no `created_by`, no `updated_by`, and no audit, revision or history table anywhere in the schema. Two different questions were unanswerable, and they need different answers:

1. **"When did this last change, and who touched it?"** A last-write stamp.
2. **"Who approved this, when, and was it created in that state or patched into it?"** A transition trail.

Neither answers the other. A stamp is overwritten by the next edit, so the approver's name survives only until somebody fixes a typo in the address. A trail says nothing about a project whose status has never moved.

## Scope: one shared table, per-module exposure

Seven records here carry a status enum with a gated transition: Project, StockAdjustment, ConstructionInvoice, AttendanceRegularization, SiteTransfer, Issue, PurchaseOrder. Three already record the actor on the document itself and all three do it differently (`Long` user ids on StockAdjustment, String usernames inherited from the finance `BaseEntity` on ConstructionInvoice, both at once on AttendanceRegularization). Four record nothing. A `project_status_history` with a project foreign key would be followed by a `site_transfer_status_history`, and by the sixth there would be six tables of identical shape drifting the way those actor columns already have.

So: **the storage and the writer are shared, the API is not.** One `status_transition` table keyed by `(entity_type, entity_id)` the way `attachment` already is, one `StatusTransitionRecorder`, and each module reads its own trail through its own endpoint under its own `@PreAuthorize`. A generic `GET /history/{type}/{id}` would need a registry of per-type read rules, because who may read a site transfer's history is not who may read a project's, and that registry is the half-general framework worth avoiding. Project is the first and so far only caller; the second is one call at its transition point plus one controller method.

## Mechanism: why not Envers

Envers would give the ability to ask what any field held on any past date. Nobody has asked for that. What it costs is not mainly the `_AUD` table per entity, it is tenant isolation: isolation here is fail-closed by construction, with `TenantIsolationLoadListener` registered against the `TenantScopedEntity` interface and the `orgFilter` switched on per transaction by `HibernateFilterConfig`. Envers reads through `AuditReader`, which queries the `_AUD` tables directly with the Hibernate filter not applied, and `REVINFO` is one global revision table with no organization column at all. Adopting it means a second, parallel tenant-scoping story for the audit read path, remembered on every query. A plain entity gets isolation by implementing an interface.

An event log solves delivery and replay rather than history, and would record only approval, since approval is the only project transition that publishes an event.

## What is recorded

Status transitions, plus a last-write stamp for everything else. Status is the only project field with behaviour behind it: `onApproval` publishes `ProjectApprovedEvent`, which draws up the compliance inspections. A change to the postal code carries nothing, and recording it costs a row per edit forever to answer a question nobody asks.

## What is here

- `project.created_by`, `project.updated_at`, `project.updated_by`, stamped on create and on every patch, and surfaced on `ProjectDto`.
- `status_transition` plus `StatusTransition`, `StatusTransitionRepository`, `StatusTransitionRecorder`, its DTO and mapper.
- `ProjectService` records the created status on create and every status change on patch, after the approval gate so a refused approval writes nothing. The service never works out whether the status moved: it records unconditionally and the recorder drops an entry whose two ends are equal, so no caller can forget.
- `GET /api/v1/project/web/{id}/status-history`, paged newest first, `PageQuery`-bound, under the same authorization as the other project reads.
- `073-project-change-history.xml`: the three columns, the table with its organization FK and indexes, and the baseline seed.
- `docs/openapi.json` regenerated with `-PupdateOpenApiSnapshot`.

Append-only is a property of the type rather than a convention: the entity is `@Immutable` and the repository extends `Repository` rather than `JpaRepository`, so no delete surface exists at all. That is the `AssetMovementRepository` pattern.

Two deliberate departures from `asset_movement`:

- **No foreign key to the record it describes.** `asset_movement` refuses to let an asset be deleted while its ledger is non-empty, because a movement is a fact about a live asset. A status trail has to outlive the record: a cascade would erase the evidence of who approved a project the moment the project was deleted, and a restrict would make an approved project undeletable.
- **`note` is nullable**, where `asset_movement.reason` is required. An entry already names both of its ends; requiring prose on every one buys a column full of "changed status".

`changed_by_name` is snapshotted beside `changed_by` for the reason `asset_movement` snapshots its names: history that evaporates when a user is renamed is not history.

**The column that answers #549 is `source`.** `CREATION` against `UPDATE` is literally "created approved, or patched there".

## What this cannot answer

Stated plainly on the issue too, so nobody expects otherwise:

- For every project that exists today, the trail begins at the migration. Each gets one `BASELINE` entry naming the status it is observed to hold at that moment, with no actor and no earlier status.
- That entry is deliberately **not** dated at `created_at`. Dating the observed status at creation time would assert the project was created in that state, which is exactly the claim #549 could not check.
- **Project 27 stays unexplained.** Its baseline row says it was observed `approved` on the migration date, by nobody nameable. It will not say who approved it, when, or whether it was created that way. Both routes into that state are already closed (`5073609`, `5a16111`), so it cannot recur; the history is not recoverable.
- `created_by` is null on every pre-existing project, for the same reason.
- A change made directly against the database bypasses this, as it bypasses everything else.

## Tests

- `StatusTransitionRecorderIT`, against the real CockroachDB on the Liquibase-built schema (Hibernate runs `ddl-auto: validate`, so the context starts only if entity and table agree): the opening entry, the no-op that must not become an entry, the snapshotted actor, newest-first ordering, one record's entries not leaking into another's, and two organizations numbering a project the same not seeing each other's trail.
- `ProjectStatusHistoryTest`: the created status opens the trail, both ends and the actor are recorded on approval and on every other transition, a refused approval records nothing, a patch with no status change still stamps the write, and a write with no user context records the transition with no actor.
- The four existing project tests take the two new collaborators as mocks; nothing about their assertions changed.

Verified on the lab box: full `./gradlew test jacocoTestCoverageVerification` green (test heap 524 MB of 2048). Negative proof: with the `recordStatusChange` call removed, 5 of the 9 new tests fail on "Wanted but not invoked".

## Follow-ups, not in scope here

- Second callers, one at a time, each at its own transition point with its own endpoint. SiteTransfer and PurchaseOrder first, since they record nothing at all today.
- The `echno-core` type and the `echno-web` screen. The backend change is purely additive, so nothing breaks in the meantime.
- Whether the bespoke `approvedBy`/`submittedBy` columns keep being written once the shared trail exists. My view is yes for now: those modules' own screens read them, and collapsing them is a separate change with its own risk.

## Summary by CodeRabbit

- **New Features**
  - Added project status-history tracking for creations and status changes.
  - Added a paginated endpoint to view a project’s status history.
  - Project details now include creator and last-update information.
  - Existing projects receive an initial baseline history entry.

- **Bug Fixes**
  - Prevented history entries for unchanged project statuses.

- **Documentation**
  - Updated the API specification with status-history schemas and endpoint details.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -11968,6 +11968,52 @@
         },
         "type": "object"
       },
+      "PageStatusTransitionDto": {
+        "properties": {
+          "content": {
+            "items": {
+              "$ref": "#/components/schemas/StatusTransitionDto"
+            },
+            "type": "array"
+          },
+          "empty": {
+            "type": "boolean"
+          },
+          "first": {
+            "type": "boolean"
+          },
+          "last": {
+            "type": "boolean"
+          },
+          "number": {
+            "format": "int32",
+            "type": "integer"
+          },
+          "numberOfElements": {
+            "format": "int32",
+            "type": "integer"
+          },
+          "pageable": {
+            "$ref": "#/components/schemas/PageableObject"
+          },
+          "size": {
+            "format": "int32",
+            "type": "integer"
+          },
+          "sort": {
+            "$ref": "#/components/schemas/SortObject"
+          },
+          "totalElements": {
+            "format": "int64",
+            "type": "integer"
+          },
+          "totalPages": {
+            "format": "int32",
+            "type": "integer"
+          }
+        },
+        "type": "object"
+      },
       "PageStockAdjustmentDto": {
         "properties": {
           "content": {
@@ -13084,6 +13130,12 @@
             "format": "date-time",
             "type": "string"
           },
+          "createdBy": {
+            "description": "Id of the user who created the project. Null on projects created before this was recorded.",
+            "example": 17,
+            "format": "int64",
+            "type": "integer"
+          },
           "customerId": {
             "description": "Finance customer the project is billed to. Null when no client is set.",
             "example": "6b1e9c22-9f8a-4a1b-9c0e-1d2f3a4b5c6d",
@@ -13193,6 +13245,18 @@
               "$ref": "#/components/schemas/TaskDto"
             },
             "type": "array"
+          },
+          "updatedAt": {
+            "description": "When the project was last written. Null on projects not written since this was recorded. It is a last-write stamp, so it is overwritten by the next edit to any field: to read who moved the project's status and when, use GET /api/v1/project/web/{id}/status-history.",
+            "example": "2026-08-28T16:41:12",
+            "format": "date-time",
+            "type": "string"
+          },
+          "updatedBy": {
+            "description": "Id of the user who last wrote the project.",
+            "example": 17,
+            "format": "int64",
+            "type": "integer"
           }
         },
         "type": "object"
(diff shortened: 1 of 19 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
Fixed the team-change stamp finding in b7a5271: `addEmployeeToProject` and `removeEmployeeFromProject` now stamp before saving, with three tests covering both directions and the fact that a team change writes nothing to the status trail.

Issue #601, opened 2026-08-30
Invoice cancel endpoints accept a reason the reversal entry cannot store


abhijithanandan:
Both cancel endpoints take the reason as a bare request parameter with no size bound:

- `InvoiceControllerWeb.cancel`: `@RequestParam String reason`
- `ConstructionInvoiceControllerWeb.cancel`: `@RequestParam String reason`

On an issued invoice the reason ends up in the reversing entry's description: `JournalPostingService.reverse` writes `"Reversal of " + entryNumber + " - " + reason` into `journal_entry.description`, a 500-character column. `entry_number` is 30 characters, so the prefix is at most 45 and anything over 455 characters overflows the column. `ReverseJournalRequest` does carry `@NotBlank @Size(max = 500)`, but the services construct it with `new ReverseJournalRequest(reason)`, and bean validation only runs on controller-bound payloads, so neither annotation fires on this path. Even if it did, 500 is already too long once the prefix is on.

**Failure:** `POST /finance/invoices/web/{id}/cancel?reason=<456+ chars>` on an issued invoice fails as a database error (a 500), not a 400 naming the field. Same for the construction-invoice cancel. An empty `reason=` is also accepted, leaving a reversal entry whose description ends in a dangling separator.

The web UI caps the reason at 455 client-side (`CANCEL_REASON_MAX_LENGTH` in echno-web#345), but that is a workaround: any other API caller still hits the overflow.

**Fix:** put `@NotBlank @Size(max = 455)` on both `reason` parameters and `@Validated` on the two controllers so the refusal is a 400 before the ledger is touched. Keep the bound derived from the same arithmetic the web constant documents (500 minus the 45-character worst-case prefix), so the two cannot drift apart silently.

PR #610, opened 2026-08-30
Name the people a document's approval stamps point at

### LINKED ISSUE #596: A document's *By stamps are user ids with no way to resolve a name, so every screen shows a number
Found while fixing echno-web#344.

`submittedBy`, `approvedBy`, `rejectedBy`, `processedBy` on `StockAdjustmentDto`, and `submittedBy` / `approvedBy` / `paymentRecordedBy` on `ConstructionInvoiceDto`, are stamped from the session with `UserContextService.getCurrentUserId()`, so they hold a **user** id.

Nothing the client can call turns that into a name:

- `GET /user/web` returns the caller's own user and no other.
- `EmployeeLookupDto` is keyed by employee id and carries no user id, and `EmployeeDto` carries none either, so there is no column to join on.

The web app resolved these through the employee lookup, which is the wrong key: it missed for most ids and rendered `EMP-012`, and where an employee id happened to equal a user id it named a different person. A wrong name against an approval is a wrong record of who approved it, so echno-web#346 now renders `User #<id>` and nothing more.

That is honest but not useful. The cheapest fix is on this side: carry the display name alongside the id on the DTOs that already carry the stamp, the way `customerName` sits beside `customerId` on `InvoiceDto` and `managerName` beside `managerId` on `EmployeeDto`.

- `submittedByName`, `approvedByName`, `rejectedByName`, `processedByName` on `StockAdjustmentDto`
- `submittedByName`, `approvedByName`, `paymentRecordedByName` on `ConstructionInvoiceDto`

Denormalised on read from the `users` row, nullable when the user has since gone. An alternative would be a batch user-lookup endpoint, but a name per row is one join on a read the client already makes, and it needs no second request per screen.

Nothing is broken until this lands; the screens just show a number where a name belongs.

abhijithanandan:
Closes #596.

`submittedBy` / `approvedBy` / `rejectedBy` / `processedBy` on `StockAdjustmentDto` and `submittedBy` / `approvedBy` / `paymentRecordedBy` on `ConstructionInvoiceDto` are stamped from the session, so they hold a **user** id, and nothing the client could call turned one into a name. The web app resolved them through the employee lookup, which is keyed by employee id and carries no user id at all: it missed for most ids and, where an employee id happened to equal a user id, printed a different person's name against an approval. echno-web#346 and #349 replaced that with `User #<id>`, which is honest and useless.

## The shape, and why

**Name fields alongside the ids on the DTOs**, resolved by the service in one batched read and passed into the mapper as a MapStruct `@Context`.

- **Not a batch lookup endpoint.** It costs a second request per screen, and it is a new authorization decision on top of a display problem: `GET /user/web/all` is `system-admin` / `hr-admin` only, so a lookup endpoint either keeps that restriction, in which case an ordinary approver still sees a number, or opens a user-directory read to every role that can see a document. Each screen would also have to know which ids to ask for and cache the answers separately.
- **Not resolved in the mapper**, at least not the naive way. #522 item A settled that a mapper converts what it is handed and asks nobody anything, enforced by `architecture/MapperDatabaseAccessTest`: a lookup in the conversion path costs a round trip per stamp per row and the call site shows none of it.
- **So: batched in the service, handed down as a `@Context`**, which is the `MaterialMapper.toWithStockDto` shape the rule points at. `UserNameDirectory.namesFor` reads every stamped id on the page in one query; `UserNameLookup` is a plain map the mapper reads. `StockAdjustmentStampNameTest` and `ConstructionInvoiceStampNameTest` pin that a three-row page costs exactly one directory call, using the real generated mappers so a name that came from a query rather than from the lookup would show up.

`physicalCountBy` on the stock adjustment is deliberately left as a bare id: it comes from the creation payload and holds an **employee** id, so naming it from this lookup would reintroduce exactly the wrong-person bug.

## What a name means, and how it degrades

Three cases, none of which may render blank, because a blank against an approval reads as nobody having approved it.

1. **A user with no employee record** — the QA raiser account, and any bootstrap administrator. The read is keyed on the `users` row, not on employment, so it resolves like any other. This is the case the old employee-keyed attempt could never handle. Covered in `UserRepositoryIT`.
2. **A user whose name column is empty.** `name` is non-null in the schema but not guaranteed non-empty, so it falls back to the email, which is the only other thing every account has.
3. **A deleted or deactivated user on a historical document.** `deleteAnUser` is a hard delete, so the row is simply gone. The name field then carries `User #<id>` rather than null, so the client renders one field unconditionally and never has to hold the branch, and the `User #<id>` string lives in one place instead of being reimplemented per screen. `nameOf` returns null only when the stamp itself is null, which distinguishes "never approved" from "approver's account is gone".

The read is also **not tenant scoped**, deliberately: a stamp is only reached through a document the caller is already authorised to read, and the approver may have left the organization since. Scoping it would blank exactly the historical records the stamp exists for. What comes back is a display name, already visible to anyone who can read the document.

## Tests

`UserNameLookupTest`, `UserRepositoryIT` (three new cases), `StockAdjustmentStampNameTest`, `ConstructionInvoiceStampNameTest`. Reverting the production change removes the API the tests are written against, so the guard was verified by mutation instead: making an unresolved id return null, and making the page resolve per row, fails 7 of them; both restored, all pass. Full suite green on the build box (heap 25% of the 2048m cap).

`docs/openapi.json` regenerated with `-PupdateOpenApiSnapshot`, and re-verified after rebasing onto the current `development`.

## Follow-up

`echno-core` response types need the new fields added by hand (additive, so nothing breaks meanwhile), and the web placeholders are tracked separately on `echno-web`.

## Summary by CodeRabbit

* **New Features**
  * Invoice records now include display names for submitters, approvers, and payment recorders.
  * Stock adjustments now include display names for users who submitted, approved, rejected, or processed them.
  * Deleted accounts display as `User #<id>`, while actions that never occurred remain null.
* **Documentation**
  * Updated API documentation to describe the new display-name fields and user references.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -3027,7 +3027,7 @@
             "type": "integer"
           },
           "approvedByName": {
-            "description": "Name of the user that approved the invoice. Reads \"User #<id>\" when the account has since been deleted; null only when the invoice was never approved.",
+            "description": "Name of the user that approved the invoice, or their email where the account carries no name. Reads \"User #<id>\" when the account has since been deleted; null only when the invoice was never approved.",
             "example": "Aneesh Johny",
             "type": "string"
           },
@@ -3122,7 +3122,7 @@
             "type": "integer"
           },
           "paymentRecordedByName": {
-            "description": "Name of the user that recorded the most recent payment. Reads \"User #<id>\" when the account has since been deleted; null only when no payment has been recorded.",
+            "description": "Name of the user that recorded the most recent payment, or their email where the account carries no name. Reads \"User #<id>\" when the account has since been deleted; null only when no payment has been recorded.",
             "example": "Anand Rajashekar",
             "type": "string"
           },
@@ -3189,7 +3189,7 @@
             "type": "integer"
           },
           "submittedByName": {
-            "description": "Name of the user that submitted the invoice. Reads \"User #<id>\" when the account has since been deleted; null only when the invoice was never submitted.",
+            "description": "Name of the user that submitted the invoice, or their email where the account carries no name. Reads \"User #<id>\" when the account has since been deleted; null only when the invoice was never submitted.",
             "example": "Anand Rajashekar",
             "type": "string"
           },
@@ -15200,7 +15200,7 @@
             "type": "integer"
           },
           "approvedByName": {
-            "description": "Name of the user who approved the adjustment. Reads \"User #<id>\" when the account has since been deleted; null only when the adjustment was never approved.",
+            "description": "Name of the user who approved the adjustment, or their email where the account carries no name. Reads \"User #<id>\" when the account has since been deleted; null only when the adjustment was never approved.",
             "example": "Aneesh Johny",
             "type": "string"
           },
@@ -15286,7 +15286,7 @@
             "type": "integer"
           },
           "processedByName": {
-            "description": "Name of the user who processed the adjustment. Reads \"User #<id>\" when the account has since been deleted; null only when the adjustment was never processed.",
+            "description": "Name of the user who processed the adjustment, or their email where the account carries no name. Reads \"User #<id>\" when the account has since been deleted; null only when the adjustment was never processed.",
             "example": "Aneesh Johny",
             "type": "string"
           },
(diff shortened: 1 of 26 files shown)


abhijithanandan:
On the review point about the seven `*ByName` properties:

**Taken:** the descriptions documented only two of the three outcomes. They now say the value may be the account's email where its name column is empty, alongside the `User #<id>` placeholder for a deleted account and null for a stamp that was never set.

**Declined, with a reason:** declaring the schemas nullable. `docs/openapi.json` is OpenAPI 3.1 and contains no nullability marker on any property anywhere in the document, neither `nullable` nor a `["string", "null"]` type union. Annotating these seven with `@Schema(nullable = true)` produced no change in the generated output at all, so it would have been a no-op annotation of exactly the kind #601 was about. If the document should carry nullability, that is a generator-configuration change across every optional field, not something to do for seven of them inside a bug fix.

PR #629, opened 2026-08-31
Let the NCR register be asked for a person's own trail

### LINKED ISSUE #626: NCR listing cannot filter on who raised the report, so the name on a detail screen leads nowhere
`GET /api/v1/ncrs/web` takes `inspectionId`, `type`, `status`, `siteEngineerId` and `open`. It does not take a filter on the raiser, and `NcrSpecifications.withFilters` has no predicate for one.

The web app (echno-web#35) turns a person's name inside an item into a link to that module's list filtered to them. On the NCR detail the site engineer is now such a link, because `siteEngineerId` exists as a query parameter and the narrowing stays server-side. "Raised by" cannot be, and it is the field a QA lead actually asks about: which reports did this person raise.

Doing it in the browser is not an option. The endpoint returns a `Page`, and the core client unwraps `.content`, so a client-side filter would run over one page and quietly drop every match outside it. A partial list that looks complete is worse than no filter.

## Asked for

Add `raisedById` (Long, optional) to the list endpoint and to `NcrSpecifications.withFilters`, alongside the existing `siteEngineerId`. Same shape, same semantics: an equality predicate on `raised_by_id`, left out when null.

`raisedById` is an employee id, resolved in `NcrService.currentEmployeeId()` through `employeeRepository.findByUserIdAndOrganizationId`, so it matches what the client already holds from the employee directory. No DTO change is needed: `NcrDto` already carries the field.

Worth considering in the same pass, since they are the same one-line predicate and the same question about a person's trail: `verifiedById` and `closedById`.

## Follow-on

Once this lands, `echno-core` picks the parameters up in `NcrListParams`, and the "Raised by" fact on the NCR detail becomes a link like the site engineer already is. The client side is a few lines and is blocked only on the endpoint.

abhijithanandan:
Closes #626.

`GET /api/v1/ncrs/web` took `siteEngineerId` and nothing else about a person, so the question a QA lead actually asks of a name on an NCR, which reports did they raise, had no answer. The web app could not answer it client-side either: the endpoint returns a `Page` and the core client unwraps `.content`, so a browser filter would have searched one page and read as the whole register. That is why echno-web#358 wired only the site engineer and left "Raised by" as plain text.

## Contract

`raisedById`, `verifiedById` and `closedById` join `siteEngineerId` on the listing. All optional `Long`, all equality predicates on `NcrSpecifications.withFilters`, all ANDed with the filters already there and with each other.

```
GET /api/v1/ncrs/web
  ?inspectionId=&type=&status=
  &siteEngineerId=&raisedById=&verifiedById=&closedById=
  &open=&page=&size=&sort=
```

No DTO change: `NcrDto` already carries the three fields. Auth is unchanged (`@inspectionSecurity.canRead()`), pagination is the existing `Pageable`, and `docs/openapi.json` is regenerated.

## They are employee ids, not user ids

Confirmed independently for each of the three, not assumed from `siteEngineerId`. `NcrService` writes all three from `currentEmployeeId()`, which resolves the session user through `employeeRepository.findByUserIdAndOrganizationId(...).map(Employee::getId)`. So the values a client already holds from the employee directory are the right ones to send.

The first run of the new test proved why this is worth checking: on a fresh database the user and employee sequences run in lockstep and handed out `26` for both, so a test that had confused the two would have passed on the coincidence. The seed now persists three employee-less users first to push the sequences apart, and the assertion is meaningful.

## Tenant isolation

The org scope is a separate condition the Hibernate `orgFilter` adds to every query, and these predicates only narrow within it. `thePeopleFiltersCannotReachAcrossOrganizations` proves it from both sides: org B holds its own NCR raised by its own employee; from org A, filtering on that employee returns nothing, and from org B, filtering on org A's people returns nothing on all three parameters. Rows and `getTotalElements` are both asserted, because the count is a separate query and a leaked total would say how many reports a rival has against a named person even with no row returned.

## Every filter proven by a failing test

Each predicate was removed in turn and the suite rerun. The scope row stands in for a listing that ignored the tenant scope.

| Mutation | Tests that fail |
|---|---|
| `raisedById` predicate removed | `listsByTheThreePeopleOnTheTrail`, `theNewFiltersAndWithTheOnesAlreadyThere`, `thePeopleFiltersCannotReachAcrossOrganizations` |
| `verifiedById` predicate removed | `listsByTheThreePeopleOnTheTrail`, `theNewFiltersAndWithTheOnesAlreadyThere`, `thePeopleFiltersCannotReachAcrossOrganizations` |
| `closedById` predicate removed | `listsByTheThreePeopleOnTheTrail`, `thePeopleFiltersCannotReachAcrossOrganizations` |
| tenant scope not applied to the listing | `thePeopleFiltersCannotReachAcrossOrganizations`, `listsThePunchListAndScopesEverythingToTheTenant` |

A fourth test, `theTrailRecordsEmployeeIdsRatherThanUserIds`, guards the id-kind question above.

## Also

Changelog `077` indexes `raised_by_id`, `verified_by_id` and `closed_by_id`, which had no index because nothing selected on them until now. `site_engineer_id` already had one for the same reason.

## Follow-on

`echno-core` picks the three parameters up in `NcrListParams`, and the "Raised by", "Verified by" and "Closed by" facts on the NCR detail become links like the site engineer already is. That is item 4 of the six remaining on echno-web#35.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -74696,7 +74696,7 @@
     },
     "/api/v1/ncrs/web": {
       "get": {
-        "description": "Returns a page of NCRs in the current tenant. Every parameter is an optional filter. Setting open=true returns the punch list: every NCR that has not been closed, whatever stage it has reached.",
+        "description": "Returns a page of NCRs in the current tenant. Every parameter is an optional filter, and they narrow together. Setting open=true returns the punch list: every NCR that has not been closed, whatever stage it has reached. siteEngineerId, raisedById, verifiedById and closedById are employee ids, the same ids the employee directory hands out, and they answer a person's trail through the register: what they were given, what they raised, what they accepted on re-inspection and what they closed.",
         "operationId": "list",
         "parameters": [
           {
@@ -74740,6 +74740,33 @@
           {
             "in": "query",
             "name": "siteEngineerId",
+            "required": false,
+            "schema": {
+              "format": "int64",
+              "type": "integer"
+            }
+          },
+          {
+            "in": "query",
+            "name": "raisedById",
+            "required": false,
+            "schema": {
+              "format": "int64",
+              "type": "integer"
+            }
+          },
+          {
+            "in": "query",
+            "name": "verifiedById",
+            "required": false,
+            "schema": {
+              "format": "int64",
+              "type": "integer"
+            }
+          },
+          {
+            "in": "query",
+            "name": "closedById",
             "required": false,
             "schema": {
               "format": "int64",
--- a/src/main/java/org/tornotron/echno_backend/inspection/domain/Ncr.java
+++ b/src/main/java/org/tornotron/echno_backend/inspection/domain/Ncr.java
@@ -44,7 +44,10 @@ import java.util.UUID;
                 @Index(name = "idx_ncr_defect", columnList = "defect_id"),
                 @Index(name = "idx_ncr_status", columnList = "status"),
                 @Index(name = "idx_ncr_type", columnList = "type"),
-                @Index(name = "idx_ncr_site_engineer", columnList = "site_engineer_id")
+                @Index(name = "idx_ncr_site_engineer", columnList = "site_engineer_id"),
+                @Index(name = "idx_ncr_raised_by", columnList = "raised_by_id"),
+                @Index(name = "idx_ncr_verified_by", columnList = "verified_by_id"),
+                @Index(name = "idx_ncr_closed_by", columnList = "closed_by_id")
         })
 @Filter(name = "orgFilter", condition = "organization_id = :organizationId")
 @Getter @Setter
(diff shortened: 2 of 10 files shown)


abhijithanandan:
Two updates since the PR was opened.

**The count assertions were not counting.** The cross-tenant test read `getTotalElements` off a page with room to spare, and Spring's `PageableExecutionUtils` skips the count query outright when the first page comes back short of its size, so the total was just the length of the row list. The strongest claim in that test was the one part of it not being exercised. It now asks for one row at a time where two match, which forces the count query to run under the tenant scope, and reads the empty-side totals the same way so a leak would fill the page and then be counted rather than failing on the rows alone.

Re-verified against the same mutation: with the tenant scope not applied to the listing, `thePeopleFiltersCannotReachAcrossOrganizations` and `listsThePunchListAndScopesEverythingToTheTenant` both fail.

**Rebased twice** onto `development` while this was open. The changelog is now **079**, after 077 and 078 landed upstream, and `docs/openapi.json` was regenerated on the rebased tree and comes back byte-identical to what is committed.

Issue #598, opened 2026-08-30
Attribution fields still trust the caller: issue createdById, comment authorId, task creatorId


abhijithanandan:
Found sweeping the codebase for the shape #589 fixed (an identity taken from the request and trusted). The state-changing instances in the attendance family are closed by the follow-up PR; what remains is a family of **attribution fields**: the record is created honestly, but who it says did it is whatever the caller typed. None of these move days or hours, which is why they are an issue and not a fix: the right answer is a judgement call about the API contract, and the clients currently send these fields.

## The instances

| Endpoint | Field | Guard | What a caller can do |
|---|---|---|---|
| `POST /issue-comments/web` | `IssueCommentCreationDto.authorId` | membership | Post a comment attributed to any colleague. The service verifies the employee exists in the tenant and stores the id as given (`IssueCommentService.addIssueComment`). |
| `POST /issues/web` | `IssueCreationDto.createdById` | membership | Raise an issue recorded as created by any colleague. Same pattern in `IssueService.addIssue`. |
| `POST /tasks/web` | `TaskCreationDto.creatorId` | `system-admin` / `project-manager` | Same shape, but role-gated, so only people already trusted to manage tasks can misattribute. Lowest priority. |

The non-web twins carry the same DTOs behind `hasAuthority(...)` guards.

Chat got this right (`ChatService`: sender resolved from the session, edit and delete checked against the stored sender), and regularization was moved to the session in `c7ff4ca`. These three predate that convention.

## Why it matters more than it looks

Comments and issue creation drive accountability: an issue "created by" the site engineer, or a comment in their name, is read as their statement. Once the compliance module reads issues and inspections, forged attribution feeds it.

## The decision to make

Two candidate fixes, and they differ in behaviour:

1. **Resolve the author from the session, ignore the field.** Honest by construction, matches chat and regularization. But an admin filing on someone's behalf is then recorded as themselves, and a signed-in user with no employee record in the tenant (bootstrap admin, cf. #592) cannot post at all unless a fallback is designed.
2. **Keep the field, enforce it is the caller unless the caller holds an admin role** (the #589 pattern). Backwards compatible with what echno-web sends today (it sends the signed-in employee's id), but still lets an admin write history in someone else's name.

[...]

PR #634, opened 2026-08-31
Make a payment voucher's verification a session-stamped action, not a payload field

### LINKED ISSUE #631: A payment voucher's verification stamp is written from the request body, so it can name anyone
Found while doing #621 (PR #630), which named the stamp. Naming it makes this visible rather than causing it.

## What happens

`ConstructionPaymentService.create` and `update` both copy the stamp straight off the payload:

```java
payment.setVerifiedBy(req.verifiedBy());
payment.setVerifiedAt(req.verifiedAt());
```

Nothing checks that the id is the caller, that it is a user in the caller's organization, or that it is a user at all. `verified_by` is a plain `BIGINT` with no foreign key (`v4.0/024-create-construction-payments.xml`). So anyone who can update a voucher can record that somebody else verified it, at a time of their choosing, or clear a verification that did happen. On a payment record that is the wrong answer twice: the audit trail says the wrong person, and it says so silently.

## Why it has not bitten

No web screen writes it. The create and edit forms send nothing for the field, and `echno-core`'s serializer only includes it when a caller sets one. The exposure is the API, not the product, which is why this is its own issue rather than a fix folded into #630.

## What would fix it

The same treatment #598 gave the four session-derived attribution fields: take the verifier from the session at the moment of verification and stop binding it from the request.

That probably wants a verify action rather than a field, since `verifiedAt` has the same problem and the pair only ever moves together:

- `POST /construction-payments/{id}/verify`, stamping `verifiedBy` from the session subject and `verifiedAt` from the clock
- `verifiedBy` and `verifiedAt` dropped from `CreateConstructionPaymentRequest` and `UpdateConstructionPaymentRequest`, so an update can no longer rewrite either
- whether a verifier may verify their own voucher is a `SelfApprovalPolicy` question, the same one #592 answered for approvals

`ConstructionInvoice` already works this way for submit and approve, and its own comment says the invoice stamps were modelled on this field's convention. This is the half of that convention the invoice moved past and the payment did not.

abhijithanandan:
Closes #631.

## What was wrong

`ConstructionPaymentService.create` and `update` copied the stamp straight off the payload:

```java
payment.setVerifiedBy(req.verifiedBy());
payment.setVerifiedAt(req.verifiedAt());
```

Nothing checked that the id was the caller, a user in the caller's organization, or a user at all. So anyone who could edit a voucher could record that a named colleague had verified it, at a time of their choosing, or clear a verification that did happen. Not reachable through the product, because no screen writes those fields. That is containment, not a fix.

## Is verify-on-behalf-of a real requirement?

No, and the evidence is the same shape #598 found for the attribution fields:

- `echno-web`'s `PaymentFormData` (the create and edit form model) has no verification fields at all, and neither `payments/new` nor `payments/[id]/edit` sets one.
- `echno-core` carries them on both request interfaces, but `appendConstructionPaymentOptionalFields` writes them only `if (dto.verifiedBy !== undefined)`, and no call site defines one.
- There is no verify screen anywhere, so there is no flow in which a person records somebody else's check.

So the fields are removed rather than validated, which is the #598 answer rather than the #589 one.

## What replaced them

A verify action, because a verification is a transition with meaning rather than an attribute someone edits, and because `verifiedAt` has the same problem and the pair only ever moves together:

- `POST /finance/construction-payments/web/{id}/verify`, stamping the verifier from the session subject and the time from the clock, behind the same `@PreAuthorize` the rest of the module carries.
- `verifiedBy` and `verifiedAt` dropped from `CreateConstructionPaymentRequest` and `UpdateConstructionPaymentRequest`, so `update()` cannot touch either column.
- Verification is refused on a cancelled voucher and on one already verified. Re-verifying is not an idempotent no-op, it would overwrite a stamp somebody else's check produced, so the existing stamp stands. There is deliberately no unverify: that would be the same silent rewrite from the other direction, and nothing has asked for one.

## Does `SelfApprovalPolicy` apply?

Yes, for the reason it applies to an approval: the verification is the second pair of eyes on money that has already gone out, and the role gate does not supply it, since every verifier may also raise a voucher.

The voucher had nothing to compare a verifier against, though. `created_by` on `BaseEntity` holds the auditing string, not a user id, and comparing across those is exactly what `ApprovalParty` rejects. So changelog **080** adds `raised_by`, stamped from the session on create, which is the move changelog 072 made for regularizations when the same rule had nothing to read. The policy then refuses the raiser, admits a `system-admin` as the recorded exception, and refuses a session that resolves to no user rather than stamping nobody. A voucher raised before the column existed carries a null and is verified without the comparison, which the policy logs at WARN rather than stranding it.

`raisedBy` and `raisedByName` join the response DTO, resolved through the same single `UserNameDirectory` read PR #630 built for `verifiedByName` (one query per page, both stamps), so the break-glass case is visible on the document where the policy's own comment says the raiser and the approver should both be readable.

## Deploy safety

**This side ships alone.** Spring discards unrecognised properties, and every value a client sends for these two fields today is already ignored, so removing them breaks no caller. Refusing a value would have needed the client first; removing one does not. `echno-core` can drop the two optional properties whenever it next changes.

## Tests

| Test | Without the fix |
|---|---|
| `ConstructionPaymentVerificationStampTest.aCreateBodyNamingAVerifier_reachesTheServiceCarryingNoVerificationStamp` | **FAILED** |
| `ConstructionPaymentVerificationStampTest.anUpdateBodyNamingAVerifier_reachesTheServiceCarryingNoVerificationStamp` | **FAILED** |
| `ConstructionPaymentVerificationStampTest.verify_isTheActionThatWritesTheStamp_andIsRoleGated` | passed |

The two red cases post `verifiedBy` and `verifiedAt` on the wire, the way the defect was reachable, and read the request the controller handed the service back as JSON rather than through an accessor. That is deliberate: an accessor would only compile against the payload that no longer has the field, so the test would pass by not existing rather than by the fix. Verified by reinstating the two fields and the two `set` calls on the build machine, leaving everything else in place.

Also added:

- `ConstructionPaymentVerifyActionTest` (real `SelfApprovalPolicy`, not a mock): the session user and the clock are what get stamped; an edit cannot move a stamp already written; the raiser is refused; a `system-admin` is admitted as the recorded exception; a session resolving to no user is refused; a voucher with no raiser recorded is still verifiable; an existing verification is not replaced; a cancelled voucher is refused.
- `ConstructionPaymentServiceIT`: create stamps the raiser, verify stamps a different session against the raiser read back from the column, a second verify is refused, and the raiser is refused. This half runs against a real CockroachDB on purpose, because the comparison depends on changelog 080 having run.
- `ConstructionPaymentStampNameTest`: the raiser name comes from the same single directory read as the verifier's.

`docs/openapi.json` regenerated with `-PupdateOpenApiSnapshot`.

Full suite green on the build machine: heap 574 MB live of the 2048 MB cap, 28% used.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -3532,6 +3532,17 @@
             "format": "int64",
             "type": "integer"
           },
+          "raisedBy": {
+            "description": "User id that raised the voucher, taken from the session that created it. Null on vouchers raised before the voucher recorded this.",
+            "example": 8,
+            "format": "int64",
+            "type": "integer"
+          },
+          "raisedByName": {
+            "description": "Name of the user that raised the voucher, on the same fallbacks as the verifier name. Null only where the voucher does not record who raised it.",
+            "example": "Hrishi",
+            "type": "string"
+          },
           "referenceNumber": {
             "description": "Cheque or reference number for the payment.",
             "example": "CHQ-000123",
@@ -3547,7 +3558,7 @@
               "CANCELLED",
               "REFUNDED"
             ],
-            "example": "VERIFIED",
+            "example": "COMPLETED",
             "type": "string"
           },
           "subContractId": {
@@ -3587,7 +3598,7 @@
             "type": "string"
           },
           "verifiedBy": {
-            "description": "User id that verified the voucher.",
+            "description": "User id that verified the voucher, taken from the session that verified it.",
             "example": 2,
             "format": "int64",
             "type": "integer"
@@ -3893,7 +3904,7 @@
         "type": "object"
       },
       "CreateConstructionPaymentRequest": {
-        "description": "Payload to create a construction payment voucher. The status is not accepted here: a new voucher always starts pending, and the payment number is generated server side.",
+        "description": "Payload to create a construction payment voucher. The status is not accepted here: a new voucher always starts pending, and the payment number is generated server side. Nor is the verification stamp, which is written only by the verify action, from the session and the clock.",
         "properties": {
           "accountNumber": {
             "description": "Bank account number used for the payment.",
@@ -4063,18 +4074,6 @@
           "vendorId": {
             "description": "Vendor being paid, when the payee is a vendor.",
             "example": 17,
-            "format": "int64",
-            "type": "integer"
-          },
-          "verifiedAt": {
-            "description": "Timestamp the voucher was verified, if already verified.",
-            "example": "2026-08-06T10:20:00Z",
-            "format": "date-time",
-            "type": "string"
-          },
-          "verifiedBy": {
-            "description": "User id that verified the voucher, if already verified.",
-            "example": 2,
             "format": "int64",
             "type": "integer"
           }
@@ -17429,7 +17428,7 @@
         "type": "object"
       },
       "UpdateConstructionPaymentRequest": {
-        "description": "Payload to fully replace an editable construction payment voucher. The status is set directly, and no ledger journal entry is created on any status transition.",
+        "description": "Payload to fully replace an editable construction payment voucher. The status is set directly, and no ledger journal entry is created on any status transition. The verification stamp is not replaced here: only the verify action writes it, from the session and the clock.",
         "properties": {
           "accountNumber": {
             "description": "Bank account number used for the payment.",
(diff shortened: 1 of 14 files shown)


abhijithanandan:
@coderabbitai review

Issue #644, opened 2026-08-31
Quota period bounds are computed in the container's timezone, so entitlement windows move with the deployment


abhijithanandan:
`SubscriptionService.calculatePeriodBounds` opens with:

```java
Instant now = Instant.now();
ZonedDateTime zdt = now.atZone(ZoneId.systemDefault());
```

and derives the hourly, daily, monthly and annual quota windows from that zone. Every boundary therefore moves with whatever timezone the container happens to run in.

This is the same defect the UTC sweep was for, and it was missed. It matters more than most survivors of that sweep because it sits on the entitlement path: `checkQuotaAccess` uses the bounds to sum usage in the current period, and `recordUsage` stamps `periodStart` and `periodEnd` on every `UsageRecord` from the same call.

Two consequences:

- A quota window computed under one zone and compared against usage rows stamped under another does not line up, so a user can be allowed or refused on a period that does not match the rows being counted.
- `MONTHLY` and `ANNUAL` are the worst cases. A month boundary shifts by a whole day for any zone east or west of the container's, which silently moves a user between quota periods.

The rest of the codebase settled on UTC for stored instants. These bounds should be computed in UTC too, so a window means the same thing wherever the process runs.

Issue #643, opened 2026-08-31
Subscription is keyed on userId while Plan describes an organization tier, and the backfill is not empty as assumed


abhijithanandan:
`Subscription` is keyed on the buyer, but `Plan` describes a tier an organization buys. The two do not agree, and the disagreement is load-bearing.

## The mismatch

`Subscription.userId` is the key: the column, the index `idx_subscription_user`, `SubscriptionRepository.findActiveSubscriptionByUserId` and `findByUserIdOrderByCreatedAtDesc`, the `SubscriptionCache` key, and `UsageRecord.userId` alongside it.

`Plan`'s own javadoc reads "A subscription tier an organization can subscribe to", and `Plan` carries `maxUsers`, a cap that only means anything against an organization. We sell to construction companies, so the organization is the paying party.

The mismatch is also what `TenantFilter` cites when it declares the whole billing surface unscoped: its reason string is literally "Billing is keyed on the user rather than the organization" (see #640). Rekeying is what removes that justification for good.

## The trivial-backfill premise is false

The payment-gateway research recorded the rekey as trivial "because there are no rows anywhere". That is not the case. Read-only counts against live staging (`echno` on `echno-cockroachdb`) today:

| table | rows |
|---|---|
| `subscription` | 2 |
| `usage_record` | 2 |
| `plan` | 4 |
| `feature` | 1 |
| `plan_feature` | 2 |
| `subscription_item` | 0 |

The two subscriptions belong to users 1 and 2, both on plan 3 (`PRO`), both `TRIALING`, with a current period that ran 2026-05-01 to 2026-05-31 and has been over for three months.

So a rekey is a data migration, not a column rename, and the mapping is not mechanical: it needs a decision per row about which organization a given user's subscription becomes, and users 1 and 2 may well be members of more than one.

Stopping here deliberately rather than migrating on the assumption.

## Two things worth noticing in that data

**The expired subscriptions are still live.** `findActiveSubscriptionByUserId` filters on `status IN (ACTIVE, TRIALING)` and never looks at `currentPeriodEnd`. Both rows are three months past their period end and still resolve as the user's active subscription. `findExpiredSubscriptions` exists for exactly this and is called from nowhere, so nothing ever transitions them.

**Staging is not a clean-room.** Its plan catalogue was seeded by hand and does not come from the changelog (#642), so behaviour observed on staging is not what a fresh environment does.

## What this needs before it can be done

- A decision on what a subscription belongs to when its buyer is in several organizations.
[...]

Issue #642, opened 2026-08-31
Creating a second organization is refused everywhere: the entitlement gate is deliberate but nothing seeds the plans it reads


abhijithanandan:
Creating a second organization fails for every user in every environment that was not seeded by hand. The refusal is correct in intent and wrong in presentation: the gate is deliberate, but nothing configures it and the denial does not say what it wanted.

## This is an unseeded gate, not a bug

`OrganizationService.requireCreateOrganizationEntitlement()` exempts a caller's first organization and puts every one after it through `checkFeatureAccess(userId, "CREATE_ORGANIZATION")`.

That shape was introduced deliberately in b6e2f63 ("Let a new account create its first organization, and say what a denial wanted", closes #521), whose message states the rule outright: "the first is part of signing up and is exempt, every one after it goes through the feature check. Usage is still recorded for both, so the quota stays truthful."

The data model agrees. On staging the feature exists as `CREATE_ORGANIZATION`, named "Organization Limit", of type `QUOTA`, and it is attached to plans with different limits: `FREE` gets 1 `TOTAL`, `PRO` gets 5 `TOTAL`. An organization allowance that varies by tier is a product decision, not an accident.

So the check is working. Two things are missing under it.

## What is actually missing

**1. Nothing seeds the plans.** No changelog under `src/main/resources/db/changelog/` inserts a row into `plan`, `feature` or `plan_feature`. The rows on staging were put there by hand on 2026-05-01 and exist nowhere else. Every fresh environment, including any new production tenant, comes up with an empty plan catalogue.

**2. A user with no subscription row is refused for the wrong reason.** `checkFeatureAccess` returns `noSubscription()`, reason "No active subscription", the moment `loadActiveSubscription` comes back empty. Nothing enrols a new account on the free tier, so the overwhelmingly common case is decided by the absence of a billing record rather than by the free tier's own limit. Seeding the plans alone does not change this, which is why the two have to move together.

The denial then reaches the user as "No active subscription" on a product that has never asked anyone to subscribe.

## Fix

- Seed `FREE`, `STARTER`, `PRO` and `ENTERPRISE`, the `CREATE_ORGANIZATION` feature and their `plan_feature` rows in Liquibase, idempotently and matching what staging already holds so staging is a no-op.
- Resolve a caller with no subscription to the default free plan, so the decision is the free tier's quota rather than the absence of a record.
- Have the refusal name the limit and what exceeded it.

[...]

Issue #641, opened 2026-08-31
The admin billing endpoints are gated on 'billing:admin', an authority this realm has no mechanism to issue


abhijithanandan:
Every administrative billing endpoint is gated on `hasAuthority('billing:admin')`. That string appears 45 times in `src/`, and all 45 are the annotations themselves or commented-out copies of them. It is granted to nobody, and it cannot be granted as things stand.

## Why it cannot be granted

`JwtAuthConverter` mints a bare `resource:scope` authority in exactly one place, `extractPermissions`, which reads the `authorization` claim of an RPT. So `billing:admin` requires a Keycloak Authorization Services resource named `billing` carrying an `admin` scope.

The multi-tenancy audit of 2026-08-18 already traced the identical string `organization:admin` end to end and confirmed against the live staging Keycloak database that zero authorization scopes exist realm-wide, the only resource on `echno-backend-client` is the scopeless `Default Resource`, and the only permission is the scopeless `Default Permission`. A scopeless permission yields no `resource:scope` authority. The same finding applies unchanged to `billing:admin`.

## Affected

- `SubscriptionController`: the six `/user/{userId}` admin endpoints
- `PlanController`: all nine endpoints, including `GET /public`
- `FeatureController`: all seven endpoints

## The `/public` one is separately wrong

`PlanController.GET /public` is named for a public plan catalogue and is the read a pricing page needs. Gating it on a platform-admin authority is wrong independently of whether that authority is grantable.

## Fix

Two different bars, because these are two different kinds of endpoint.

The plan and feature catalogue reads are ordinary authenticated reads. A signed-in user, including one who belongs to no organization yet, needs to see what plans exist.

The catalogue writes and the cross-user subscription administration are platform-staff operations. They must not fall to a customer's own `system-admin`, which would let any tenant edit the global plan catalogue. They need a real platform-staff role, which means a Keycloak **client role** (delivered through `resource_access` as `ROLE_...`) rather than an authorization scope, since client roles are the mechanism that demonstrably works in this realm today.

Replacing the phantom with a grantable role leaves these endpoints refused until the realm carries it, but that is then a deliberate deny with a ticket against it rather than an accident.

PR #647, opened 2026-08-31
Make the billing surface reachable and the organization gate configured

### LINKED ISSUE #521: A self-registered user cannot create an organisation and the API gives no reason
Aneesh went through the public register page on `echno.in` on 28 August and got stuck one step in. Registration itself works now that the Keycloak misconfiguration is fixed, but the next action does not: creating an organisation is restricted to a user holding the organisation admin role, and a freshly self-registered account does not hold it. The response is the generic Spring "Access Denied", so nothing on screen tells him which role is missing or what he is meant to do instead.

Two separate things to fix.

**1. Authorization failures should say what was required.**
Every `@PreAuthorize` denial currently collapses into one opaque message. Map `AuthorizationDeniedException` to a response that names the resource and the role the caller was missing, in wording a non-engineer can act on. Hrishi's `error-message-refactor` branch already has most of the shape for this, so start from there instead of from scratch, and pull it out before the stale-branch cleanup takes it.

**2. Decide and then build the self-signup path.**
Either the first user of a new tenant is granted the organisation admin role at registration so they can create their own organisation, or the register page stops offering a route that dead-ends. Today it is neither, which is what Aneesh walked into.

Raised by Aneesh, 28 August 2026.

### LINKED ISSUE #640: Billing is declared unscoped by path, so every self-service billing endpoint is refused before it is reached
Every self-service billing endpoint returns 403, in every environment, to every caller.

## Mechanism

`TenantFilter.preTenantReason()` matches any path starting `/api/v1/billing` and returns the reason "Billing is keyed on the user rather than the organization". `doFilterInternal` then calls `TenantContext.declareUnscoped(reason)` and short-circuits the chain.

`declareUnscoped` records a reason and nothing else. It never sets an organization id, so `TenantContext.getCurrentOrgId()` is null for the whole request.

`OrganizationSecurityService.hasAnyOrgRoleForCurrentTenant()` and `isMemberOfCurrentTenant()` both open with a null check on that id and return false when it is null. Every `SubscriptionController` endpoint guarded by either one is therefore refused unconditionally:

- `GET /current`, `GET /history`, `POST` (create), `PUT /change-plan`, `POST /cancel` on `@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin')`
- `GET /features/{code}/access`, `POST /usage` on `@orgSecurity.isMemberOfCurrentTenant()`

The short-circuit runs before authentication is even read, so no token and no role changes the outcome.

## Why the declaration is there

It was added deliberately and enumerated in the #547 audit of the unscoped surface. Its stated reason is a description of the current implementation, not a design principle: billing is keyed on `Subscription.userId`, so at the time it genuinely had no organization to be scoped to. That premise is itself the subject of a separate issue.

## Fix

Remove the `/api/v1/billing` branch from `preTenantReason` so billing runs inside tenant isolation like every other authenticated surface.

No billing entity implements `TenantScopedEntity`, and neither does `User`, so nothing in the billing request chain changes what the Hibernate org filter returns. The only behaviour change is that `TenantContext` now carries an organization id, which is what the `@PreAuthorize` expressions were written against.

The genuinely cross-organization admin reads should be declared unscoped on the method, not by a path prefix.

abhijithanandan:
Stage 0 of the payment-gateway work (ClickUp 86d456frt): make the billing surface reachable and the organisation gate configured. **No gateway, no Razorpay, no adapter.** Everything from stage 1 onward stays blocked on the business questions on that ticket, and none of them are guessed here.

Four commits, one per issue, each independently reviewable.

## The headline: the second-organisation block is an unseeded gate, not a bug (#642)

Creating a second organisation fails for every user in every environment that was not seeded by hand, and that is the most user-facing of the findings. I went looking for intent rather than assuming it.

b6e2f63 ("Let a new account create its first organization, and say what a denial wanted", closes #521) introduced this shape deliberately and states the rule in its own message: the first organisation is part of signing up and is exempt, every one after it goes through the feature check. The data model agrees, since `CREATE_ORGANIZATION` is a QUOTA feature named "Organization Limit" whose limit differs per tier. An organisation allowance that varies by plan is a product decision.

So the check is working. Two things were missing under it:

- **Nothing seeded the plans.** No changelog inserts a row into `plan`, `feature` or `plan_feature`. The rows on the IITM box were typed in by hand in May 2026 and exist nowhere else.
- **The refusal did not explain itself.** It reached the caller as the raw reason "No active subscription", which is true and useless to someone who has never been asked to subscribe to anything.

Both are fixed. Each seed changeSet is guarded on its table being empty and marked as run otherwise, so **staging keeps exactly the rows it has and this is a no-op there**. Plan-feature rows resolve their parent ids by code, not by literal id, so an environment that seeded the parents in a different order still attaches each limit to the right tier.

Prices are null and `is_public` is false on every row, deliberately. The price list is an open business question and the RBI e-mandate threshold makes the monthly figure a decision with product consequences, so promoting the May placeholders into every environment would have made them look decided. What the seed does give is a catalogue that exists, which is what lets a plan be granted by hand and the first customers invoiced.

## Billing was refused to everyone, everywhere (#640)

`TenantFilter.preTenantReason()` matched `/api/v1/billing` and called `declareUnscoped`, which records a reason and sets no organisation id. Every billing endpoint is guarded by `hasAnyOrgRoleForCurrentTenant` or `isMemberOfCurrentTenant`, and both refuse a null organisation. The declaration was refusing the whole self-service surface before authentication was even read.

**What removing it breaks: nothing, and I checked rather than assumed.** No billing entity is a `TenantScopedEntity` and neither is `User`, so the Hibernate org filter has nothing to act on and no query changes. #547's audit had enumerated `/billing` among the pre-tenant paths, and its stated reason was a description of how `Subscription` is stored rather than a reason to sit outside isolation. That enumeration is updated in both places it is written down.

One caller-visible change, and it is an improvement: a user in several organisations now gets a 400 naming the `X-Organization-Id` header instead of a 403 that told them nothing. Pinned by a test.

## The admin endpoints were guarded by a string that cannot exist (#641)

`hasAuthority('billing:admin')` occurred 45 times in `src/`, all of them the annotation or a commented-out copy. `JwtAuthConverter` mints a bare `resource:scope` authority only from an RPT `authorization` claim, so it needed a Keycloak resource named `billing` with an `admin` scope. The 2026-08-18 multi-tenancy audit checked the live realm for the identical `organization:admin` and found zero authorization scopes realm-wide, one scopeless Default Resource, one scopeless Default Permission. Not merely unseeded: ungrantable.

Two bars now. Plan catalogue reads are ordinary authenticated reads, since a signed-in user with no organisation has to see what plans exist before being sold one; `GET /plans/public` is the read a pricing page makes and was gated on the admin authority, which is wrong on its own terms. Catalogue writes, the feature catalogue and cross-user subscription administration are platform-staff work and deliberately do **not** fall to a customer's own `system-admin`, which every tenant's administrator holds and which would let any customer rewrite what every other customer is sold. Those are now `ROLE_platform-admin`, a Keycloak client role, which is the grant mechanism this realm has.

**They still refuse until the realm carries that role.** That is now a realm change with a ticket against it rather than an impossibility, and it is called out in the code.

## Quota windows moved with the container timezone (#644)

`calculatePeriodBounds` used `ZoneId.systemDefault()`. The bounds are both the window usage is summed over and the stamps written onto every `UsageRecord`, so the two only agree while every process agrees where a period begins. A month boundary shifts a whole day between zones. Now UTC.

## Stopped deliberately: the rekey to organizationId (#643)

The research recorded this as trivial "because there are no rows anywhere". **That is not the case.** Read-only counts against live staging: `subscription` 2, `usage_record` 2, `plan` 4, `feature` 1, `plan_feature` 2. So it is a data migration, and the mapping is not mechanical: it needs a decision per row about which organisation a user's subscription becomes. Filed with the open questions rather than migrated on an assumption.

Two things that fell out of looking: the two subscriptions are three months past their period end and still read as active, because `findActiveSubscriptionByUserId` filters on status only and `findExpiredSubscriptions` is called from nowhere. And staging's catalogue came from no changelog, so behaviour observed there is not what a fresh environment does.

## Every fix has a test that fails without it

Verified by reverting each fix on the build box and confirming the specific test goes red, not by assuming.

| Test | Pins | Without the fix |
|---|---|---|
| `TenantFilterTest.billingRunsInsideTenantIsolation...` | #640 | FAILED |
| `TenantFilterTest.theBillingGuardsThatRefusedEveryoneNowResolve` | #640, composed with the real `OrganizationSecurityService` | FAILED |
| `TenantFilterTest.aMultiOrganizationCallerMustSayWhich...` | the 400 behaviour change | new |
| `BillingEndpointAuthorityTest.noBillingEndpointIsGuardedBy...` | #641 | FAILED |
| `BillingEndpointAuthorityTest.thePublicPlanCatalogueIsReadable...` | #641 `/public` | FAILED |
| `BillingMappingIT.theChangelogSeedsThePlanCatalogue...` | #642 seed, against a real CockroachDB | FAILED |
| `BillingMappingIT.theSeededCatalogueIsNeitherPricedNorPublic` | #642 price/visibility decision | FAILED |
| `OrganizationServiceEntitlementTest.aSecondOrganizationIsRefused...` | #642 message | FAILED |
| `OrganizationServiceEntitlementTest.aRefusalOnAQuotaNamesTheLimit...` | #642 message | FAILED |
| `OrganizationServiceEntitlementTest.aRefusalOnASingleOrganization...` | #642 singular wording | FAILED |
| `SubscriptionServiceFeatureAccessTest.quotaPeriodBoundsAreComputedInUtc...` | #644 | FAILED |

Two existing tests pinned the old behaviour and were rewritten rather than deleted: `billingDeclaresItselfUnscoped` asserted the exact defect, and the entitlement test asserted the message was literally "No active subscription".

`BillingEndpointAuthorityTest` also pins the endpoint count, so a guard deleted rather than corrected cannot make the rule pass by leaving nothing to check.

## Notes for review

- `docs/openapi.json` regenerated with `-PupdateOpenApiSnapshot`, and re-verified after rebasing onto the movement-record work that landed meanwhile.
- Changelog is **082**; 081 was taken by #635 while this was in flight.
- Whether a user on no plan should fall back to the free tier is left alone on purpose. It is a product decision (open question 4 on the ticket) and today such a user is still refused, which is the behaviour that was already there.

Closes #640, #641, #642, #644. #643 stays open by design.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -25870,7 +25870,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -25980,7 +25980,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -26085,7 +26085,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -26197,7 +26197,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -26310,7 +26310,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -26421,7 +26421,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -26542,7 +26542,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -26655,7 +26655,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -26760,7 +26760,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -26870,7 +26870,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -26982,7 +26982,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller is not authenticated"
           },
           "404": {
             "content": {
@@ -27087,7 +27087,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller is not authenticated"
           },
           "404": {
             "content": {
@@ -27200,7 +27200,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -27311,7 +27311,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller is not authenticated"
           },
           "404": {
             "content": {
@@ -27432,7 +27432,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -27545,7 +27545,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -27668,7 +27668,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -27789,7 +27789,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -28688,7 +28688,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -28809,7 +28809,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
@@ -28931,7 +28931,7 @@
                 }
               }
             },
-            "description": "Caller lacks the billing admin authority"
+            "description": "Caller lacks the platform-admin role"
           },
           "404": {
             "content": {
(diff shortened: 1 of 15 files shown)


abhijithanandan:
@coderabbitai review

Issue #650, opened 2026-08-31
Resources has no role between member and system-admin, and org-manager is the unused role that could have been one


abhijithanandan:
Filing the role-model gap in the Resources domain and the `org-manager` question together, because they are the same gap seen from two ends: there is no role between "member" and "system administrator", and the one role that might have become it never did.

## Reading a material requires system-admin

Eleven `*ControllerWeb` classes in the Resources domain carry `@PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin')")` on every method. Not class-level, per-method, and the `@GetMapping`s carry the identical expression as the `@PostMapping`s and `@DeleteMapping`s:

| Controller | Path |
| --- | --- |
| `MaterialControllerWeb` | `/api/v1/materials/web` |
| `StorageLocationControllerWeb` | `/api/v1/storage-locations/web` |
| `GoodsReceivedNoteControllerWeb` | `/api/v1/grns/web` |
| `SiteTransferControllerWeb` | `/api/v1/site-transfers/web` |
| `PurchaseOrderControllerWeb` | `/api/v1/purchase-orders/web` |
| `PurchaseOrderItemControllerWeb` | `/api/v1/purchase-order-items/web` |
| `IndentControllerWeb` | `/api/v1/indents/web` |
| `IndentItemControllerWeb` | `/api/v1/indent-items/web` |
| `MaterialConsumptionControllerWeb` | `/api/v1/material-consumptions/web` |
| `InventoryTransactionControllerWeb` | `/api/v1/inventory-transactions/web` |
| `VendorControllerWeb` | `/api/v1/vendors/web` |

Two are not: `StockAdjustmentControllerWeb` and `AssetControllerWeb`, both of which read on `isMemberOfCurrentTenant() or hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')` and write on the role pair. That is a two-tier model and it works.

The sidebar gates none of this. `nav/metadata/resources.meta.ts` in echno-web declares no `access` on any entry, `nav/compose` defaults a missing one to `OPEN_ACCESS`, and only two metadata files in the entire nav directory declare access at all. So every Resources link is shown to everybody, and the nav layer could not express this even if somebody wanted it to: its role vocabulary is a three-tier `'admin' | 'manager' | 'employee'` that does not name Keycloak roles.

The result is a storekeeper who sees Materials in the menu, clicks it, and gets a 403. Which is the visible symptom, and the smaller half of the problem.

[...]

PR #654, opened 2026-08-31
Say what the org-scoped roles guide describes, which is seven roles

### LINKED ISSUE #653: docs/org-scoped-roles.md is stale: four roles where the code has seven, and it still says to create Keycloak subgroups by hand
`docs/org-scoped-roles.md` is the only description of this authorization layer anybody reads, and three things in it are wrong.

**It names four roles where `OrgRole` holds seven.** The document lists `system-admin`, `org-manager`, `hr-admin` and `project-manager`, in the enum listing and in both group-tree illustrations. `qa-engineer`, `safety-officer` and `site-engineer` were added and the guide was not. A role that exists and is undocumented is worse than an undocumented feature, because the guide reads as complete: somebody building an authorization decision on it is working from a list with entries missing.

**It tells an operator to create subgroups by hand.** Scenario 8, step 3:

> For **existing** organizations, you need to manually create the subgroup. You can do this by:
>   - Going to the Keycloak admin console → Groups → find `org-{id}` → create child group `finance-admin`
>   - Or writing a one-time migration script that calls `createDefaultRoleSubgroups` for all existing orgs

and note 5:

> **Existing organizations won't have subgroups.** Organizations created before this change don't have the role subgroups. You'll need to either recreate them or add subgroups via the Keycloak admin console.

Neither is true. `KeycloakGroupService.assignOrgRole` calls `ensureRoleSubgroup`, which creates a missing subgroup on first assignment, handles the race where two assignments of the same new role collide, and takes the name from the enum so it cannot invent an authority. Following the old instruction is not harmless: it is per-tenant console work the code already does, and it invites hand-typed group names that no enum vouches for.

**It leaves `org-manager` looking like a role with a purpose.** The table says it "can manage org settings and employees". It gates nothing: `'org-manager'` appears in zero `@PreAuthorize` expressions in `src/main`. What it does grant is manager standing, through `OrgRole.getManagerRoles()`, which decides who may be named the manager on a project invite code (`ProjectInviteCodeService`) and who appears in the manager listings (`EmployeeHierarchyService`). Someone holding it and nothing else can be named a manager and will see no screen that says so. That is worth a sentence next to the role, both because it is surprising and because it means removing the constant would be a behaviour change rather than a cleanup.

The javadoc on `createDefaultRoleSubgroups` lists the same four names and has the same problem.

Fixing all of it, with `OrgRoleDocumentationTest` pinning the guide against the enum so it cannot drift again. The check is a name-presence one on purpose: anything stricter would be a second copy of the enum living in a test, which is the problem rather than the fix.

Found while writing the user guide (tornotron/echno-web#371).

abhijithanandan:
Closes #653.

`docs/org-scoped-roles.md` had drifted to naming four roles where `OrgRole` holds seven, and two of its instructions had become wrong rather than merely incomplete.

**The roles.** `qa-engineer`, `safety-officer` and `site-engineer` were added and the guide was not, in the enum listing and in both group-tree illustrations. The guide reads as complete, so somebody building an authorization decision on it was working from a list with entries missing. Also added the note on why the three inspection roles live in this layer rather than among the realm occupation roles, since that is the question the addition raises.

**The Keycloak console instructions.** Scenario 8 step 3 and note 5 both told an operator to create role subgroups by hand for every existing organization, or to recreate the organizations. Neither has been true since `ensureRoleSubgroup` began creating a missing subgroup on first assignment. Following them is not harmless: it is per-tenant console work the code already does, and it invites hand-typed group names that no enum vouches for.

**`org-manager`.** The table had it down as managing org settings and employees. It gates nothing: `'org-manager'` appears in zero `@PreAuthorize` expressions. What it grants is manager standing through `getManagerRoles()`, which decides who may be named the manager on a project invite code and who appears in the manager listings. Worth a sentence next to the role, both because it is surprising and because it means removing the constant would be a behaviour change rather than a cleanup. The wider role-model question is #650; this only stops the guide from misdescribing it.

The javadoc on `createDefaultRoleSubgroups` listed the same four names and is now written in terms of the enum.

## The test

`OrgRoleDocumentationTest` pins all three against the code. The role check is a name-presence one on purpose: anything stricter would be a second copy of the enum living in a test, which is the problem rather than the fix, and anything looser would not have caught the drift it exists to prevent.

Verified red before green. With the pre-change document in place all three fail; with it restored all three pass.

```
OrgRoleDocumentationTest > everyOrgRoleIsNamedInTheGuide() FAILED
OrgRoleDocumentationTest > theGuideDoesNotStillAskForSubgroupsToBeCreatedByHand() FAILED
OrgRoleDocumentationTest > theGuideDescribesWhatOrgManagerActuallyGrants() FAILED
```

Full `./gradlew build` green on the lab box, 8m 50s, test heap 556 MB live of the 2048 MB cap.

Found while writing the user guide (tornotron/echno-web#371).

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for three additional organization roles: QA Engineer, Safety Officer, and Site Engineer.
  * Clarified role semantics, manager-role definitions, and role subgroup structure.
  * Documented automatic creation of missing role subgroups during role assignment.
* **Tests**
  * Added checks to keep the role documentation aligned with available roles and current subgroup-management behavior.

### DIFF
--- a/docs/org-scoped-roles.md
+++ b/docs/org-scoped-roles.md
@@ -21,7 +21,10 @@ org-5/                      ← This is a folder (Keycloak group) for the organi
     ├── system-admin        ← This is a subfolder (subgroup) for the role
     ├── org-manager
     ├── hr-admin
-    └── project-manager
+    ├── project-manager
+    ├── qa-engineer
+    ├── safety-officer
+    └── site-engineer
 ```
 
 - When a user **joins an organization**, they are placed inside the `org-5` folder.
@@ -43,7 +46,7 @@ When someone creates an organization (let's say org with ID 5), your `Organizati
 
 This method does two things:
 1. Creates the parent group `org-5` in Keycloak (this already existed before)
-2. **NEW**: Creates role subgroups inside it: `system-admin`, `org-manager`, `hr-admin`, `project-manager`
+2. Creates one role subgroup inside it for every value of `OrgRole`
 
 After this, Keycloak's group tree looks like:
 ```
@@ -51,7 +54,10 @@ org-5/
     ├── system-admin
     ├── org-manager
     ├── hr-admin
-    └── project-manager
+    ├── project-manager
+    ├── qa-engineer
+    ├── safety-officer
+    └── site-engineer
 ```
 
 **File:** `KeycloakGroupService.java` → `createOrganizationGroup()` and `createDefaultRoleSubgroups()`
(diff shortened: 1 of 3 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
All three taken, plus the outside-diff one on `createOrganizationGroup`.

- **Concurrent assignment.** Right, and the way it was worded was worse than incomplete: it said the loser is told the group already exists and stopped there, which reads as if that caller were left holding an error. It is caught inside `ensureRoleSubgroup`, the subgroup is read back, and the assignment goes through, so both callers succeed. Reworded to say so.
- **`createOrganizationGroup` javadoc.** Same four-role tree, stale in exactly the way this branch exists to fix. Rewritten to point at `OrgRole` rather than name the roles, since naming them is what let both it and the guide fall three behind.
- **The test assertions.** The fair half of this is the first-assignment one: asserting only the absence of the old console instructions would pass a guide that deleted the paragraph and said nothing, leaving a reader who has just added a role with no answer about existing tenants, and the answer they would reach for is the console one that was deleted. Now pinned by name, and `recreate them` from the old note 5 is refused too. The `org-manager` check wanted the same treatment: the word `getManagerRoles` alone does not say what holding one of them does, so it now pins the invite-code manager and the manager listings.

MD040 language on both group-tree fences.

Re-ran `OrgRoleDocumentationTest` on the lab box, three passing.

abhijithanandan:
Merging on green CI. CodeRabbit reviewed the branch and returned three actionable comments plus one outside the diff, all four taken in cfb2b67; the re-review of that follow-up came back "Review rate limited" and no second review was commissioned, so **recording this as merged with the follow-up commit unreviewed**.

My own reasoning for merging it as it stands: that commit changes one Markdown sentence about concurrency, one javadoc block that now names no roles at all, two fence languages, and three additional string assertions in a test that has no production callers. `build-and-test` is green over all of it. The only thing in the branch that executes in production is the javadoc, which executes nothing.

PR #656, opened 2026-08-31
Freeze a payment voucher once it is verified, and make cancelling the way back

### LINKED ISSUE #636: A verified payment voucher can still be edited, so the stamp vouches for figures the verifier never saw
Found reviewing PR #634. That PR made the verification stamp session-derived and unwritable from a payload, which closed #631. It did not close this, and it is the half that becomes meaningful once a verify action exists.

## What happens

`ConstructionPaymentService.update` is a full replacement, and its only lifecycle guard is that a CANCELLED voucher cannot be updated. A verified voucher is editable: the amount, the payee, the bank details and the payment date can all be replaced while `verifiedBy` and `verifiedAt` stand untouched.

PR #634 added a test pinning that the stamp survives an edit (`ConstructionPaymentVerifyActionTest.anUpdateCannotWriteTheStamp_soAVerifiedVoucherSurvivesAnEdit`). Surviving is the right half of the answer and the wrong whole one: the stamp then attests to figures nobody checked. Someone can raise a voucher for 1,000, have a colleague verify it, and edit it to 100,000 with the verification still reading as current.

## Precedent in the codebase

`StockAdjustmentService` freezes the document once it is posted, for exactly this reason. `ConstructionInvoiceService.update` refuses to move an invoice into APPROVED, because approval is a posting event, though it does not freeze an approved invoice either.

## Candidate fixes

1. **Refuse an edit of a verified voucher**, on the stock-adjustment pattern. Cleanest, and the one that makes the stamp mean something. Needs an answer for correcting a genuine mistake: either the voucher is cancelled and re-raised, or an unverify action is added, which #631 deliberately left out.
2. **Clear the verification when the voucher is edited**, so an edit sends it back for checking. Weaker, but it never strands anyone and it cannot leave a stale attestation standing.
3. **Freeze only the fields the verification is about** (amount, payee, payment date, bank details) and leave description and notes editable.

This is a lifecycle decision rather than a bug fix, which is why it is an issue. Note the module posts no ledger entries in this increment, so nothing downstream is wrong yet; what is wrong is the record.

Related: the voucher's status set is `PENDING / PROCESSING / COMPLETED / FAILED / CANCELLED / REFUNDED`, and only CANCELLED is refused for verification. Worth confirming at the same time that verifying a FAILED or REFUNDED voucher is intended.

abhijithanandan:
Closes #636.

## What was wrong

`ConstructionPaymentService.update` was a full replacement whose only lifecycle guard was that a cancelled voucher could not be touched. A verified voucher was fully editable: the amount, the payee, the bank details and the payment date could all be replaced while `verifiedBy` and `verifiedAt` stood untouched.

PR #634 added a test pinning that the stamp survives an edit. It does survive, and that turned out to be the right half of the answer and the wrong whole one, which is what #636 was raised for: the stamp then attests to figures nobody checked. Raise a voucher for 1,000, have a colleague verify it, edit it to 100,000, and the verification still reads as current. #631 stopped the stamp being forged; it did not stop it being outlived.

## The shape, and why not the other two

The issue offered three, and they fail differently, so the choice is argued rather than picked.

**Clearing the verification when the voucher is edited.** Rejected. It never strands anyone, which is its real merit, but it makes every edit an unverify. #631 deliberately left out an unverify action because clearing a verification is the same silent rewrite of the record as forging one, from the other direction, and it is no better for arriving as a side effect of an edit than as a button. It also hands anyone who wants a verification gone a trivial way to get it: change a note.

**Freezing only the fields the verification is about.** Rejected. The boundary between "the figures" and "the annotations" is a judgement, and it drifts: every field added later has to be classified, nothing catches a misclassification, and the hole reopens silently. The voucher has 22 editable fields today.

**Refusing the edit outright.** Implemented. It is the rule `StockAdjustmentService` already applies once a document is posted, for the same reason its comment gives: changing the lines afterwards would leave the record describing a document that no longer exists. Its failure mode is being stuck, and that is answered below rather than by leaving the document editable.

## Freezing needs a route back, and there wasn't one

This is the part that turned out to matter. The issue names cancel-and-re-raise as the correction route, and it existed only in theory:

- the sole way to reach `CANCELLED` was `PUT /{id}` with `status: CANCELLED`, which is exactly the call the freeze now blocks;
- nothing in `echno-web` or `echno-core` cancels a voucher at all, so there is no cancel path to preserve.

Freezing on its own would therefore have left a verified voucher with no way out. So cancelling becomes an action of its own, `POST /{id}/cancel`, on the same argument that made verification one in #631: it is a transition with meaning rather than a field to set, and routing it through a full replacement of every other field is how a cancellation ends up also changing an amount.

- **A verified voucher can be cancelled.** Voiding a document is not editing the figures its verification attested to; it withdraws the document those figures were on. The stamp stays where it is, so the record still shows who checked what and that it was later thrown out.
- **The reason is required**, for the reason `StockAdjustmentService.reject` requires one: a voided voucher that does not say what was wrong with it explains nothing, and on a verified voucher it is the only record of why somebody's check was set aside. Changelog **083** adds `cancellation_reason`.
- **`update` refuses to set the cancelled status**, so there is one cancellation route and it always records a reason.
- **Cancelling is one-way**, and there is still no unverify. The replacement is raised alongside.

## Also here: the same race review found on #639

Every write path on this voucher reads its state, decides on it, and then writes. Two arriving together each read the voucher as it was before the other, both pass the guard, and both act: an edit could land on a voucher that had just been verified, or a second verification could replace the first. `update`, `verify` and `cancel` now read through a locking finder, the way the stock-adjustment, leave-approval and payable paths do. Reads that only display a voucher keep the unlocked lookup.

## The status question the issue asked

Verification stays refused on a cancelled voucher only, and that is deliberate rather than an oversight. A `FAILED` or `REFUNDED` payment still has figures worth checking, and a verification of one is a meaningful record. `CANCELLED` is the only status that says the document itself is void, so it is the only one with nothing left to verify.

## Tests

Each case was proved red by deleting exactly the code it pins, one deletion at a time, on the build machine.

| Test | Reddened by deleting |
|---|---|
| `ConstructionPaymentVerifyActionTest.anEditIsRefusedOnceTheVoucherIsVerified_ratherThanOutlivingTheStamp` | the freeze guard |
| `ConstructionPaymentFreezeAndCancelTest.aVerifiedVoucherCannotBeEdited` | the freeze guard |
| `...theRefusalNamesTheCorrectionRoute` | the freeze guard |
| `...anUnverifiedVoucherIsStillEditable` | nothing, by design: it is the pin against an over-broad freeze, and goes red when the guard is made unconditional |
| `...cancellingRecordsTheReasonAndKeepsTheStamp` | the reason write |
| `...aSecondCancellationIsRefused` | the reason write |
| `...anUpdateCannotCancel` | the update-cancel guard |
| `...everyWritePathReadsUnderAWriteLock` | the locking finder |
| `ConstructionPaymentVerificationStampTest.cancel_isRoleGated_andRefusesAVoucherVoidedWithNoReasonGiven` | the endpoint, its `@PreAuthorize`, or the `@NotBlank` on the reason |
| `ConstructionPaymentServiceIT.aVerifiedVoucherIsFrozen_andCancellingIsWhatCorrectsIt` | the freeze guard or the reason write |

Two things worth saying plainly about these.

`anUpdateCannotWriteTheStamp_soAVerifiedVoucherSurvivesAnEdit`, which #634 added, is **replaced** rather than kept. It asserted that the stamp survived the edit, which is the behaviour this PR removes; keeping it would have pinned the defect. The case is rewritten in place, under the name above, so the history of what it used to assert stays visible, and it now also pins that the amount did not move. The fixture gained a distinct amount for that: it was defaulting to zero, so an amount assertion would have compared zero to zero and passed whether or not the edit went through.

**The IT case only passes because the migration ran.** `cancellation_reason` is a real column there, against a real CockroachDB, not a field on a mock, which is what makes it evidence that changelog 083 works rather than that the code compiles.

`docs/openapi.json` regenerated with `-PupdateOpenApiSnapshot`. Full suite green on the build machine: 599 MB live of the 2048 MB cap, 29% used.

## Client

`POST /{id}/cancel` is new and nothing calls it yet, so nothing breaks. The client work is filed rather than built.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -2029,6 +2029,19 @@
         },
         "type": "object"
       },
+      "CancelConstructionPaymentRequest": {
+        "description": "Why a payment voucher is being voided. Required: a cancelled voucher that does not say what was wrong with it explains nothing, and on a verified voucher this is the only record of why the verification was set aside.",
+        "properties": {
+          "reason": {
+            "description": "What was wrong with the voucher.",
+            "example": "Duplicate of CPMT-000118, raised twice for the same invoice",
+            "maxLength": 1000,
+            "minLength": 0,
+            "type": "string"
+          }
+        },
+        "type": "object"
+      },
       "CategoryCreationDto": {
         "description": "Payload to create a work category used to classify tasks.",
         "properties": {
@@ -3418,6 +3431,11 @@
           "bankName": {
             "description": "Bank the payment was drawn on.",
             "example": "HDFC Bank",
+            "type": "string"
+          },
+          "cancellationReason": {
+            "description": "Why the voucher was voided, recorded by the cancel action. Null on a voucher that has not been cancelled.",
+            "example": "Duplicate of CPMT-000118, raised twice for the same invoice",
             "type": "string"
           },
           "currency": {
@@ -37349,7 +37367,7 @@
     "/api/v1/finance/construction-invoices/web/{id}/cancel": {
       "post": {
         "description": "Cancels an invoice and records the supplied reason. If the invoice was already posted, the ledger entry is reversed. The reason is recorded on the reversing entry's description, which also carries a prefix naming the entry being reversed, so it must be present and at most 455 characters.",
-        "operationId": "cancel_1",
+        "operationId": "cancel_2",
         "parameters": [
           {
             "in": "path",
@@ -38224,7 +38242,7 @@
         ]
       },
       "put": {
-        "description": "Replaces the editable fields of a payment voucher, including its status, which is set directly. No ledger journal entry is created on any status transition.",
+        "description": "Replaces the editable fields of a payment voucher, including its status, which is set directly. No ledger journal entry is created on any status transition. A verified voucher is frozen and cannot be edited, because the verification would then stand against figures nobody checked; cancel it and raise a replacement. The cancelled status cannot be set here either: use the cancel action, which records why.",
         "operationId": "update_4",
         "parameters": [
           {
@@ -38266,7 +38284,7 @@
                 }
               }
             },
-            "description": "Validation failed, or the voucher is not in an editable state"
+            "description": "Validation failed, the voucher is verified or cancelled, or the payload asks for the cancelled status"
           },
           "402": {
             "content": {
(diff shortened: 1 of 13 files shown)


abhijithanandan:
@coderabbitai review

PR #658, opened 2026-08-31
Approve a stock adjustment against the balance it was raised against

### LINKED ISSUE #651: A stock adjustment approves against today's balance and overwrites the count sheet, so anything posted in the window is silently reversed
A stock adjustment records a physical count against the system quantity at the moment the count sheet is written. On approval it is posted against the balance as it stands then, and the line is rewritten to match.

`StockAdjustmentService.applyLineItems` stores what the client sent:

```java
item.setSystemQuantity(lineDto.getSystemQuantity());
item.setPhysicalQuantity(lineDto.getPhysicalQuantity());
item.setAdjustmentQuantity(lineDto.getAdjustmentQuantity());
```

Approval re-reads current stock and overwrites both figures:

```java
Optional<Double> existingBalance = location != null
        ? inventoryService.findStockAtLocation(material.getId(), project.getId(), location.getId())
        : inventoryService.findUnlocatedStock(material.getId(), project.getId());
...
Double balance = existingBalance.orElse(0.0);
Double movement = resolveMovement(line, balance, material, id);

// Record what was actually posted, so the document and the ledger agree.
line.setSystemQuantity(balance);
line.setAdjustmentQuantity(movement);
```

and `resolveMovement` ignores the drafted `systemQuantity` entirely:

```java
if (line.getPhysicalQuantity() != null) {
    return line.getPhysicalQuantity() - balance;
}
if (line.getAdjustmentQuantity() != null) {
    return line.getAdjustmentQuantity();
}
```

The header total is recomputed the same way.

This is deliberate and the javadoc says so: the adjustment is "posted as the difference between that count and the balance as it stands now, not the `systemQuantity` recorded when the count sheet was written." The document and the ledger agree afterwards, which is the property it was written for and is a real one.

## What it costs

The drafted `systemQuantity` becomes advisory. A line carrying a physical count always lands the balance exactly on the counted figure, so anything that moved between the count and the approval is silently absorbed. A goods receipt booked in that window is reversed by the approval. A consumption posted in that window is undone. Neither raises anything, because from the adjustment's point of view the count is simply right.

That is the case worth being concrete about, since it is not exotic. Someone counts a store on Friday, an adjustment is raised, a delivery is received on Monday morning, and the approval that afternoon takes the balance back down to Friday's count. The receipt is still in the ledger. The stock is not.

The second cost is to the document. After approval you cannot see what the system said at count time, because the field that said it was overwritten. So the one piece of evidence that would show the window mattered is destroyed by the same operation that made it matter. A line with only an `adjustmentQuantity` and no count behaves as a plain signed delta and the movement is the drafted one, but `systemQuantity` is still overwritten with today's balance.

## What this needs a decision on

Not whether the code is doing what it says. It is. Whether the approval should:

- **Refuse when the balance has moved**, and make the approver re-count or explicitly override. Safest, and it puts a stall in the middle of a workflow that people currently complete without thinking.
- **Post the drafted variance rather than a recomputed one**, so a receipt in the window survives and the balance ends up wrong by the amount that was received, which is arguably the honest answer if the count was of a shelf that had not yet been added to.
- **Keep posting to the count and record the drift**, adding a field for what the system said at draft time so the two can be compared afterwards. Cheapest, changes no behaviour, and turns an invisible reversal into a visible one.

The third is a small change and might be worth doing regardless of which of the first two is chosen, because none of them can be evaluated on live data while the evidence is being overwritten.

The user guide flags this as the first thing that will surprise you in the module, which is where it was found: echno-web #371. Verified against `development` at the time of filing.

abhijithanandan:
Closes #651.

## What was wrong

Approval re-read the balance, recomputed the variance from it, and overwrote the line with the result. A line carrying a physical count therefore always landed the balance exactly on the counted figure, so anything booked between the count and the approval was absorbed with nothing anywhere saying so. A goods receipt in that window is reversed: it stays on the ledger, the stock does not. And the `systemQuantity` that would have shown the window mattered was overwritten by the same operation that made it matter.

This is the gap #559 (segregation of duties) and #631/#635 (a verification is stamped by whoever performed it) leave open here. A second signature on figures that changed between raising and approving is not a second signature on anything.

## What the draft's system quantity was actually for

Nothing. `applyLineItems` stored whatever the client sent, nothing validated it, `resolveMovement` ignored it, and approval overwrote it. On the web form (`features/stock-adjustments/components/stock-adjustment-form.tsx`) "Current Stock" is a free-text number input a person types, defaulting to `0`, never read from the stock listing. So it was a hand-typed number the server never looked at.

That fact decides the design. A guard that compared the current balance against a hand-typed number would fire on typing mistakes rather than on stock moving, and with a form defaulting to zero it would refuse nearly every approval. So the field has to become the server's before it can be guarded.

## The decision

**Refuse the approval when the balance has moved.** Of the three options on the issue:

- *Post the drafted variance instead* is not safer, it just guesses the other way. Whether goods received after a count were already on the shelf when it was taken is not something the data can answer, so that default either absorbs a real receipt or double-counts one.
- *Record the drift and keep posting* makes an invisible reversal visible, but still posts it.

Both of the silent options need a person to look, and the approval is the moment a person is standing there. Refusing costs a stall only when the balance actually moved. The way back is to save the document again, which takes up the current balance, then approve once the count has been checked against the figure that moved. That keeps the decision with a person and leaves the document saying what was agreed.

## The change

- `stampOpeningBalance` reads `systemQuantity` from the balance when the document is written, and recomputes the variance from it where the line carries a count, so the line's three figures are one piece of arithmetic. Same reasoning as `submittedBy` being taken from the session rather than the body. A line naming no material, or on a document naming no project, keeps what was sent: there is no balance to read, and such a line cannot be approved either way.
- `requireBalanceUnmoved` refuses the approval when the balance has moved away from that figure, naming both figures and the route forward.
- `readBalance` is extracted so the stamp and the posting read the same row. If they read different rows, every approval would look moved.
- The two writes in `approve` now record rather than rewrite: with the guard passed they change nothing on a line carrying an opening balance. They are what fills the figures in on a line raised before the stamp existed, which posts as before.

`#559` self-approval and `#572` scope narrowing are untouched.

No schema change. `docs/openapi.json` regenerated for the three Schema description changes.

## Evidence from the one real use

`SA-563` on `echno.in` (read-only): raised 14:03, approved 19:11 on 2026-08-30. Its stored line figures agree with the ledger rows exactly (400 / -30 / 370 and -30 / +30 / 0), so the correction is self-explaining. But that holds only because nothing moved in those five hours: the ledger's previous row is 2026-08-28. The rewrite was a no-op by luck of a quiet window, not by design. A five-hour gap between raising and approving is the exposure.

## Tests

Five new tests fail against `development` and pass here.

| Test | Without the fix |
|---|---|
| `aCountedLineIsRefusedWhenTheBalanceHasMovedSinceTheDocumentWasRaised` | Posts, driving the balance to the count and reversing the receipt |
| `aSignedDeltaLineIsRefusedWhenTheBalanceHasMovedToo` | Posts against a balance the approver never saw (figures chosen to keep the closing figure clear of zero, so only the moved balance can refuse it) |
| `create_stampsTheOpeningBalanceRatherThanTrustingTheFigureSent` | Stores the client's 999 |
| `create_recomputesTheVarianceFromTheStampedOpeningBalance` | Stores 400 / 370 / -130, which is not one piece of arithmetic |
| `update_restampsTheOpeningBalance` | Keeps the stale 400, so the route back from a refusal does not work |

Three more pin new code that has no counterpart on `development`, so they pass there by construction: `anApprovalPostsTheFiguresTheDocumentWasRaisedWith` (delete the guard and it fails, because approval writes 50 over the raised 48), `aBalanceMovedOnlyByFloatingPointResidueIsNotTreatedAsMoved` (delete the `NO_MOVEMENT` tolerance and it fails), `aLineRaisedWithNoOpeningBalanceIsPostedAndHasItsFiguresFilledIn` (delete the null check and it throws an NPE). `create_lineWithNoMaterial_keepsTheFigureSent` pins the early return; delete it and `material.getId()` throws.

The old `aCountIsPostedAgainstTheBalanceAsItStandsNotTheSystemQuantityOnTheCountSheet` pinned the defect and is replaced.

Full `./gradlew build` green on `lab-node-116-f`. Test heap 559 MB of 2048 MB (27%), 8 contexts cached.

## Follow-up, not in this PR

The header `totalVarianceQuantity` is still copied from the request on a draft and is not checked against the lines. Approval recomputes it, and the ledger is never written from it, so it is a display looseness rather than an audit one, but it is the same class of problem. Worth a separate issue.

The user guide the issue was found from (echno-web #371) documents the old behaviour and needs the paragraph rewritten.

## Summary by CodeRabbit

* **New Features**
  * Stock adjustments now record the current opening balance when created or updated.
  * Adjustment quantities are recalculated from recorded and counted values.
  * Approval validates that stock has not changed since the adjustment was created.
  * Approved adjustments retain consistent opening balances and movement values.

* **Bug Fixes**
  * Prevented posting adjustments against outdated stock balances.
  * Improved handling of multi-line adjustments, missing balances, and minor floating-point differences.

* **Documentation**
  * Clarified quantity calculation, balance stamping, and approval behavior in the API documentation.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -15911,7 +15911,7 @@
         "description": "A single material line on a stock adjustment, comparing the system quantity to the physically counted quantity.",
         "properties": {
           "adjustmentQuantity": {
-            "description": "Adjustment quantity (physical minus system).",
+            "description": "Adjustment quantity (physical minus system). Recomputed from the stamped system quantity whenever a physical count is given, so the line's three figures are one piece of arithmetic. Sent on its own, with no physical count, it is the signed movement the line posts and is kept as sent.",
             "example": -20.0,
             "format": "double",
             "type": "number"
@@ -15960,7 +15960,7 @@
             "type": "string"
           },
           "systemQuantity": {
-            "description": "Quantity on record before the adjustment.",
+            "description": "The quantity on record before the adjustment. Stamped from the balance this line is raised against, so any value sent here is ignored: it is the opening figure the approval is checked against and therefore the server's to state. A value sent for a line naming no material, or on a document naming no project, is kept, because there is no balance to read; such a line cannot be approved either way.",
             "example": 480.0,
             "format": "double",
             "type": "number"
@@ -16052,7 +16052,7 @@
             "type": "string"
           },
           "systemQuantity": {
-            "description": "Quantity on record before the adjustment.",
+            "description": "Quantity on record before the adjustment, read from the balance when the document was last written. An approval is refused if the balance has moved away from it, so on a posted line this is also the opening figure the ledger entry carries.",
             "example": 480.0,
             "format": "double",
             "type": "number"
--- a/src/main/java/org/tornotron/echno_backend/stockAdjustment/StockAdjustmentService.java
+++ b/src/main/java/org/tornotron/echno_backend/stockAdjustment/StockAdjustmentService.java
@@ -51,10 +51,16 @@ import java.util.stream.Stream;
  * then moves the balance, so the resulting figure stays explainable from the ledger like
  * every other movement.
  *
- * <p>{@link #create} and {@link #update} persist the document only. Posting happens once,
- * through {@link #approve}, and a posted document is then frozen: edits and deletes are
- * refused, because changing the lines afterwards would leave the ledger describing a
- * document that no longer exists.
+ * <p>{@link #create} and {@link #update} persist the document only, with one thing read from
+ * the stock rather than the request: each line is stamped with the balance it is being raised
+ * against, by {@link #stampOpeningBalance}. Posting happens once, through {@link #approve}, and
+ * a posted document is then frozen: edits and deletes are refused, because changing the lines
+ * afterwards would leave the ledger describing a document that no longer exists.
+ *
+ * <p>Those two halves meet at {@link #requireBalanceUnmoved}: an approval is refused when the
+ * balance has moved since the document was raised. The figures approved are then the figures
+ * raised, which on a document whose whole purpose is to be an auditable correction is what the
+ * second signature is a signature on.
  *
  * <p>A draft has one other way off the line: {@link #reject}, which records that an approver
  * looked at the correction and refused it. It moves no stock, and it is the only outcome that
@@ -238,14 +244,19 @@ public class StockAdjustmentService {
      * Approves a stock adjustment and posts its lines to the stock ledger.
      *
      * <p>This is the controlled way to set or correct a balance. Each line resolves the
-     * balance row it applies to, from the line's own storage location or the document's,
-     * and the movement it needs to reach the counted figure. A line carrying a physical
-     * count is posted as the difference between that count and the balance <em>as it stands
-     * now</em>, not the {@code systemQuantity} recorded when the count sheet was written, so
-     * the balance ends at exactly the counted figure even if stock moved in between; the
-     * line is rewritten with the figures actually posted. A line with no count falls back to
+     * balance row it applies to, from the line's own storage location or the document's, and
+     * the movement it needs to reach the counted figure: a line carrying a physical count posts
+     * the difference between that count and the balance, and a line with no count falls back to
      * its signed {@code adjustmentQuantity}. Lines that come out at no movement are skipped
      * rather than writing a ledger row that changes nothing.
+     *
+     * <p>The approval is refused when the balance has moved since the document was raised, so
+     * what is posted is the arithmetic the approver read and not a recomputation of it. See
+     * {@link #requireBalanceUnmoved}, which carries the reasoning; the opening figure it checks
+     * against is stamped onto the line when the document is written, by
+     * {@link #stampOpeningBalance}. Because the balance is then known to be the one on the line,
+     * the figures written back to the line are the raised figures, and the document, the count
+     * sheet and the ledger all describe one movement.
      *
      * <p>The approval is refused up front when the person approving is the one who raised the
      * document, unless they hold the break-glass role, in which case the posting is allowed and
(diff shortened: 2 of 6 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Added a second commit and rebased onto `development` (the GRN/PO reconciliation landed; no Java overlap, and the `docs/openapi.json` merge verifies clean with `./gradlew openApiSnapshot`).

**What the second commit fixes, found by probing my own change rather than by review.** The guard as first written compared every line against the live balance. A count sheet may split one shelf's variance across a line per reason, so much damaged and so much lost, and both lines are raised against the same opening figure. The first line moved the row, so the second was then refused for drift the document had itself just created, and since the whole approval is one transaction it rolled back: a document that used to post correctly stopped posting at all.

`approve` now tracks what it has already posted per `(material, location)` balance row and discounts it before comparing. The guard then measures stock that moved underneath the document, which is what it is for. `aShelfWhoseVarianceIsSplitAcrossSeveralLinesPostsAllOfThem` pins it: remove the discount and it fails.

That makes six tests that fail without the change and pass with it.

**Merging unreviewed.** CodeRabbit reported "Review rate limited" on this PR and per the policy set today no review agent was dispatched in its place, so this goes in on green CI plus my own reading. Recording that here so it is on the record rather than implied.

**One consequence a reviewer should know about.** Two lines counting the *same* shelf (both carrying a `physicalQuantity` for the same material and location) still behave as before: the first posts and the second comes out at no movement and is skipped. That document is incoherent either way, and refusing it up front is a separate validation rather than something to fold into this guard.

Issue #648, opened 2026-08-31
Goods receipts and site transfers post stock and reconcile nothing: the statuses on both are labels


abhijithanandan:
Two documents in the materials module read as controls and are not. A goods receipt and a site transfer both post stock and both leave the thing they are supposed to reconcile untouched. Filing them together because the shape is one shape: a document with a status field, a parent it references, and no code anywhere that compares the two.

## A goods receipt neither checks its purchase order nor updates it

`PurchaseOrderItem.receivedQuantity` is written exactly twice in `src/main`, and both are `0` at creation:

- `purchaseOrder/PurchaseOrderService.java:344` — `item.setReceivedQuantity(0);`
- `purchaseOrderItem/PurchaseOrderItemService.java:76` — `poItem.setReceivedQuantity(0);`

Nothing else writes it. No JPQL or native `UPDATE` touches the column; `PurchaseOrderItemRepository` has one `@Query` and it is a `SUM(poi.totalPrice)`. The only two reads are a delete guard at `PurchaseOrderItemService.java:179` (`if (item.getReceivedQuantity() != null && item.getReceivedQuantity() > 0)`) and DTO mapping at `:231`. So the delete guard is permanently satisfied: a purchase order line can always be deleted, however much has been received against it, because the field it consults never moves off zero.

`GoodsReceivedNoteService` resolves the purchase order only to hang it on the header. `purchaseOrder.getItems()` is never called. The item loop takes the ordered quantity from the request:

```java
grnItem.setOrderedQuantity(itemDto.getOrderedQuantity());
grnItem.setReceivedQuantity(itemDto.getReceivedQuantity());
```

So `orderedQuantity` on a receipt line is whatever the browser typed, not what the order says, and even a display-level cross-check is absent. There is no approve or receive step to put a check in: `GoodsReceivedNoteControllerWeb` exposes a create, a patch and reads. The only downstream effect is `GrnCreatedEvent` to `InventoryEventListener.handleGrnCreated`, which writes an `InventoryTransaction` and raises stock. It never mentions the purchase order.

[...]

PR #659, opened 2026-08-31
Post a goods receipt back onto the purchase order it cites

abhijithanandan:
Closes the goods-receipt half of #648. The site-transfer half is specified in a separate issue rather than half-built, because it changes what the document means to the people using it.

## What was wrong

`PurchaseOrderItem.receivedQuantity` was written `0` twice at creation and never again. Nothing in `src/main` compared a goods receipt with the order it cited. Three consequences followed from the one gap: an order for ten read as fully outstanding after ten thousand had been received against it; the delete guard at `PurchaseOrderItemService:179` was permanently satisfied, so any line could always be deleted however much had arrived; and `PARTIALLY_RECEIVED` and `FULLY_RECEIVED` were statuses no code could reach, so an order only moved when somebody patched it by hand from a screen showing zeroes.

Live staging shows it exactly. One purchase order, `PO-2026-000001`, one line for material 3 ordered 1, received 0. Five goods receipts cite it: four are for a material that is not on the order at all, and the fifth received 5 against an order for 1. The order still reads `DRAFT` with nothing received.

## The three decisions

**Over-receipt is refused by default and recorded on request.** Neither extreme survives contact with a site. Accepting any quantity silently is the behaviour being replaced. Refusing outright is worse than it looks: the supplier who delivers 105 bags against an order for 100 has left 105 bags on the site whether the system likes it or not, and a receipt that cannot be filed puts them outside the stock ledger, which is exactly the condition SA-563 showed makes a balance unexplainable. So the default refuses and names the line, the quantity ordered, the quantity already received and the quantity now offered, which is enough for a person to recognise a typed digit; and a payload that sets `allowOverReceipt` records the excess and marks the note `overReceiptAcknowledged`, so the document explains itself to whoever reads it six months on. Whatever the answer, `receivedQuantity` now reflects what arrived.

A receipt line for a material that is **not** on the order reconciles nothing, which is the truth about it, and posts stock exactly as before. Refusing it would be wrong: a lorry can carry an item ordered separately or supplied free.

**The order's status moves on its own, and the actor is the receipt.** `PARTIALLY_RECEIVED` when something is outstanding, `FULLY_RECEIVED` when every line is met. A move like that has no person behind it, so inventing one would be worse than admitting it. It is written to the shared status trail (#565) under a new `StatusTransitionSource.SYSTEM` whose note names the goods receipt that caused it; that receipt carries `receivedBy`, which is as close to a person as a derived transition honestly gets. Manual status changes are now filed against the person who made them too, otherwise the trail would read as though the system made every change an order ever went through, and existing orders get one `BASELINE` entry the way projects did in 073. `GET /api/v1/purchase-orders/web/{id}/status-history` reads it.

`DRAFT` and `CANCELLED` orders keep their status while still recording what arrived. A received quantity is a fact; a status is a lifecycle claim, and a receipt must not quietly approve an order nobody approved or reinstate one somebody cancelled.

**Site transfers are specified, not built.** See the linked follow-up issue.

## Ledger safety

No new inventory movement shapes. The stock posting path is untouched. The only new writes are `purchase_order_item.received_quantity`, `purchase_order.status`, `status_transition`, and the new `goods_received_note.over_receipt_acknowledged`. The one change to when movements happen is that a refused over-receipt now writes none, and the caller is told which line and given the way past it. Reconciliation runs inside the create transaction and before `GrnCreatedEvent`, so a refusal takes the note with it.

Concurrency: the order's lines are taken with `SELECT ... FOR UPDATE` ordered by id, so two receipts against the same order queue rather than each judging themselves against the figure that stood before the other. Without it both would pass the over-receipt check and the second write would erase the first, leaving an order reading as under-received while the ledger held both deliveries.

## Verified against live staging

- `status_transition.source` is `VARCHAR NOT NULL` with no check constraint, so `SYSTEM` stores without a schema change.
- 28 `PROJECT` `BASELINE` rows exist, no `PURCHASE_ORDER` rows, so changeset `084-02`'s precondition passes and seeds the one existing order.
- `goods_received_note.over_receipt_acknowledged` does not exist, so `084-01` adds it.
- Under this change, `GRN-2026-000002` (5 against an order for 1) would have been refused with the figures named. The four receipts for the unordered material would post as before.

## Tests

22 new, all failing without the change (a mutation run neutering `applyReceipt` fails 17 of them; the rest are the service wiring and trail tests, which cannot compile against the old code because the collaborator does not exist).

| Test | What it pins |
|---|---|
| `addsTheReceivedQuantityToTheMatchingOrderLine` | the field that never moved off zero |
| `addsToTheQuantityAlreadyReceivedRatherThanReplacingIt` | successive deliveries accumulate |
| `refusesAnOverReceiptThatWasNotAcknowledged` | 10,000 against 10 is stopped, with the figures |
| `countsWhatIsAlreadyReceivedWhenJudgingAnOverReceipt` | the check is cumulative, not per note |
| `recordsAnOverReceiptThatWasAcknowledged` | reality can still be recorded |
| `movesTheOrderToPartiallyReceivedWhileSomethingIsOutstanding` | the status is reachable |
| `movesTheOrderToFullyReceivedWhenEveryLineIsMet` | and so is the other one |
| `recordsTheDerivedMoveAsASystemTransitionNamingTheReceipt` | the actor question |
| `leavesADraftOrderWhereItIsWhileStillRecordingWhatArrived` | quantities are facts, statuses are claims |
| `leavesACancelledOrderWhereItIs` | a receipt does not reinstate an order |
| `reconcilesNothingForAReceiptLineThatIsNotOnTheOrder` | unmatched lines are honest, not refused |
| `takesTheOrderedQuantityOnTheReceiptLineFromTheOrder` | the order is the authority on what was ordered |
| `spreadsAQuantityAcrossTwoOrderLinesForTheSameMaterial` | split lines |
| `putsAnAcknowledgedExcessOnTheLastLineSoTheLinesStillSumToWhatArrived` | the lines sum to reality |
| `sumsTwoReceiptLinesForTheSameMaterialBeforeJudgingTheOverReceipt` | the check cannot be split around |
| `postsTheReceiptBackOntoTheOrderItCites` | the wiring exists |
| `passesTheAcknowledgementFromThePayloadThrough` | the flag reaches the check |
| `raisesNoStockWhenTheOrderRefusesTheDelivery` | no event on refusal |
| `marksTheNoteWhenTheOverReceiptWasLetThrough` | the document says so |
| `leavesTheNoteUnmarkedWhenTheReceiptStaysWithinTheOrder` | and does not say so otherwise |
| `recordsAStatusChangeSomebodyMadeAgainstThatPerson` | manual moves keep their actor |
| `refusesToReadTheTrailOfAnOrderThatIsNotInThisOrganization` | tenant scoping on the new read |

Full suite on `.116`: 1679 tests, 0 failures. Test heap 570 MB live of 2048 MB (28%), 8 contexts cached of 8. `docs/openapi.json` regenerated with `-PupdateOpenApiSnapshot`.

## Contract

Additive for `echno-core`: `allowOverReceipt` on the GRN creation payload, `overReceiptAcknowledged` on the response, and a new status-history endpoint. The one behaviour change a client can see is a 400 where an unacknowledged over-receipt previously returned 201. The web app needs a way to offer the acknowledgement; filed separately.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -5639,8 +5639,13 @@
         "type": "object"
       },
       "GoodsReceivedNoteCreationDto": {
-        "description": "Payload to record a goods received note. The received quantities on the items are posted into stock at the given storage location. The GRN number is allocated by the server and returned on the created note; it is not part of this payload.",
+        "description": "Payload to record a goods received note. The received quantities on the items are posted into stock at the given storage location and added to the matching lines of the purchase order, whose status then follows from the quantities received against it. The GRN number is allocated by the server and returned on the created note; it is not part of this payload.",
         "properties": {
+          "allowOverReceipt": {
+            "description": "Set this when the delivery really did exceed the purchase order and the site took it anyway. Without it, a line that would take a material past the quantity ordered is refused, which is what catches a mistyped quantity. With it the excess is recorded and the note is marked as an acknowledged over-receipt. It has no effect on a receipt that stays within the order.",
+            "example": false,
+            "type": "boolean"
+          },
           "deliveryChallanNumber": {
             "description": "Vendor delivery challan number.",
             "example": "DC-88213",
@@ -5750,6 +5755,11 @@
               "$ref": "#/components/schemas/GrnItemDto"
             },
             "type": "array"
+          },
+          "overReceiptAcknowledged": {
+            "description": "True when this receipt took a material past the quantity its purchase order asked for and the excess was acknowledged on the payload.",
+            "example": false,
+            "type": "boolean"
           },
           "projectId": {
             "description": "Project the goods were received for.",
@@ -5878,7 +5888,7 @@
             "type": "string"
           },
           "orderedQuantity": {
-            "description": "Quantity ordered for this material.",
+            "description": "Quantity ordered for this material. Where the material is on the cited purchase order, the value returned is the order's, not the one submitted: the order is the authority on what was ordered.",
             "example": 100,
             "format": "int32",
             "minimum": 0,
@@ -43022,7 +43032,7 @@
         ]
       },
       "post": {
-        "description": "Records a goods receipt from a vendor with its line items. The received quantities are posted into stock at the given storage location as GRN transactions.",
+        "description": "Records a goods receipt from a vendor with its line items. The received quantities are posted into stock at the given storage location as GRN transactions, and added to the matching lines of the purchase order the note cites, whose status then follows from the quantities received against it. A line that would take a material past the quantity ordered is refused unless the payload sets allowOverReceipt, and nothing is posted when it is refused.",
         "operationId": "createGrn",
         "requestBody": {
           "content": {
@@ -43053,7 +43063,7 @@
                 }
               }
             },
-            "description": "Validation failed on the request body"
+            "description": "Validation failed on the request body, or a line would take a material past the quantity its purchase order asked for and the over-receipt was not acknowledged"
           },
           "402": {
             "content": {
@@ -43719,7 +43729,7 @@
         ]
       },
       "post": {
-        "description": "Creates a GRN recording materials received against a vendor with its line items.",
+        "description": "Creates a GRN recording materials received against a vendor with its line items. The received quantities are posted into stock and added to the matching lines of the purchase order the note cites, whose status then follows from the quantities received against it. A line that would take a material past the quantity ordered is refused unless the payload sets allowOverReceipt, and nothing is posted when it is refused.",
         "operationId": "createGrn_1",
         "requestBody": {
           "content": {
@@ -43750,7 +43760,7 @@
                 }
               }
             },
-            "description": "A field failed validation"
+            "description": "A field failed validation, or a line would take a material past the quantity its purchase order asked for and the over-receipt was not acknowledged"
           },
           "402": {
             "content": {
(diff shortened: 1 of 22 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Merged unreviewed. CodeRabbit returned "Review rate limited" on this PR, so per the standing rule this went in on green CI plus my own read of the diff rather than waiting the quota out or standing up a substitute reviewer.

What I checked by hand, since nothing else did:

- The over-receipt check is cumulative per material across the whole order, so it cannot be split around by sending two receipt lines for the same material in one note. Pinned by `sumsTwoReceiptLinesForTheSameMaterialBeforeJudgingTheOverReceipt`.
- The lines are taken with `SELECT ... FOR UPDATE` ordered by id, so two concurrent receipts queue rather than each reading the figure that stood before the other, and two callers take the rows in the same sequence so they cannot deadlock against each other.
- Reconciliation runs inside the create transaction and before `GrnCreatedEvent` is published, so a refusal rolls the note back and the after-commit listener never runs. `raisesNoStockWhenTheOrderRefusesTheDelivery` pins it.
- `derive` reads the same `PurchaseOrderItem` objects the allocation mutated, so the status is worked out from the post-receipt figures rather than the pre-receipt ones.
- `FULLY_RECEIVED` is in the movable set, so an order whose ordered quantity is later raised past what has arrived moves back to `PARTIALLY_RECEIVED` and the move is recorded. That is the honest outcome rather than a stuck status.
- No new inventory movement shapes. The stock path is untouched; the only new way for a movement not to happen is a refusal that names the line and offers the way past it.
- `status_transition.source` is `VARCHAR NOT NULL` with no check constraint on live staging, so the new `SYSTEM` value needs no schema change, and `084-02`'s precondition was verified against the live table before writing it.
- Changelog number re-checked against `origin/development` immediately before committing: 083 was the maximum, 084 was free.

Issue #645, opened 2026-08-31
The OpenAPI document declares no field nullable, on any of its 3069 properties


abhijithanandan:
Raised by review on PR #639, where a new nullable `verifiedById` was flagged for being declared `"type": "integer"` rather than the OpenAPI 3.1 union `["integer", "null"]`. The finding is right about the semantics and wrong about the scope, so it is filed here rather than patched on that one field.

## What is actually true of the document

`docs/openapi.json` is OpenAPI 3.1.0 and carries **no nullability information at all**:

- 0 of 3069 schema properties use a type union;
- 0 use the `nullable` keyword;
- 0 use `anyOf` or `oneOf` to admit null.

Meanwhile a large share of those properties are null in practice. Three examples that sit next to each other: `MovementRecordDto.verifiedBy`, `verifiedAt` and `verifiedById` are all null on every unverified movement, which is most rows, and all three are declared as plain scalars. `AttendanceResponseDto.approvedBy` / `approvedById` are the same fields in the same role, same treatment.

So a consumer reading this document is told every response field is always present and non-null, and that is wrong for a large fraction of them.

## Why it is not a one-line fix

The document is generated from the code and CI-verified against it (`openApiSnapshot`, `-PupdateOpenApiSnapshot`), so it cannot be edited directly. Making a field nullable means annotating the DTO, which is a decision to take once and apply across the DTO layer rather than wherever a reviewer happens to look. Patching single fields as they come up would leave the document in a worse state than the current uniform silence, because a reader would then have to guess whether an undeclared field is non-null or merely unannotated.

## What to decide

1. Whether springdoc should infer this, for instance from `Optional`, JSpecify or `@NotNull` presence, rather than from a hand-added `@Schema(nullable = true)` on several hundred fields.
2. Whether the response DTOs are annotated wholesale or only where a consumer actually depends on it.
3. Whether a build check should hold the decision, the way the snapshot check holds the document itself.

## Who this affects today

Not the web client directly: `echno-core` is hand-maintained with no codegen, and its zod schemas already model null with `nullableString` and friends. It does affect the contract check `echno-core` runs against this document (the one that found 19 live bugs, core #57), which currently cannot use the document to tell a genuinely optional field from a required one.

Low priority, but worth a single pass rather than accumulating one-field exceptions.

abhijithanandan:
## Why nothing was nullable

Measured rather than guessed. The annotation was never the missing piece: it was there to be used and produced no output.

[...]

PR #661, opened 2026-08-31
Let the document say a field may be null, which under 3.1 it never could

abhijithanandan:
Closes part of #645. The wiring is done; the remainder is specified in a comment on the issue rather than left implied.

## The finding

`docs/openapi.json` said nothing about null on any of its 3069 properties, and the reason is not that nobody annotated anything.

springdoc 2.8.6 defaults `springdoc.api-docs.version` to `OPENAPI_3_1`, which is why the document reads `"openapi": "3.1.0"` with no configuration saying so. In 3.1 nullability is a type union rather than the 3.0 `nullable` keyword. swagger-core 2.2.29 reads `@Schema(nullable = true)` onto the model and then discards it: `Schema31Mixin` marks `getNullable()` ignored and writes `types` instead, with no translation between them. Running the resolver directly over an annotated field shows both halves:

```
annotatedNullable   specVersion=V31 nullable=true types=[string]
Json30 render: { "description": "...", "nullable": true }
Json31 render: { "type": "string", "description": "..." }
```

So the one annotation an author would reach for resolved, was held in memory, and produced no diff. Nothing else fills the gap either: `Optional`, JSpecify `@Nullable` and boxed types carry nothing into the schema, and `@NotNull` produces `required`, which is presence rather than nullability. `@Column(nullable = false)` is real information but it sits on the entities, and the DTOs are separate Lombok classes filled by MapStruct.

## The change

`NullableSchemaCustomizer` runs on the assembled document and writes the flag in the form 3.1 uses: a typed property gains `"null"` as a second member of its type, and a bare `$ref` becomes `anyOf` of the reference and the null type, since a reference has no type to add to. A document served as 3.0 is left alone, because the keyword is written out as it stands there.

57 response properties across 25 schemas are then marked. The selector is deliberate rather than opportunistic: they are the properties whose own `description` already said the value may be null, written as prose because the schema could not hold it. Each was established before this change rather than by it, and six were re-checked against the column or the mapper behind them. Nothing is newly declared non-null, so an unmarked field says exactly what it said before, which is nothing.

## Proof

`OpenApiNullabilityTest` reads the committed document and fails when an annotated field is not described as admitting null. Run against `docs/openapi.json` as it stood before this change it reports all 57:

```
OpenApiNullabilityTest > everyNullableFieldIsDeclaredNullable FAILED
    java.lang.AssertionError at OpenApiNullabilityTest.java:96
```

It reads the committed copy rather than booting an application, since `OpenApiSnapshotTest` already holds that copy to the served document in its own task.

`NullableSchemaCustomizerTest` covers the shapes the DTO layer produces: scalar, reference, collection, a model that contains itself, an inline parameter and request body, and the 3.0 document that is left alone.

## Verified on lab `.116`

- `./gradlew test` green (8m 37s, heap 571 MB live of 2048 MB, 28%)
- `./gradlew openApiSnapshot` green after the rebase onto `development`, so the committed document still matches the code
- the document diff is 54 type unions plus 3 `anyOf` conversions and nothing else

## Not in this change

The request side (1094 properties in schemas reachable from a request body) needs null told from absent, and the partial-update surfaces need one decision per class rather than per field: `TaskUpdateFieldsDto` says at class level that a field sent as null is cleared while `categoryId` on the same class cannot be cleared, and `LeaveRequestUpdateFieldsDto.startDate` rejects null outright. The response fields nobody has written down yet (1975 properties in response-only schemas) are per-field reading work, tractable module by module. Both are set out on #645.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,27 @@ runs the check on every pull request, so a stale copy fails the build rather tha
 The task needs Docker (it starts CockroachDB through Testcontainers, as the test suite does) and
 runs in a JVM of its own, which is why it is not part of `test` or `check`.
 
+### Saying that a field may be null
+
+A field that can come back null is marked on the DTO:
+
+```java
+@Schema(description = "Finance customer the project is billed to. Null when no client is set.",
+        example = "6b1e9c22-9f8a-4a1b-9c0e-1d2f3a4b5c6d", nullable = true)
+private UUID customerId;
+```
+
+The document is OpenAPI 3.1, where that is written as a type union rather than a `nullable` flag,
+so the property comes out as `"type": ["string", "null"]`. swagger-core reads the annotation but
+does not translate it under 3.1, so `NullableSchemaCustomizer` does the conversion and
+`OpenApiNullabilityTest` fails if an annotated field is not described that way in the committed
+document. Before that existed the annotation produced no output at all, which is why nothing in
+the document was marked nullable for as long as it has been published.
+
+Mark only what you have checked. A field the document declares non-null is a field a client is
+entitled to parse without a guard, so leaving one unmarked is recoverable and marking one wrongly
+is not.
+
 Serving the document from a deployment is a separate decision: `/v3/api-docs` and the Swagger UI
 answer 401 unless `SWAGGER_PUBLIC_ACCESS=true` is set for that environment.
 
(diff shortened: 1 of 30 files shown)


abhijithanandan:
@coderabbitai review

PR #663, opened 2026-08-31
Answer which materials have reached their reorder level

### LINKED ISSUE #652: Reorder levels raise nothing: no job, no notification, no query, and the only 'Low Stock Alert' is a badge computed in the browser
A reorder level is stored on every material and, at the material's location, on `MaterialLocationThreshold`. Nothing on the server ever compares it to stock.

Every occurrence in `src/main` is a write, a DTO field or a column definition:

- `material/Material.java:73-74` — `@Column(name = "reorder_level") private Double reorderLevel;`
- `material/threshold/MaterialLocationThreshold.java:63-64` — the per-location copy
- `MaterialService.java:110` (create) and `:266-267` (patch), `MaterialLocationThresholdService.java:103` (set) and `:136` (copy to DTO)
- `MaterialDto`, `MaterialCreationDto`, `MaterialUpdateDto`, `MaterialLocationThresholdDto`, `MaterialLocationThresholdUpsertDto`
- Liquibase: `v3.0/008-add-stock-levels-to-material.xml`, `v4.0/045-*`

No repository query mentions it. `MaterialRepository` has no `@Query` comparing stock to a threshold. No `@Scheduled` method touches materials at all: the only schedulers in the codebase are chat streaming and heartbeat, the subscription and RPT cache statistics, and the two compliance ones. There is no notification package in the backend.

## Where the alert actually lives

`MaterialDto` ships both `currentStock` and `reorderLevel`, and the web app computes the comparison in the browser, in six places:

- `features/materials/components/materials-insights-row.tsx` filters `m.currentStock <= m.reorderLevel` and renders the count under a heading reading **"Low Stock Alert"**
- `features/materials/components/materials-dashboard-table.tsx` does the same and offers a `LOW_STOCK` filter
- `material-list-item.tsx`, `material-overview-tab.tsx` and `material-stock-by-location-tab.tsx` each colour a badge

So the product has a low-stock alert. It is a number on a dashboard card, computed over whatever page of materials that browser happened to load, visible only while somebody has the Materials screen open. Nobody is told anything. Nothing arrives when the level is crossed at two in the morning, or on a Sunday, or on a site whose storekeeper does not open that screen.

The wording is the part that will catch somebody out. A card headed "Low Stock Alert" reads as a thing that alerts. Setting a reorder level on a material is an action a person takes expecting a consequence, and there is none.

## What it needs before it can be built

- **What is raised, and to whom?** An in-app notification (there is no notification subsystem to hang it on), an email, a task, an automatically drafted indent. Each is a different amount of work and the last one changes what the procurement flow means.
- **When?** On the transition across the level, which needs somewhere to remember that it already fired, or on a schedule, which will re-raise the same material every night until somebody orders. The first is right and the second is what usually gets built.
- **Per organization or per project?** Stock is held per project and per location, and there are two thresholds: the material's own and the per-location one. A material below its org-wide level while one site has plenty is a different event from a site running out.
- **Who is the recipient in a model with no stores role?** Most of Resources is `system-admin` only (see the role-model issue filed alongside this), so today the only person who can act on a low-stock alert is an organization administrator.
- **The scheduling problem.** `@Scheduled` elects no leader here; `ComplianceRuleSweep` documents this at length and works around it with a shared spend cap. Anything periodic added for reorder levels inherits the same problem and should read that class first.

Filing rather than building because the answer to the first question determines everything else, and because a nightly job that emails an administrator about forty materials is the version that gets muted in a week.

Found while writing the materials and inventory user guide (echno-web #371). Verified against `development` at the time of filing.

abhijithanandan:
Closes #652.

A reorder level is stored on every material and, at a material's location, on `MaterialLocationThreshold`. Nothing on the server has ever compared it to stock. The only low-stock signal in the product is a badge the web console computes in the browser, over whatever materials that page had loaded, visible while somebody has the screen open.

## What the levels look like on real data

Read on `echno.in`, read only:

| | |
|---|---|
| materials | 15 |
| with a reorder level set | 14 |
| **at or below it, organization wide** | **9** (8 strictly below) |
| per-location threshold overrides in the whole database | 0 |
| (material, project, location) balances at or below the material's level | **14**, across 7 projects |

`TNT Steel` holds 60 against a level of 30, so the console counts it as healthy. It is at 1, 18, 1, 7 and 1 unit at five separate project locations. The aggregate the console compares against is the one figure that cannot see a site running out.

The level itself is authored, not computed: `material-form.tsx` derives it in the browser as `safetyStock + 2*ltc` and posts the result as a plain number. `ltc` is persisted only since #078, and is NULL on all 15 existing rows, so the backend cannot re-derive it.

## What this adds

`GET /api/v1/materials/low-stock`, and the `/web` twin. Paginated through `PageQuery`, `@PreAuthorize` on both, `docs/openapi.json` regenerated.

Three scopes, matching the ones `GET /materials/{id}/stock` already reads at, because the answer differs at each:

- **neither id**: totalled across the organization, every catalogue material a candidate.
- **`projectId`**: totalled over that project's locations, only materials the project holds.
- **`projectId` + `storageLocationId`**: that one location, and here the `MaterialLocationThreshold` override replaces the material's global level, for the reorder level and the minimum order quantity both.

Decisions a second implementation would get wrong, each pinned by a test:

- **At or below**, not below. The console's six comparisons are all `<=`, and a reorder level is the level at which you reorder.
- **Organization scope left-joins** the stock rows, so a material stocked nowhere is reported rather than dropped. It is the most depleted material there can be.
- **Project and location scope inner-join**, so the catalogue is not reported as out of stock at every project that never carried it.
- **Ordering by the fraction of the level that is missing**, since an absolute shortfall is measured in the material's own unit. A level of zero sorts as fully depleted rather than dividing by zero, which the database would raise on.
- **A null level is never reported; a zero counts as set**, matching the console.
- **A project or location that does not exist is a 404**, and one belonging to another project is refused through the existing `StorageLocationScope` rule. An empty page here reads as "nothing has run out", and that must never be the answer to a question about somewhere that does not exist.
- **`totalElements` is the count**, so the dashboard card can ask for `pageSize=1` and read the true number without pulling a row.

## What this deliberately does not add

A scheduled sweep. Alert delivery is failing end to end on both stagings (echno-deployment#66: the Discord webhook targets a forum channel and every send is refused), and the backend has no notification package, no mail sender and no in-app inbox. A job emitting into that would reproduce that failure per material per tenant per night. There is also no stores role, so the only recipient today is an organization administrator.

The design for one, following `ComplianceRuleSweep` and off by default for the same reasons, is written up on the issue along with the three product decisions that gate it.

## Tests

Every query decision was checked by mutating it and confirming the matching test fails.

| Mutation | Test that fails |
|---|---|
| organization scope `LEFT JOIN` to `JOIN` | reports a material that has no stock row anywhere |
| `<=` to `<` | reports a material sitting exactly on its reorder level |
| severity ordering unguarded against a zero level | a reorder level of zero sorts as fully depleted rather than dividing by zero |
| severity ordering by absolute shortfall | orders by the fraction of the level that is missing, not by quantity |
| location scope ignores the override | storage-location scope applies the location's override |
| organization count query changed | organization scope counts every match, not just the page |

`LowStockRepositoryIT` (12 cases, real CockroachDB, shares the existing `@DataJpaTest` context) and `LowStockServiceTest` (9 cases, plain Mockito). 108 tests green across `material`, `common.pagination` and `architecture`; test heap 278 MB of 2048 MB.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -9525,6 +9525,69 @@
         },
         "type": "object"
       },
+      "LowStockMaterialDto": {
+        "description": "A material whose stock has reached or fallen below the reorder level in force at the scope that was asked about.",
+        "properties": {
+          "currentStock": {
+            "description": "Quantity on hand at the scope that was asked about. Zero when the material has no stock there at all.",
+            "example": 3501,
+            "format": "double",
+            "type": "number"
+          },
+          "materialId": {
+            "description": "Material id.",
+            "example": 12,
+            "format": "int64",
+            "type": "integer"
+          },
+          "materialName": {
+            "description": "Name of the material.",
+            "example": "TMT Bar 12mm",
+            "type": "string"
+          },
+          "moq": {
+            "description": "The minimum order quantity in force at that scope, as the smallest quantity a purchase can be raised for. Null when none is recorded.",
+            "example": 10000,
+            "format": "double",
+            "type": "number"
+          },
+          "projectId": {
+            "description": "Project the stock was counted within, null when the whole organization was counted.",
+            "example": 5,
+            "format": "int64",
+            "type": "integer"
+          },
+          "reorderLevel": {
+            "description": "The reorder level that was applied. At storage-location scope this is the location's override when it has one, and the material's own level otherwise.",
+            "example": 11000,
+            "format": "double",
+            "type": "number"
+          },
+          "shortfall": {
+            "description": "How far below the reorder level the stock is, never negative. Zero for a material sitting exactly on its level.",
+            "example": 7499,
+            "format": "double",
+            "type": "number"
+          },
+          "sku": {
+            "description": "Stock keeping unit code.",
+            "example": "TMT-12MM-001",
+            "type": "string"
+          },
+          "storageLocationId": {
+            "description": "Storage location the stock was counted at, null unless one was asked about.",
+            "example": 2,
+            "format": "int64",
+            "type": "integer"
+          },
+          "unit": {
+            "description": "Unit of measure for the material.",
+            "example": "kg",
+            "type": "string"
+          }
+        },
+        "type": "object"
+      },
       "MaterialConsumptionCreationDto": {
         "description": "Payload to record material consumed against a project. Stock sufficiency is checked at the material and project level, or at the storage location level when one is given.",
         "properties": {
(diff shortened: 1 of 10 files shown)


abhijithanandan:
@coderabbitai review

PR #664, opened 2026-08-31
Publish the validation constraints the endpoints already enforce

### LINKED ISSUE #657: The cancel-voucher schema does not say its reason is required, so the contract check cannot
`CancelConstructionPaymentRequest.reason` is `@NotBlank`, and the published `docs/openapi.json` does not say so.

The schema PR #656 generated:

```json
"CancelConstructionPaymentRequest": {
  "properties": {
    "reason": {
      "maxLength": 1000,
      "minLength": 0,
      "type": "string"
    }
  },
  "type": "object"
}
```

No `required` array, and `minLength: 0`. So the document permits `{}` and permits `{"reason": ""}`, both of which the endpoint refuses at runtime.

## Why it is worth fixing rather than living with

`echno-core` reduces this document into `etc/backend-request-fields.json` and checks every outgoing write call against it (`bun run contract:report`). That check is what found nineteen live bugs on its first run. It reads `required` to decide whether a caller may omit a field, so today it would pass a cancellation that sends no reason at all, and the first thing to notice would be the 400.

It came up in review on echno-core PR #82, which added the client for this endpoint. Hand-editing the reduced JSON there is not the fix: it is generated, and `contract:refresh` would drop the edit on the next run.

## Where it comes from

The `@NotBlank` is on the field and the runtime honours it, so this is the OpenAPI generator not carrying the annotation across rather than a missing constraint. `@NotBlank` should produce `required: ["reason"]` plus `minLength: 1`; `minLength: 0` is what a bare `@Size(max = 1000)` would give. Worth checking whether the same is true of the other `@NotBlank` request fields in the document, since if the generator is dropping the annotation it is unlikely to be dropping it only here.

Not urgent. Nothing is broken at runtime: the endpoint refuses a blank reason whatever the document claims.

abhijithanandan:
Closes #657. Same shape as #661: a focused customization plus a test that reads the committed document and fails when an annotated field is not described correctly.

## The finding

`docs/openapi.json` described far less than the code refuses, in four ways, all of them from `ModelResolver.applyBeanValidatorAnnotations` in swagger-core 2.2.29:

| | count | cause |
|---|---|---|
| fields that reject null but were not in `required` | **127** | `required` is reached only through the `@NotNull` branch, so `@NotBlank` and `@NotEmpty` never put a property in it, although both reject null exactly as `@NotNull` does |
| properties publishing `minLength: 0` | **48** | `@Size` assigns `minLength` from `Size.min()`, which defaults to 0, and runs *after* `@NotBlank` has set it to 1 |
| `@Positive` / `@PositiveOrZero` with no numeric bound | **12** | not read at all |
| `@Email` with no `format` | **7** | not read at all |

`minLength: 0` is the worst of the four, because it is not silence: it positively permits the empty string the endpoint answers 400 for.

The earlier guess on the issue, that the generator drops `@NotBlank`, was wrong. `@NotBlank` alone does produce `minLength: 1`. It is the interaction with `@Size` that loses it, which is why only something seeing both annotations together can settle the value.

## Why a model converter and not a document customizer

Nullability survived resolution onto the assembled document, which is what let #661 work from an `OpenApiCustomizer`. These constraints do not. They are consumed as each property is resolved, and `required` is written on the parent at that same moment, so by the time a customizer runs a `maxLength: 255` with no `minLength` beside it is indistinguishable from a plain `@Size`, and a name missing from `required` is indistinguishable from a genuinely optional field.

`ValidationConstraintsModelConverter` delegates down the chain first and repairs afterwards. Running after rather than before is the point: swagger-core applies `@Size` after `@NotBlank`, so a converter writing `minLength: 1` on the way in would have it overwritten on the way out.

## What it will not say

A document that overstates a constraint is worse than one that omits it, because a client legitimately pre-validates against it and refuses input the server would have accepted. So:

- `@NotBlank` rejects a string of blanks and `minLength: 1` does not say so. Left understated on purpose; the rest needs a `pattern`, which would collide with a `@Pattern` already on the field.
- a constraint naming validation `groups` is skipped, since whether it applies depends on the invocation and a schema keyword applies to all of them. (Nothing in the repo uses groups today; the guard is for the next one that does.)
- `@Min`, `@Max`, `@DecimalMin`, `@DecimalMax`, `@Size` and `@Pattern` are left to swagger-core, which already writes them correctly. An existing bound is never loosened: `@Positive @Min(5)` keeps `minimum: 5`.
- `@Past`, `@Future`, `@Digits`, cross-field and custom class-level constraints have no keyword that means what they mean.
- constraints on controller method parameters go through springdoc's parameter builder rather than this resolver, so a `@NotBlank @RequestParam` is untouched.

`@Positive` is written as `exclusiveMinimum: 0` rather than `minimum: 1`, which would be right for an integer and wrong for the `BigDecimal` amounts most of these are.

## One thing the test found that was not a missing keyword

`CreateInvoiceRequest.LineRequest` and `PostJournalRequest.LineRequest` share a simple name, so only one of the two could be published. The journal line won, and `CreateInvoiceRequest.lines` pointed at a schema describing an account id with a debit and a credit. A client following the document sent a body the endpoint rejects. The invoice line is now `@Schema(name = "InvoiceLineRequest")`; the journal line keeps the name it had, so nothing that reads `LineRequest` today changes meaning. (`BalanceSheetReport.AccountLine` and `ProfitAndLossReport.AccountLine` collide too, but they are structurally identical and carry no constraints, so that one is harmless and left alone.)

## Proof

`OpenApiValidationConstraintsTest` reads the committed document and fails when an annotated field is not described the way the annotation behind it behaves. Run against `docs/openapi.json` as it stood before this change it reports all **195**:

```
OpenApiValidationConstraintsTest > everyConstrainedFieldIsDescribed FAILED
    java.lang.AssertionError at OpenApiValidationConstraintsTest.java:155

  127 not in required, although the constraint rejects null
   48 no minimum length or item count, although the constraint rejects empty
   12 no numeric bound, although the constraint rejects the wrong sign
    7 no email format
    1 annotated on a published schema, but the document has no such property
```

It reads the committed copy rather than booting an application, since `OpenApiSnapshotTest` already holds that copy to the served document in its own task.

`ValidationConstraintsModelConverterTest` runs 15 cases through the real `ModelConverters` chain rather than a hand-built schema, because the behaviour being fixed is an ordering one inside `ModelResolver`: a test that built the schema itself would not reproduce it. It covers each shape the DTO layer produces, plus the four things the converter must leave alone.

## The document diff, audited entry by entry

142 changes, and nothing else in the file:

- 48 `minLength` raised from 0 to 1 (no other value, in either direction)
- 127 names added to `required`, 58 on schemas that had no list at all
- 7 `exclusiveMinimum: 0` and 5 `minimum: 0`
- 7 `format: "email"`
- 1 new schema `InvoiceLineRequest` and the one `$ref` repointed at it

Nothing removed, no bound loosened, no property newly declared that the endpoint does not enforce. Every `@NotBlank` and `@NotEmpty` in the repo sits on a request-reachable schema, so nothing on the response side gained a `required` it cannot honour.

## What this does for the contract check

`echno-core` reduces this document into `etc/backend-request-fields.json` and reports `unsendable-required-field` against every write call site. That check is request-side and reads `required`, so it was working from the `@NotNull` fields alone. Measured over the reduced form:

- 219 operations take a request body. **115 of them named no required field at all; now 65.**
- total required names across operations: **232 → 410**, with 102 operations gaining at least one.

The `required: ["reason"]` CodeRabbit asked for on core PR #82 now comes out of `contract:refresh` rather than needing a hand edit the next refresh would undo.

## Verified on lab `.116`

- `./gradlew test` green (8m 55s, 591 MB live of the 2048 MB cap, 29%)
- `./gradlew openApiSnapshot` green after the rebase onto `development`, so the committed document still matches the code

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,35 @@ Mark only what you have checked. A field the document declares non-null is a fie
 entitled to parse without a guard, so leaving one unmarked is recoverable and marking one wrongly
 is not.
 
+### Saying what a field will be rejected for
+
+Validation constraints on a DTO are published, so writing one is also writing the contract:
+
+```java
+@NotBlank @Size(max = 1000) String reason
+```
+
+comes out as `"minLength": 1, "maxLength": 1000` with `reason` in the schema's `required`.
+
+Left to swagger-core it does not. `required` is written from `@NotNull` and nothing else, so a
+`@NotBlank` field read as optional; `@Size` assigns `minLength` after `@NotBlank` has set it and
+its `min` defaults to 0, so the pairing above published `minLength: 0`, which positively permits
+the empty string the endpoint refuses; and `@Positive`, `@PositiveOrZero`, `@Negative`,
+`@NegativeOrZero` and `@Email` were not read at all. `ValidationConstraintsModelConverter` fills
+those in and `OpenApiValidationConstraintsTest` fails when an annotated field is not described.
+
+It states only what the annotation itself guarantees. `@NotBlank` also rejects a string of blanks
+and `minLength: 1` does not say so, which is deliberate: a document that understates a constraint
+costs a client one rejected request, and one that overstates it makes the client refuse input the
+server would have taken. `@Min`, `@Max`, `@DecimalMin`, `@DecimalMax`, `@Size` and `@Pattern` are
+already published correctly and are left alone; `@Past`, `@Future`, `@Digits`, cross-field rules
+and constraints naming validation groups have no keyword that means what they mean, so the
+document says nothing about them.
+
+Two nested types with the same simple name cannot both be published, and the loser is silently
+replaced by the winner rather than reported. Give one of them a name of its own with
+`@Schema(name = "...")` when that happens.
+
 Serving the document from a deployment is a separate decision: `/v3/api-docs` and the Swagger UI
 answer 401 unless `SWAGGER_PUBLIC_ACCESS=true` is set for that environment.
 
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -446,6 +446,9 @@
             "type": "string"
           }
         },
+        "required": [
+          "name"
+        ],
         "type": "object"
       },
       "AssetDto": {
@@ -684,7 +687,7 @@
             "description": "Why the asset moved. Required: an entry with no stated reason is what makes a ledger unexplainable.",
             "example": "Mobilised to the Marina Heights site for the piling phase",
             "maxLength": 255,
-            "minLength": 0,
+            "minLength": 1,
             "type": "string"
           },
           "referenceNumber": {
@@ -720,6 +723,9 @@
             "type": "integer"
           }
         },
+        "required": [
+          "reason"
+        ],
         "type": "object"
       },
       "AssetMovementDto": {
@@ -1621,7 +1627,8 @@
           "movementTrackingEnabled",
           "photoRequiredOnCheckIn",
           "photoRequiredOnCheckOut",
-          "regularizationApprovalRequired"
+          "regularizationApprovalRequired",
+          "settingName"
         ],
         "type": "object"
       },
@@ -2063,10 +2070,13 @@
             "description": "What was wrong with the voucher.",
             "example": "Duplicate of CPMT-000118, raised twice for the same invoice",
             "maxLength": 1000,
-            "minLength": 0,
+            "minLength": 1,
             "type": "string"
           }
         },
+        "required": [
+          "reason"
+        ],
         "type": "object"
       },
       "CategoryCreationDto": {
@@ -2076,7 +2086,7 @@
             "description": "Short description of what the category covers.",
             "example": "TMT bars, wire mesh and other reinforcement materials",
             "maxLength": 255,
-            "minLength": 0,
+            "minLength": 1,
             "type": "string"
           },
           "icon": {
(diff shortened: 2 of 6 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Merged unreviewed: CodeRabbit answered "Review rate limited" and a second-opinion review agent is no longer dispatched on cost grounds, so this goes in on green CI plus the reasoning above.

What I checked myself, since nothing else did:

- The document diff was audited entry by entry rather than eyeballed. 142 changes: 48 `minLength` raised 0 to 1 and no other value in either direction, 127 names added to `required`, 7 `exclusiveMinimum: 0`, 5 `minimum: 0`, 7 `format: "email"`, one new schema and one `$ref` repointed at it. Nothing removed, no bound loosened.
- Every `@NotBlank` and `@NotEmpty` in the repo sits on a schema reachable only from a request body, so no response schema gained a `required` the server cannot honour. Checked by walking reachability from request bodies and from responses separately: 143 request-reachable schemas, 168 response-reachable, 4 in both, and none of the four carries one of these constraints.
- The converter never loosens an existing bound. `@Positive @Min(5)` keeps `minimum: 5`, which has its own case in the unit test.
- The only remaining simple-name collision after this change is `BalanceSheetReport.AccountLine` against `ProfitAndLossReport.AccountLine`. Those two records are structurally identical, differ only in their `example` text, and carry no validation constraints, so the collision writes nothing wrong. Left alone.

PR #669, opened 2026-08-31
Let a project manager read the balance, and compute the adjustment totals

### LINKED ISSUE #665: Stock adjustment header and line totals are still unchecked client values on a draft
Follow-up flagged in #658. The line figures are now the server's, the header totals are not.

`applyHeaderFields` copies both totals straight from the request:

```java
stockAdjustment.setTotalAdjustmentValue(dto.getTotalAdjustmentValue());
stockAdjustment.setTotalVarianceQuantity(dto.getTotalVarianceQuantity());
```

`approve` recomputes `totalVarianceQuantity` from what it actually posts. It never recomputes `totalAdjustmentValue`. So on a draft both are whatever the client sent, checked against nothing, and the value total is that for the life of the document.

This is looser than it looks now that #658 stamps the lines. A client computing the header total from the figures it sent is summing figures the server has since overwritten, so a draft can show a header variance that does not match the sum of its own lines. `StockAdjustmentLineItem.totalAdjustmentValue` has the same shape: `applyLineItems` stores it verbatim, and `stampOpeningBalance` recomputes the quantity beside it but not the value.

The ledger is never written from any of these, so this is a display looseness rather than an audit one. It is still the same class of problem as the `systemQuantity` #658 closed: a figure that is arithmetic over other figures should be computed where those figures are.

Suggested: derive all three in `applyLineItems` / `applyHeaderFields` — line `totalAdjustmentValue` from the stamped variance and the unit value, header `totalVarianceQuantity` and `totalAdjustmentValue` by summing the lines — and mark the three request fields ignored in the DTO schema the way `systemQuantity` now is. A line the balance cannot be read for keeps what was sent, for the same reason `stampOpeningBalance` does.

echno-web is being changed alongside this (tornotron/echno-web#383) to stop sending `systemQuantity` and `adjustmentQuantity`, and to omit the value totals rather than fabricate them when it cannot read the balance. Once the server derives them, the web can stop sending them entirely.

### LINKED ISSUE #666: A project manager can raise a stock adjustment but cannot read the balance it is stamped against
Found while wiring the stock adjustment form to the real balance (tornotron/echno-web#383).

`MaterialControllerWeb`:

```java
@GetMapping("/{id}/stock")
@PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin')")
```

`StockAdjustmentControllerWeb` create, update, approve and reject:

```java
@PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')")
```

So a project manager may raise a stock adjustment and may approve one, but may not read the balance the document is stamped against and the approval is checked against. Since #658 that balance is the whole substance of the document: the form has no honest figure to show that role, and no way to explain a refused approval before the refusal arrives.

It did not matter for the other consumers of this endpoint. Material consumptions and site transfers are `system-admin` end to end, so the callers already held the role.

`/materials/web/{id}/stock` is a read of one quantity for one material at one project and location. Widening it to `project-manager` is the smallest change; whether the wider material reads should move with it is the part worth deciding rather than assuming.

abhijithanandan:
Closes #666. Closes #665.

Both follow #658, which made a stock adjustment's line figures the server's and made approval refuse when the balance has moved. Neither touches that guard or the per-balance-row accounting behind it.

## #666: the role that raises the document cannot read what it is about

A project manager may raise, approve and reject a stock adjustment. Since #658 the balance the document is stamped with is the substance of it: the approval is refused when that balance has moved between raising and approving. Every read on `MaterialControllerWeb` and `StorageLocationControllerWeb` was `system-admin` only, so that role could not read the figure it is being asked to correct, could not list the materials to name one, and could not pick the location it counted.

The full shape, for the whole workflow rather than the one endpoint on the issue:

| What the workflow needs | Endpoint | Project manager, before | After |
|---|---|---|---|
| List and read the documents | `GET /stock-adjustments/web…` | yes (member or role) | unchanged |
| Create, edit, approve, reject, delete | `/stock-adjustments/web` writes | yes | unchanged |
| Pick the project | `GET /projects/web…` | yes (member or role) | unchanged |
| Pick the material | `GET /materials/web`, `/search`, `/{id}` | **403** | yes |
| Read the balance | `GET /materials/web/{id}/stock` | **403** | yes |
| Read the reorder level that explains it | `GET /materials/web/{id}/location-thresholds`, `/low-stock` | **403** | yes |
| Pick the location counted | `GET /storage-locations/web`, `/project/{id}` | **403** | yes |

So widening `/{id}/stock` alone would have left a balance lookup reachable for a material the caller could not name. The reads move together; every write on both controllers stays with `system-admin`, because creating a material, editing it, deleting it and moving its reorder thresholds are catalogue decisions rather than count decisions.

**Not the same defect as the project read inconsistency.** That one was inside a single controller: `ProjectControllerWeb` read on `isMemberOfCurrentTenant()` while its writes read on `hasAnyOrgRole(...)`, so a role holder without `ORG_MEMBER_` could write and not read. It was fixed in `eec6c37` by making the read the union of the two, and `ProjectControllerWeb` reads that way on `development` today. This is the other axis: the guard on `StockAdjustmentControllerWeb` is internally consistent, and the read it depends on sits in a different module gated a tier higher.

**Independent of #650, not a workaround for it.** #650 asks what role a storekeeper should hold to *write* goods receipts, when the only role that can is `system-admin`. Whatever it lands on, `project-manager` will still write stock adjustments and will still need these reads. Nothing here widens anything to plain members, so the two-tier model #650 describes is left as it found it. The ten other Resources controllers are untouched.

## #665: the header totals

`applyHeaderFields` copied `totalVarianceQuantity` and `totalAdjustmentValue` straight from the request, and `applyLineItems` stored the line's `totalAdjustmentValue` verbatim. Once #658 began stamping each line's variance from the balance, a client computing the header from the figures it sent was summing figures the server had already replaced, so a draft could assert a total that did not match the sum of its own lines.

The ledger is written from the lines, so this is a document misreporting itself rather than a wrong stock figure. It is still the shape #658 closed on `systemQuantity`: a number the server accepts and never checks is one somebody eventually relies on.

**Computed, not guarded**, following #658's reasoning. A guard would have to refuse a document whose header disagreed with its lines, which fires on a client that rounded differently rather than on anything worth stopping, and a draft is where a person is still allowed to be halfway through.

- `stampLineValue` derives the line's value from the stamped variance and the unit value. A line with no unit value keeps what was sent, the same rule `stampOpeningBalance` uses for a line with no balance to read: the cost such a line posts at is the running average at the moment of posting, which no draft can state.
- `stampHeaderTotals` sums both header figures from the lines once they are attached, on create and on update alike. A document with no lines totals zero rather than keeping a number it cannot support.
- `approve` restates both from what it actually posted, and writes each posted line's value at the cost the ledger entry was written with. It already did this for the variance and not for the value, which is how a posted document could carry a money figure from a draft that no longer described it. A line whose movement comes out at nothing contributes nothing to either.

Values are rounded to the two decimal places `total_adjustment_value` is declared with, so a document read back straight after being written has not changed.

The three request fields are documented as derived, the way `systemQuantity` now is, and kept on the schema rather than marked read-only, so the field stays where echno-web sends it until the web stops sending it (tornotron/echno-web#383).

## Tests

Nineteen new assertions fail against `development` and pass here.

| Test | Without the fix |
|---|---|
| `MaterialControllerWebAuthzTest.aProjectManagerMayReadTheCatalogueAndItsBalances` (9 paths) | 403 on every read, the balance lookup included |
| `StorageLocationControllerWebAuthzTest.aProjectManagerMayReadTheLocations` (5 paths) | 403, so the counted shelf cannot be named |
| `StockAdjustmentTotalsTest.aLineValueIsRecomputedFromTheStampedVarianceAndTheUnitValue` | Keeps the client's value, computed from a quantity the server replaced |
| `StockAdjustmentTotalsTest.theHeaderTotalsAreSummedFromTheLinesAndNotTakenFromTheBody` | Header says -40 and -15200.00 over lines summing to -15 and -7350.00 |
| `StockAdjustmentTotalsTest.aDocumentWithNoLinesTotalsZeroRatherThanWhateverWasSent` | Keeps a total no line supports |
| `StockAdjustmentTotalsTest.anEditRestatesTheHeaderTotalsFromTheLinesItLeavesBehind` | Keeps the totals of the lines the edit removed |
| `StockAdjustmentApprovalTest.approvingRestatesBothTotalsFromWhatItActuallyPosted` | Posted document keeps the draft's value total |
| `StockAdjustmentApprovalTest.aLineThatMovesNothingContributesNothingToEitherTotal` | Header claims a variance no ledger row backs |

Four more pass either way and are there as ratchets rather than as proof: `aProjectManagerMayNot{Create,Edit,Delete}AMaterial`, `aProjectManagerMayNotMoveAReorderThreshold` and `aProjectManagerMayNotDeleteALocation` fail if a later change widens the writes along with the reads; `aMemberHoldingNeitherRoleIsStillRefusedTheRead` fails if the read is ever widened to plain membership; `aLineWithNoUnitValueKeepsTheValueItWasSentWith` fails if `stampLineValue` is simplified into overwriting unconditionally, which would null the only figure such a line has.

Full `./gradlew build` green on the lab box, test heap 29% of the 2048 MB cap. No schema change, so no changelog. `docs/openapi.json` regenerated after rebasing onto `a899907`; the diff is five description changes and nothing else.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -16033,12 +16033,12 @@
             "type": "integer"
           },
           "totalAdjustmentValue": {
-            "description": "Total monetary value of the adjustment across all line items.",
+            "description": "Total monetary value of the adjustment across all line items. Summed from the line items, so any value sent here is ignored: it is arithmetic over figures the server states and a header that disagreed with the sum of its own lines would be reporting neither. Approving restates it from what was posted.",
             "example": 18500.0,
             "type": "number"
           },
           "totalVarianceQuantity": {
-            "description": "Total variance quantity across all line items.",
+            "description": "Total variance quantity across all line items. Summed from the line items, so any value sent here is ignored: each line's variance is stamped from the balance it was raised against, so a header total computed by the client is a sum of figures the server has since replaced. Approving restates it from what was posted.",
             "example": -120.0,
             "format": "double",
             "type": "number"
@@ -16322,7 +16322,7 @@
             "type": "number"
           },
           "totalAdjustmentValue": {
-            "description": "Total monetary value of this line's adjustment.",
+            "description": "Total monetary value of this line's adjustment. Computed as the stamped adjustment quantity times the unit value, so any value sent here is ignored whenever both of those are known. A line carrying no unit value keeps what was sent, because there is nothing to multiply by; approving such a line writes its value from the running average cost the ledger entry was posted at.",
             "example": -7600.0,
             "type": "number"
           },
@@ -106979,7 +106979,7 @@
       "name": "Materials"
     },
     {
-      "description": "Web-console counterpart of the materials API, restricted to the system-admin role for the current tenant. Covers the same create, browse, search, update, stock lookup and delete operations as the standard materials endpoints, for use from the admin web console rather than the mobile app.",
+      "description": "Web-console counterpart of the materials API, gated by organization role for the current tenant. Covers the same create, browse, search, update, stock lookup and delete operations as the standard materials endpoints, for use from the admin web console rather than the mobile app. Reads, including the stock lookup, are open to the system-admin and project-manager roles, because a project manager raises and approves stock adjustments and those are documents about a material's balance. Changing the catalogue, its thresholds included, stays with system-admin.",
       "name": "Materials (Web)"
     },
     {
@@ -107095,7 +107095,7 @@
       "name": "Storage Locations"
     },
     {
-      "description": "Web-console counterpart of the storage location API, gated by organization role instead of a flat authority. Covers the same create, read and lookup operations as the mobile API plus partial update and delete, which the mobile API does not expose. Every operation requires the system-admin role in the caller's current tenant.",
+      "description": "Web-console counterpart of the storage location API, gated by organization role instead of a flat authority. Covers the same create, read and lookup operations as the mobile API plus partial update and delete, which the mobile API does not expose. Reads are open to the system-admin and project-manager roles in the caller's current tenant, because a project manager raises and approves stock adjustments and cannot pick a location to count without them. Creating, editing and deleting a location stays with system-admin.",
       "name": "Storage Locations (Web)"
     },
     {
--- a/src/main/java/org/tornotron/echno_backend/material/MaterialControllerWeb.java
+++ b/src/main/java/org/tornotron/echno_backend/material/MaterialControllerWeb.java
@@ -31,10 +31,13 @@ import java.util.List;
 @Validated
 @Tag(
         name = "Materials (Web)",
-        description = "Web-console counterpart of the materials API, restricted to the system-admin "
-                + "role for the current tenant. Covers the same create, browse, search, update, stock "
-                + "lookup and delete operations as the standard materials endpoints, for use from the "
-                + "admin web console rather than the mobile app."
+        description = "Web-console counterpart of the materials API, gated by organization role for the "
+                + "current tenant. Covers the same create, browse, search, update, stock lookup and "
+                + "delete operations as the standard materials endpoints, for use from the admin web "
+                + "console rather than the mobile app. Reads, including the stock lookup, are open to "
+                + "the system-admin and project-manager roles, because a project manager raises and "
+                + "approves stock adjustments and those are documents about a material's balance. "
+                + "Changing the catalogue, its thresholds included, stays with system-admin."
 )
 public class MaterialControllerWeb {
 
@@ -70,7 +73,7 @@ public class MaterialControllerWeb {
     }
 
     @GetMapping("/{id}")
-    @PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin')")
+    @PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')")
     @Operation(
             summary = "Get a material by id",
             description = "Returns a single material with its creator and current aggregate stock value."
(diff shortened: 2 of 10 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Merged unreviewed. CodeRabbit answered "Review rate limited" on both the automatic pass and the explicit `@coderabbitai review`, and the policy as of today is not to substitute a review agent for it. Merging on green CI plus the reasoning above.

What I checked myself, in place of the review:

- Both fixes are proved by tests that fail against `development`, listed in the description with what each one does without the fix. The four that pass either way are named as ratchets rather than counted as evidence.
- The authorization change is reads only. I read every `@PreAuthorize` on both controllers before and after: seven GETs on `MaterialControllerWeb` and five on `StorageLocationControllerWeb` moved, and the eight writes across the two are byte-identical. `EndpointAuthorizationTest` still passes with its grandfather set empty, so nothing lost its guard.
- `isMemberOfCurrentTenant()` appears nowhere in this diff, so no read became open to plain membership. `aMemberHoldingNeitherRoleIsStillRefusedTheRead` pins that on both controllers.
- The totals change writes no ledger row and reads no new one. `requireBalanceUnmoved`, `readBalance`, `stampOpeningBalance` and the `postedHere` per-balance-row map are untouched; the eight existing approval tests, including the two balance-moved refusals and the multi-line count sheet, pass unchanged.
- The only new arithmetic on the posting path is `lineValue(movement, unitCost)` after the ledger entry is written. It returns null when either figure is missing, which is the same null `unitCost` the transaction row already tolerates, so a material with no average cost behaves as it did.
- Rounding is to two decimal places, the scale `total_adjustment_value` is declared with, so nothing changes between writing a document and reading it back.
- `docs/openapi.json` was regenerated after rebasing onto `a899907` and the snapshot task passes on the rebased tree. The diff is five description strings. No field was marked read-only, so no request property left the schema and echno-core's write call sites are unaffected.
- Full `./gradlew build` green on the lab box, test heap 29% of the 2048 MB cap, well under the 80% gate.

PR #670, opened 2026-08-31
Say which partial-update fields a null clears, and refuse the rest

abhijithanandan:
Closes the request half of #645 where the question is genuinely hard: the **eight partial-update surfaces, 87 properties**, where a null and an absent field are different requests. The remaining ~1000 request-side properties are specified at the bottom rather than left implied.

## Why this half is a different question

The response side (#661) needed one fact per property: may this be null. This side needs two, and until now the document could carry neither.

They now come from different halves of the document, and neither says anything about the other:

| | says | made accurate by |
|---|---|---|
| `required` | whether the field may be **absent** | #664 |
| the 3.1 type union | whether it may be **null when present** | #661 |

That composition is what makes the request side tractable at all. A partial-update property that is not in `required` and carries `["string", "null"]` means: you may omit this, and if you send it as null the field is cleared. One that is not in `required` and carries `"string"` means: you may omit this, but a null is refused.

## The disagreement, which was the actual work

`TaskUpdateFieldsDto` said at class level that "a field sent as null is cleared". `categoryId` on the same class said it cannot be cleared. `LeaveRequestUpdateFieldsDto.startDate` said "null is rejected". Three statements, one document, no convention.

Reading every branch of all eight switches found **three** behaviours, not two:

| what a null did | keys |
|---|---|
| cleared the field, 200 | 74 |
| refused with a 400 naming the field | 2 |
| **threw `NullPointerException`, which no handler names → 500** | **11** |

The eleven were not a decision anyone had taken. Each reached an unguarded `Enum.valueOf(null)`, `((Number) null).longValue()` or `LocalDate.parse(null)`:

`TaskUpdateFieldsDto.status`, `ProjectUpdateFieldsDto.status`, `projectLatitude`, `projectLongitude`, `IssueUpdateFieldsDto.assignedToId`, `EmployeeUpdateFieldsDto.status`, `managerId`, `UserUpdateFieldsDto.role`, `LeavePolicyUpdateFieldsDto.annualQuota`, `LeaveRequestUpdateFieldsDto.startDate` and `endDate`.

So the honest answer to "what does this class mean by null" was, for seven of the eight, "three different things depending on the key, and one of them is a 500."

## How the eleven were settled, which is not a matter of taste

**A null is refused where the column behind the field is `NOT NULL`, and clears where it is nullable.** The database already enforces that, so the alternative to a refusal was a 500 now or a `DataIntegrityViolation` at flush.

| | column | now |
|---|---|---|
| `Task.status` | `nullable = false` | **400**, "status cannot be cleared" |
| `LeavePolicy.annualQuota` | `nullable = false` | **400** |
| `LeaveRequest.startDate` / `endDate` | `nullable = false` | **400** |
| `Issue.assignedTo` | nullable join column | **clears** — this is what unassigning an issue is |
| `Employee.manager` | nullable join column | **clears** — this is what the top of a reporting line is |
| `Project.status`, `Employee.status`, `User.role` | nullable | **clears** |
| `Project.projectLatitude` / `projectLongitude` | nullable | **clears** |

Two of those were a user-visible gap rather than only a 500: there was no way to unassign an issue or detach an employee from their manager, because the only request that means it answered 500.

## Where the code contradicts itself, said rather than picked

Two keys refuse a null although their column is nullable. Both now say so on their own schema instead of by surprise.

- **`TaskUpdateFieldsDto.categoryId`** is a stated product rule ("a task without a category is a state creation cannot produce"), and the join column is nullable, so it is a rule the schema does not enforce. Kept, and the schema now says which kind of rule it is.
- **`UserUpdateFieldsDto.defaultOrganizationId`** was not a rule at all. It answered `"defaultOrganizationId must be a number"` to a null, which is a type-error message misapplied to a null by a `parseDefaultOrganizationId` that had no null branch. The column is nullable and rows already exist with it null, so it clears now like every other nullable column.

`IssueUpdateFieldsDto.type` and `status` refuse a null on a nullable column too, but deliberately and with their own message, from the enum parser they share with creation. Kept.

## What is published

**80 of the 87 properties marked `@Schema(nullable = true)`; 7 not.** The class-level descriptions no longer assert a blanket rule; each now says that an absent field is untouched, a field the schema declares nullable is cleared by an explicit null, and a field it does not declare nullable refuses one with a 400. The seven say why on themselves.

## Three tests, holding one list

`PartialUpdateSurfaces` gains the table: per surface, the keys that refuse a null. Everything else clears. Written down rather than derived, because the two facts that have to agree live in different places.

1. **`PartialUpdateNullBehaviourTest`** sends `{key: null}` for every declared key of every surface and requires the outcome the table claims. It is driven off the schema class, so a field added to one of these DTOs is covered the day it appears.
2. **`PartialUpdateNullContractTest`** requires the annotation to match the table in both directions: a clearing field must be marked, a refusing one must not.
3. **`OpenApiNullabilityTest`** (existing, from #661) carries the annotation into `docs/openapi.json`.

A key added to a switch without a decision fails all three.

## Proof, against the unfixed tree

Both new tests were run with the seven services and eight DTOs reverted to `development` and the tests and the table kept. **17 failures**, and they are exactly the eleven plus `defaultOrganizationId` plus the 80 undeclared:

```
employee  ["status: threw NullPointerException ("Name is null"), which is neither clearing
            the field nor a 400. An unguarded null reaching a cast or a parser answers 500",
           "managerId: threw NullPointerException ("Cannot invoke
            "java.lang.Number.longValue()" because "value" is null") …"]
project   ["status: …", "projectLongitude: …", "projectLatitude: …"]
task      ["status: threw NullPointerException ("Name is null") …"]
user      ["defaultOrganizationId: threw IllegalArgumentException
            ("defaultOrganizationId must be a number"), which is neither clearing the field
            nor a 400", "role: …"]
leave request ["startDate: threw NullPointerException ("text") …", "endDate: …"]
leave policy  ["annualQuota: threw NullPointerException …"]

PartialUpdateNullContractTest
  ["TaskUpdateFieldsDto.title: clears on null, but is not marked
    @Schema(nullable = true), so the document says it may never be null and a client
    will not send one", … 80 of them]
```

## The document diff, audited entry by entry

**95 changes and nothing else in the file.** No schema added or removed.

- 78 properties gaining `"null"` in a type union or an `anyOf` branch
- 2 more (`assignedToId`, `managerId`) that gained the union *and* a description saying a null unassigns or detaches
- 15 description-only changes: the 8 class descriptions and the 7 refusing fields

## Also fixed while there

An unparseable leave date threw `DateTimeParseException`, which is not an `IllegalArgumentException` and reached no handler either, so a mistyped date answered 500 as surely as a null did. It is a 400 now, pinned by its own test.

## Verified on lab `.116`

- `./gradlew test` **BUILD SUCCESSFUL** in 8m 53s; heap 574 MB live of the 2048 MB cap (28%), 8 Spring contexts cached of a maximum 8
- `./gradlew openApiSnapshot -PupdateOpenApiSnapshot` regenerated on a tree rebased onto `development`, and the diff audited as above

## Not in this change, and what it needs

**The bean-bound request DTOs, roughly a thousand properties.** These are a different and much easier question: they deserialize through Jackson into a typed bean, so absent and null both arrive as a null field and the endpoint cannot tell them apart. There is no "cleared by an explicit null" to describe, and `required` plus the validation constraints #664 published already carry most of what a client needs. The remaining work there is per-field reading — which optional fields the service tolerates as null on the way to the entity — and it is tractable module by module, in the same shape as the response-side pass #661 left open.

**The response-only schemas, 1975 properties**, remain as #661 left them.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -5269,67 +5269,100 @@
         "type": "object"
       },
       "EmployeeUpdateFieldsDto": {
-        "description": "Fields a partial employee update may change. Every field is optional; only the fields present in the request are applied. Keys not listed here are ignored.",
+        "description": "Fields a partial employee update may change. Every field is optional and an absent field is left untouched. A field this schema declares nullable is cleared by sending an explicit null; a field it does not declare nullable refuses a null with a 400 rather than clearing. Keys not listed here are ignored.",
         "properties": {
           "dateOfBirth": {
             "description": "Date of birth.",
             "example": "1998-04-12T00:00:00",
             "format": "date-time",
-            "type": "string"
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "department": {
             "description": "Department the employee belongs to.",
             "example": "Execution",
-            "type": "string"
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "designation": {
             "description": "Job title held.",
             "example": "Site Engineer",
-            "type": "string"
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "emailAddress": {
             "description": "Work email address.",
             "example": "hrishi@echno.xyz",
-            "type": "string"
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "employeeId": {
             "description": "Employee code within the organization.",
             "example": "ECH-0412",
-            "type": "string"
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "employeeName": {
             "description": "Full name of the employee.",
             "example": "Hrishikesh R",
-            "type": "string"
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "joiningDate": {
             "description": "Date the employee joined.",
             "example": "2024-06-01T00:00:00",
             "format": "date-time",
-            "type": "string"
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "managerId": {
-            "description": "Id of the employee's manager. Must belong to the caller's organization, and may not introduce a cycle in the reporting line.",
+            "description": "Id of the employee's manager. Must belong to the caller's organization, and may not introduce a cycle in the reporting line. Send null to detach the employee from their manager.",
             "example": 7,
             "format": "int64",
-            "type": "integer"
+            "type": [
+              "integer",
+              "null"
+            ]
           },
           "phoneNumber": {
             "description": "Contact phone number.",
             "example": "+91 99400 00000",
-            "type": "string"
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "salary": {
             "description": "Monthly salary. Must be sent as a decimal number.",
             "example": 62000.0,
             "format": "double",
-            "type": "number"
+            "type": [
+              "number",
+              "null"
+            ]
           },
           "shiftTimingId": {
             "description": "Id of the shift timing the employee works. Must belong to the employee's organization; null clears the assignment.",
             "example": 3,
             "format": "int64",
-            "type": "integer"
+            "type": [
+              "integer",
+              "null"
+            ]
           },
           "status": {
             "description": "Employment status of the employee.",
@@ -5342,7 +5375,10 @@
               "probation",
               "suspended"
             ],
-            "type": "string"
+            "type": [
+              "string",
+              "null"
+            ]
           }
         },
         "type": "object"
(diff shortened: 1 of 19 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
CodeRabbit reported `Review rate limited` on this PR, so per the standing policy this merges on green CI plus my own reasoning rather than waiting the quota out. **Merged unreviewed**, recorded here.

The reasoning I merged on:

- Every behaviour change is proved by a test run against the unfixed tree first, reporting **17 failures** that name the exact eleven `NullPointerException` sites, the one misapplied `IllegalArgumentException`, and the 80 undeclared properties. Nothing here is a change I could not show was wrong before.
- The eleven were settled by the column behind each field rather than by preference, so the direction of each is derivable from the schema and not from my judgment. The two places where the code contradicts that rule are kept as they are and now say so on their own schema, rather than being quietly aligned.
- Nothing that answered 2xx before answers differently now. Four keys move 500 → 400, seven move 500 → 200, and one moves 400 → 200 on a nullable column. No request that previously succeeded can now fail.
- Checked the read side too: `echno-core` already parses all four newly-clearable values as nullable (`Project.status` and `Employee.status` are `nullableString` with defaults, `Issue.assignedToId` is `optionalNumericId`), so a cleared value cannot break a screen.
- Rebased onto `cd53dfd` after it landed, and the document **regenerated on the rebased tree is byte-identical to the text merge**, so the generated file is not carrying a silent conflict. The diff against `development` is 95 entries, every one of them inside the eight partial-update schemas and none outside; `paths` is unchanged.
- Full suite re-run on lab `.116` after the rebase: **BUILD SUCCESSFUL in 8m 54s**, heap 600 MB live of the 2048 MB cap (29%), 8 contexts cached of a maximum 8. CI `build-and-test` green on the rebased head.

Issue #635, opened 2026-08-31
A movement record's verification stamp is a request parameter, so it can name anyone


abhijithanandan:
Found reviewing PR #634, which closed the same defect on payment vouchers (#631). This is the last instance of the shape in the backend, and unlike #631 it is reachable through the product today.

## What happens

`MovementRecordService.verifyMovement(Long movementId, String verifiedBy)` writes the stamp from its argument:

```java
record.setIsVerified(true);
record.setVerifiedBy(verifiedBy);
record.setVerifiedAt(LocalDateTime.now());
```

`src/main/java/org/tornotron/echno_backend/attendance/service/MovementRecordService.java:136`

The argument comes straight off the wire as `@RequestParam String verifiedBy`, on both `MovementRecordController` and `MovementRecordControllerWeb`. Nothing checks that the value is the caller, that it names a person in the caller's organization, or that it names a person at all: the column is a free-text string.

## Why this one is worse than #631

#631 was containment: no screen wrote the field, so the exposure was the API. Here the client sends it. `echno-core`'s `movement-service.ts` calls `verifyMovement(id, verifiedBy)` with a caller-supplied value, and `use-movement-mutations.ts` takes `{ id, verifiedBy }` from its caller. So the flow is live, and the verifier on an attendance movement is whatever the client typed.

The time is already server-side (`LocalDateTime.now()`), so only the identity half is wrong here. Note it is also a naive `LocalDateTime` rather than an `Instant`, unlike the payment stamp.

## What would fix it

The #598 / #631 answer, if nothing legitimately verifies on someone else's behalf: resolve the verifier from the session, drop the parameter from both controllers, and let `echno-core` stop sending it. Establish that first, the way #631 did, by checking every call site in `echno-core` and `echno-web` for whether any of them passes anything other than the signed-in person.

If something does verify on behalf of someone else, the #589 shape applies instead: keep the parameter and refuse a value that is not the caller unless they hold a role permitting it.

Worth deciding at the same time whether this wants a stored user id rather than a string, so the name can be resolved the way #610 and PR #630 resolve the finance stamps, and whether the naive `LocalDateTime` should become an `Instant`.

## Deploy sequencing

Removing the parameter is **not** safe alone, because the client sends it: a removed `@RequestParam` that is still supplied is ignored, but the client would then silently stop influencing the value, which is the intended outcome and needs the client updated in the same window so nobody is surprised by whose name appears. Refusing a mismatched value would need the client first.

PR #671, opened 2026-08-31
Wait for the lorry before saying the stock arrived

### LINKED ISSUE #660: A site transfer writes both legs at creation, so stock arrives before anybody sees it
Split out of #648. The goods-receipt half of that issue is done (see the PR that closes it): received quantities now post back onto the purchase order, an over-receipt is refused unless the payload acknowledges it, and the order's status follows from the arithmetic. This is the other half, which is larger and changes what the document means to the people using it, so it is specified here rather than built alongside.

## The decision that shapes it

A site transfer writes both inventory legs at creation. Whether that is wrong depends entirely on whether the transfer crosses a site boundary, and the design should follow that split rather than treat all transfers alike.

**Within one project**, where the transfer moves material between storage locations on the same site, the material never leaves that site's custody. There is no window in which it is in transit and nobody is accountable for it. Writing both legs at creation is correct, and such a transfer should be `COMPLETED` the moment it is written rather than sitting in `PENDING` waiting for a confirmation that has nothing to confirm.

**Between projects**, there is a lorry, a road, and a gap of hours or days. Writing the inbound leg at creation asserts that stock arrived at a site where nobody has seen it. That is the claim that makes a receiving-site count unexplainable: the balance says twenty bags, the yard holds none, and the ledger offers a `TRANSFER_IN` whose closing figure nobody can reconcile. This is the case that needs a second step.

## What a between-sites transfer becomes

**Creation writes only the outbound leg.** `TRANSFER_OUT` for the sent quantity at the sending project and location; no `TRANSFER_IN`. Status `PENDING`. The quantity is then in transit, visible as the difference between what a transfer sent and what it has recorded as received.

There is a reporting consequence that must be stated rather than discovered: an organization-wide total of on-hand stock will be short by everything in transit. That is the truth, and the reports that sum stock across projects need a labelled in-transit figure beside the total. The alternative, giving the organization a transit storage location per transfer so the stock is somewhere rather than nowhere, is cleaner for reporting and more machinery; recommend starting without it and adding it if the in-transit figure turns out to need a location to hang on.

**`SiteTransferItem` gains `receivedQuantity`**, nullable until the transfer is received. Today the line has no such field, so a short delivery cannot be recorded even as a note, and ten sent against eight arriving leaves two bags of stock at a site that does not have them.

**A receive action**: `POST /api/v1/site-transfers/{id}/receive` and its `/web` twin, carrying per-line received quantities. It writes the `TRANSFER_IN` legs for what actually arrived, sets `receivedQuantity` per line, and derives the status: every line received in full is `COMPLETED`, some received is `PARTIALLY_TRANSFERRED`, nothing received stays `PENDING`. Unlike the purchase order's derived move, this transition *is* somebody's act, so it takes a session-derived actor (`CurrentEmployeeService.requireCurrentEmployee`) and is filed in the shared status trail as an ordinary `UPDATE` against that person, not as `SYSTEM`.

The existing `PATCH .../status` should stop being able to reach `COMPLETED` once this exists. A status that means stock arrived must not be settable by a payload that moves no stock.

**A short delivery is a shortfall, and it must be closed deliberately.** After a partial receipt the ledger reads honestly: the sending project is down ten against a named transfer, the receiving project is up eight against the same transfer, and two are unaccounted for. That gap should be reported as an open variance on the transfer and closed by a stock adjustment that references it, rather than by the transfer writing a loss movement of its own. A loss written automatically is a stock correction nobody authorised, and #651 exists precisely to put an approver in front of stock corrections.

**Over-receipt on a transfer** follows the purchase order's policy for the same reasons: refused by default, with the refusal naming the line and the figures, and recorded when the payload acknowledges it. Refusing outright would leave arrived stock outside the ledger, which is the failure mode SA-563 showed.

**Cancelling a `PENDING` transfer must reverse the outbound leg.** Without this the two-step document is strictly worse than the one-step one it replaces, because a transfer that is abandoned in transit leaves the sending project permanently short with no way back. This is in scope, not a follow-up.

## The awkward part: what already exists

Every transfer in the database was written with both legs and has no received quantity. The backfill has to reflect what the ledger actually did, not what the new document would have done:

- Set `receivedQuantity = sentQuantity` on every line of every existing transfer, whatever its status. The inbound leg *was* written for the full quantity, so this is the truth about them.
- Do **not** backfill `PENDING` transfers as unreceived. Their stock has already moved. Recording them as unreceived would make the document and the ledger disagree, which is the exact condition this work exists to remove.
- Existing `PENDING` and `PARTIALLY_TRANSFERRED` rows should therefore be moved to `COMPLETED` by the migration, with a note in the status trail saying the status was corrected to match movements that had already been posted. Anything else leaves rows whose status contradicts the ledger.

Check the live counts before writing the migration. At the time of filing, staging held one transfer and one transfer line, so the backfill is cheap to verify there; production tenants will differ.

## Size

Entity change plus migration, three endpoints, DTOs on both sides of the contract, the cancellation reversal, and a backfill whose correctness rests on facts about live data. Larger than the purchase order half, and it changes what a transfer means to a storekeeper, so Anand and Abin should see the shape before it is built.

abhijithanandan:
Closes #660, and with it the site-transfer half of #648. The goods-receipt half landed in #659; this follows its precedent on every question the two share.

## What was wrong

`InventoryEventListener.handleSiteTransferCreated` wrote both legs in one loop: a `TRANSFER_OUT` at the sending project and a `TRANSFER_IN` at the receiving one, before anybody at the receiving site had seen the lorry. `SiteTransferItem` had no received quantity, so a short delivery could not be recorded even as a note. And `PATCH .../status` assigned whatever it was handed and moved no stock, so `COMPLETED` was a label somebody typed.

## The decision the design follows

Whether both legs at creation is wrong depends entirely on whether the transfer crosses a site boundary.

**Within one project it is right.** The material is handed from one store to another on a site whose storekeeper is accountable for it throughout. There is no window in which nobody is. Both legs are written, every line is received in full, and the transfer is `COMPLETED` from the moment it exists rather than sitting in `PENDING` waiting for a confirmation that has nothing to confirm.

**Between two projects it is wrong.** There is a lorry, a road and a gap of hours or days. Creation writes only the outbound leg, the transfer is `PENDING`, and the quantity is in transit until somebody at the far end records what arrived.

## What a between-sites transfer became

- **`SiteTransferItem.receivedQuantity`**, nullable. Null says nobody has confirmed anything about the line yet, which is different from confirming that nothing came.
- **`POST /{id}/receive`** and its `/web` twin, carrying per-line quantities. It posts the inbound legs for what actually arrived and derives the status: every line met is `COMPLETED`, some received is `PARTIALLY_TRANSFERRED`, nothing received stays `PENDING`. Unlike the purchase order's derived move this transition *is* somebody's act, so the actor comes from the session (`CurrentEmployeeService.requireCurrentEmployee`) and it is filed in the shared trail as an ordinary `UPDATE` against that person, not as `SYSTEM`. No actor field on the payload, so this is not a sixth caller-supplied actor after #589, #599, #607, #631 and #635.
- **`POST /{id}/cancel`** with a required reason, reversing the outbound leg back onto the sending balance. In scope, not a follow-up: without it a transfer abandoned in transit leaves the sending project permanently short with no way back, which would make the two-step document strictly worse than the one-step one it replaces. Only reachable from `PENDING`; once something has been received, part of the material is at the far site and its fate is a decision rather than a reversal.
- **`PATCH .../status` now refuses every status** and names the endpoint the caller wanted. The route is kept so an existing client gets an answer rather than a 404.
- **`GET /web/{id}/status-history`**, the same shape the project and purchase order trails use.

## The shortfall policy, and why it does not mirror the over-receipt one

Over-receipt follows #659 exactly: refused by default with the line and the figures named, recorded when the payload sets `allowOverReceipt`.

A shortfall is accepted with no acknowledgement, and that is the same rule applied to the other direction rather than an inconsistency. The over-receipt guard exists because the document would otherwise assert stock the transfer never sent. Eight arriving against ten sent asserts nothing false: the ledger holds an honest outbound leg for ten and an honest inbound leg for eight, and the two missing bags are the difference. Refusing the short receipt would force the storekeeper either to claim ten came or to leave the eight that did outside the ledger, which is the SA-563 failure the over-receipt escape hatch exists to prevent.

The gap is left as an **open variance**, reported per line as `inTransitQuantity`, and closed by a stock adjustment naming the transfer. The transfer writes no loss movement of its own: a loss written automatically is a stock correction nobody authorised, and #651 exists to put an approver in front of those.

## Reporting consequence, stated rather than discovered

An organization-wide total of on-hand stock is now short by everything in transit. That is the truth about material on a lorry. Each line reports its own in-transit figure so a report summing stock across projects can show a labelled in-transit total beside the on-hand one. The alternative, a transit storage location per transfer, is cleaner for reporting and more machinery; not built, per the issue's recommendation.

## Ledger safety

No new movement shapes. A cancellation's reversal is a `TRANSFER_IN` against the sending balance, which is where the stock really is arriving back, with the remark saying why.

One correctness fix the split forced: **the arrival is valued at the unit cost the outbound leg carried**, read back off the ledger, not at the sending balance's average cost as it stands at receive time. By the time a lorry is unloaded the sender may hold none of the material, and `getAverageCost` would return zero and silently destroy the stock value that left. Falls back to the current average only when no outbound movement is found.

The lines are taken with `SELECT ... FOR UPDATE` ordered by id, so two people confirming the same lorry queue rather than each judging themselves against the figure that stood before the other. Receive and cancel run under the house `TransactionRetryTemplate`, as create does; the inbound movements publish after commit, so an attempt that rolls back posts nothing.

## Live data, checked read-only before the migration was written

Staging holds **one transfer and one line**: `TRF-2026-000001`, `COMPLETED`, project 2 to project 6 (so cross-project), 18 of TNT Steel, with both legs posted (`TRANSFER_OUT` -18 opening 50 closing 32, `TRANSFER_IN` +18 opening 0 closing 18). No `PENDING` or `PARTIALLY_TRANSFERRED` rows, so 086-04 touches nothing there but is needed for other tenants. `site_transfer.status` is `VARCHAR` with no check constraint, so `CANCELLED` needs no schema change. `status_transition` holds 28 `PROJECT` and 1 `PURCHASE_ORDER` row and no `SITE_TRANSFER` rows, so 086-03's precondition passes. **No row resists the new model**: the one that exists is exactly what the backfill describes.

## The backfill makes the documents agree with the ledger, not the reverse

Every existing transfer wrote both legs, so its stock reached the receiving site whatever its status said. Rewriting the movements to match a document that has since changed shape would break the property `SA-563` depended on, which is that every balance is explainable from its movements.

- `086-01` adds the nullable column.
- `086-02` sets `received_quantity = sent_quantity` on every existing line, scoped to nulls so a re-run cannot overwrite a real receipt.
- `086-03` seeds one `BASELINE` per transfer, **before** the correction, so a reader sees the status the transfer was observed to hold and then the entry that moved it.
- `086-04` moves `PENDING` and `PARTIALLY_TRANSFERRED` to `COMPLETED` and writes a `SYSTEM` entry saying the status was corrected to match movements already posted. `SYSTEM` and not `BASELINE`, because this is a real change the migration made rather than an observation of a status already held, and nobody decided it.

The one window the changeset names in its own comment: a rolling deploy that leaves an older instance serving could write an old-style both-legs `PENDING` transfer after the backfill. Staging and the compose environments replace the single instance rather than rolling.

## Tests

46 new. Every one was checked by mutation: each mutant below was applied to `development` plus this change, the suite run, and the failures recorded. All 17 were caught.

| Mutation | Caught by |
|---|---|
| over-receipt refusal removed | `refusesAnOverReceiptThatWasNotAcknowledged`, `countsWhatHasAlreadyArrivedWhenJudgingAnOverReceipt` |
| inbound leg written at creation again | `aTransferBetweenTwoProjectsPostsOnlyTheOutboundLegAtCreation` + 3 |
| creation never completes a within-project transfer | `create_betweenTwoStoresOnOneProject_isCompletedAndReceivedInFull` |
| receivable guard removed | the three refusal tests |
| arrival priced off the sending balance now | `aReceiptValuesTheArrivalAtWhatLeftTheSendingSiteNotAtWhatIsLeftThere` + 4 |
| status patch sets and saves again | `updateStatus_refusesToClaimAStatusThatSaysStockMoved` |
| cancellation publishes no reversal | `cancellingAPendingTransferReversesTheOutboundLeg` |
| received quantity never written to the line | 10 tests |
| the move filed as SYSTEM with no actor | `filesTheMoveAgainstThePersonWhoConfirmedTheDelivery` |
| lines read without a write lock | `takesTheLinesUnderAWriteLock` + 3 |
| shortfall raises stock for the full sent quantity | `acceptsAShortDeliveryWithoutAnAcknowledgementAndLeavesTheGapOpen` + 1 |
| unmatched receipt line skipped instead of refused | `refusesAReceiptNamingALineThatIsNotOnThisTransfer` |
| actor taken leniently instead of fail-closed | `refusesACallerWithNoEmployeeRecordBeforeAnyStockMoves` + 4 |
| every transfer treated as crossing a boundary | `crossesProjectBoundaryIsTrueOnlyWhenTheTwoProjectsDiffer` + 3 |
| cancel guard removed | `refusesToCancelATransferThatHasAlreadyHadSomethingReceived` + 1 |
| cancellation records nothing on the trail | `cancellingRecordsWhyOnTheStatusTrail` |
| in-transit quantity always zero | `aLineNobodyHasConfirmedReportsTheWholeSentQuantityAsInTransit` + 1 |

`SiteTransferBackfillIT` runs the changeset's own `<sql>`, read out of the XML rather than copied, against a real CockroachDB over rows shaped like the live one. Its five mutants were checked the same way. One of them found a gap worth naming: making the changeset write its trail entries for a narrower set of rows than its `UPDATE` moves passed at first, because the transition assertion only covered a `PENDING` transfer. A `PARTIALLY_TRANSFERRED` row would then have changed status with nothing in its history saying it ever did, which is the silent change the trail exists to prevent. The test now asserts both rows carry an entry naming their own prior status.

## Contract

Additive apart from one break. New: `receivedQuantity` and `inTransitQuantity` (both read-only) on `SiteTransferItemDto`, `CANCELLED` on `SiteTransferStatus`, `SiteTransferReceiptDto`, `SiteTransferReceiptLineDto`, `SiteTransferCancellationDto`, and three endpoints on each controller. The break is `PATCH .../status`, which now always returns 400. Filed for the web app as tornotron/echno-web#387, the sibling of #380.

`docs/openapi.json` regenerated with `-PupdateOpenApiSnapshot`.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -15637,6 +15637,22 @@
         },
         "type": "object"
       },
+      "SiteTransferCancellationDto": {
+        "description": "Why a transfer that never arrived is being abandoned. Cancelling returns the stock to the sending project and location it was drawn from, which is a real movement on the ledger, so the reason is required and is kept on the transfer's status trail beside the movement it caused.",
+        "properties": {
+          "reason": {
+            "description": "Why the transfer is being cancelled.",
+            "example": "Lorry turned back at the gate, materials never left the yard",
+            "maxLength": 500,
+            "minLength": 1,
+            "type": "string"
+          }
+        },
+        "required": [
+          "reason"
+        ],
+        "type": "object"
+      },
       "SiteTransferCreationDto": {
         "description": "Payload to create a site transfer moving materials from a sending project to a receiving project. The transfer number is allocated by the server and returned on the created transfer; it is not part of this payload. The two sides must differ: a transfer naming the same project and the same storage location on both sides moves nothing and is refused, while two storage locations inside one project is a real move and is accepted.",
         "properties": {
@@ -15685,7 +15701,7 @@
             "type": "integer"
           },
           "status": {
-            "description": "State the transfer starts in. Optional, and PENDING is the only value accepted, so leaving it out is the same as sending PENDING. PARTIALLY_TRANSFERRED and COMPLETED say the receiving site has taken delivery, which is recorded through PATCH /{id}/status; that endpoint wants a different authority from create, and a transfer issued as COMPLETED would stand as received with nobody having confirmed receipt.",
+            "description": "State the transfer starts in. Optional, and PENDING is the only value accepted, so leaving it out is the same as sending PENDING. A transfer between two projects is issued PENDING with only its outbound leg posted, and the quantity is in transit until somebody at the receiving site records what arrived with POST /{id}/receive. A transfer between two storage locations on one project is written COMPLETED by the server, because the material never leaves that site's custody and both legs are posted at once; that is the server's own finding and cannot be asked for here. PARTIALLY_TRANSFERRED and COMPLETED say the receiving site has taken delivery and CANCELLED says the stock came back, and each of those now follows from a movement rather than from a payload.",
             "enum": [
               "PENDING"
             ],
@@ -15773,13 +15789,14 @@
             "type": "string"
           },
           "status": {
-            "description": "Current status of the transfer.",
+            "description": "Current status of the transfer. PENDING means it has left the sending site and nothing has been recorded as arriving, so its quantity is in transit; PARTIALLY_TRANSFERRED and COMPLETED follow from what the receiving site confirmed; CANCELLED means it was abandoned in transit and its stock returned to the sender.",
             "enum": [
               "PENDING",
               "PARTIALLY_TRANSFERRED",
-              "COMPLETED"
+              "COMPLETED",
+              "CANCELLED"
             ],
-            "example": "IN_TRANSIT",
+            "example": "PENDING",
             "type": "string"
           },
           "transferNumber": {
@@ -15799,6 +15816,13 @@
             "format": "int64",
             "type": "integer"
           },
+          "inTransitQuantity": {
+            "description": "How much of this line is neither at the sending site nor recorded as having reached the receiving one: sent minus received, or the whole sent quantity while nothing has been received. On a transfer still in transit this is stock on a lorry. On one that has been received it is an open variance, closed by a stock adjustment naming this transfer rather than by the transfer writing a loss of its own. Zero once everything sent has arrived.",
+            "example": 0,
+            "format": "int32",
+            "readOnly": true,
+            "type": "integer"
+          },
           "materialId": {
             "description": "Id of the material being transferred.",
             "example": 21,
@@ -15809,6 +15833,16 @@
             "description": "Name of the material being transferred.",
             "example": "TMT Bar 12mm",
             "type": "string"
+          },
+          "receivedQuantity": {
+            "description": "Quantity recorded as having arrived at the receiving site, in the material's unit. Null while the transfer is in transit and nobody has confirmed anything about this line yet, which is not the same as saying nothing arrived. A transfer between two stores on one project is received in full at creation, since the material never leaves that site's custody.",
+            "example": 18,
+            "format": "int32",
+            "readOnly": true,
+            "type": [
+              "integer",
+              "null"
+            ]
           },
           "remarks": {
             "description": "Optional remarks for this line item.",
@@ -15826,6 +15860,64 @@
         "required": [
           "materialId",
           "sentQuantity"
+        ],
+        "type": "object"
+      },
+      "SiteTransferReceiptDto": {
+        "description": "What arrived at the receiving site against a transfer that is in transit. Only the quantities are in this payload: who confirmed the delivery is taken from the session, not from the request, so a receipt is the caller's own statement and cannot be filed under a colleague's name.",
+        "properties": {
+          "allowOverReceipt": {
+            "description": "Set when the delivery really did carry more than the transfer sent. Without it an over-receipt is refused and the refusal names the line and the figures, which is enough to recognise a typed digit. With it the excess is recorded and the stock is posted, because material standing in the yard is there whether the system likes it or not and a receipt that cannot be filed leaves it outside the ledger. A short delivery needs no such flag: recording less than was sent puts nothing false into the ledger and leaves an open variance for a stock adjustment to close.",
+            "example": false,
+            "type": "boolean"
+          },
+          "items": {
+            "description": "One entry per transfer line being confirmed. Lines left out of the payload are untouched, so a delivery that answers only part of a transfer names only the lines it answers.",
+            "items": {
+              "$ref": "#/components/schemas/SiteTransferReceiptLineDto"
+            },
+            "minItems": 1,
+            "type": "array"
+          },
+          "receivedOn": {
+            "description": "Date and time the delivery was taken. Optional; the moment the receipt is filed is used when it is left out. It dates the inbound movements this receipt writes.",
+            "example": "2026-01-17T14:05:00",
+            "format": "date-time",
+            "type": "string"
+          },
+          "remarks": {
+            "description": "Optional note about the delivery, carried onto the inbound movements and the transfer's status trail.",
+            "example": "Two bags split in transit, driver informed",
+            "maxLength": 500,
+            "minLength": 0,
+            "type": "string"
+          }
+        },
+        "required": [
+          "items"
+        ],
+        "type": "object"
+      },
+      "SiteTransferReceiptLineDto": {
+        "description": "How much of one transfer line arrived. The line is named by its own id rather than by its material, because a transfer may carry the same material on more than one line and a receipt has to say which of them it answers.",
+        "properties": {
+          "itemId": {
+            "description": "Id of the site transfer line this quantity is against.",
+            "example": 84,
+            "format": "int64",
+            "type": "integer"
+          },
+          "receivedQuantity": {
+            "description": "Quantity that arrived, in the material's unit. Zero is a legitimate answer and means this line's material did not arrive on this delivery; the quantity stays in transit and the transfer is not moved on by it.",
+            "example": 18,
+            "format": "int32",
+            "minimum": 0,
+            "type": "integer"
+          }
+        },
+        "required": [
+          "itemId",
+          "receivedQuantity"
         ],
         "type": "object"
       },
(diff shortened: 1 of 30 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
CodeRabbit is rate limited and has been through two pushes, so this merges on green CI plus my own reading rather than on a bot review. **Recording it as merged-unreviewed.**

What I checked by hand, since nothing else did:

- **The refactor of `handleSiteTransferCreated` is behaviour-preserving for the case it still covers.** The outbound and inbound bodies were lifted into `writeTransferOut` and `writeTransferIn` verbatim, including the two details that are easy to lose: the outbound leg reads the average cost *before* `updateCurrentStock` runs, so it prices the stock that is leaving and not what remains behind it; and it passes `null` rather than the cost into `updateCurrentStock`, which is what the original did. A within-project transfer produces exactly the two movements it produced before.
- **No new movement shapes.** The only new writes are `site_transfer_item.received_quantity`, `site_transfer.status`, `status_transition`, and inbound legs that were previously written at creation instead. A cancellation's reversal is a `TRANSFER_IN` against the sending balance, which is where the stock is genuinely arriving back.
- **Refusals happen before anything moves.** `requireReceivable` and the over-receipt check both throw before `saveAll` and before any event is published, and the events are `AFTER_COMMIT`, so a rolled-back attempt posts nothing. `raisesNoStock`-style coverage for that is `refusesACallerWithNoEmployeeRecordBeforeAnyStockMoves` and the three refusal tests.
- **Tenant scope on the one new read.** `findUnitCostsForReference` filters on `organization.id`, taken off the transfer rather than the thread-local, so it cannot drift with the ambient context inside an `AFTER_COMMIT` listener. The new `lockBySiteTransferIdAndOrganizationId` is scoped the same way as the purchase order's equivalent from #659.
- **Lazy proxies.** Everything the listener and the guards touch on a lazy `@ManyToOne` is `getId()`, which resolves off the proxy without a load.
- **Authorities.** Receive and cancel reuse `site-transfer:update` / `site-transfer:admin` and the `system-admin` org role, so no realm change is needed and `EndpointAuthorizationTest` covers the guard's presence. Two web-slice tests pin the exact role.
- **The migration was executed, not just read.** `SiteTransferBackfillIT` runs the changeset's own `<sql>` against a real CockroachDB and was mutation-checked five ways; one of those found a genuine gap, now closed (see the PR body).

Full suite on `.116` after the last push: **1734 tests, 0 failures**, test heap 588 MB of 2048 (29%), 8 contexts cached of 8.

PR #674, opened 2026-09-01
Total the stock value across the organization, not across a page

### LINKED ISSUE #673: There is no organization-wide total of stock value, so the console cannot show one honestly
Found while fixing `echno-web#389`, and filed rather than worked around because the number cannot be produced in a browser.

## What the console wants

The materials dashboard has a "Total Stock Value" tile: one money figure, rendered in lakhs and crores, described as "current inventory value". It is currently `materials.reduce((s, m) => s + (m.stockValue ?? 0), 0)` over the array from `GET /materials/web`.

That endpoint is `UnpagedResultCap.respond(...)`, so it serves at most `MAX_ROWS` = 500 rows. Past 500 materials the tile shows the value of whichever 500 rows the query returned first and presents it as the inventory. It is a money figure, so it will be quoted.

## Why it cannot be fixed on the client

A count can be. `GET /materials/web/all` is a `Page`, so `totalElements` is the true catalogue size, and `echno-core#86` now reads it. A **sum** cannot: totalling 743 rows means holding 743 rows, and the point of the cap is that the client does not.

So `echno-web#389` gates the tile instead. Below the cap the browser sum is the whole catalogue and is shown; at or above it the tile shows no figure and says how many of how many materials it holds. Honest, and empty exactly when a customer's catalogue gets large enough to matter.

## What would fix it

A materials summary read under `/materials/web`, tenant-scoped like the rest, answering the aggregates the dashboard strip is built from:

- `totalStockValue` — `SUM(cs.stockValue)` across the organization. `CurrentStockRepository` already has `sumStockValueByMaterial` and `sumStockValueByMaterialAndProject`; this is the same aggregate with the material predicate dropped.
- `materialCount` — the catalogue size. Available today from `/all`'s `totalElements`, but it belongs on the same payload once one exists, and it saves the extra `pageSize=1` request the console makes for it now.
- `distinctUnits` — `COUNT(DISTINCT m.unit)`. Third tile in the strip ("Unique Units"), counted over the same capped array and under-reporting for the same reason.

A `projectId` scope, matching `/materials/web/low-stock`, would also let the per-project material views show their own totals rather than organization-wide ones.

One query per aggregate, computed in the database. Nothing here is derivable from a page of rows, which is the whole reason to add it.

## Not urgent

Staging holds 15 materials. Every instance is well under the cap today, so no figure is wrong right now. The consequence of not doing it is that the tile goes blank at 501 materials rather than showing a wrong number, which is the correct failure but still a missing feature.

abhijithanandan:
Closes #673.

`GET /api/v1/materials/web/summary` returns the four figures a materials dashboard strip is built from, totalled in the database:

```json
{ "projectId": null, "materialCount": 743, "distinctUnits": 9,
  "totalStockValue": 50148300.00, "unvaluedHoldingCount": 0 }
```

`?projectId=` narrows every figure to one project. Read is `system-admin` or `project-manager`, matching the listing it feeds (widened in #669). Writes are untouched.

## The shape, and why one endpoint

The strip is three tiles over the same scope, and the console asks for them together. A `Page` cannot carry them: `totalElements` gives the count and there is nowhere on the envelope for a sum. Putting the value on `MaterialDto` would put a catalogue-wide figure on every row of a page of the catalogue. So one read, four figures, one scope for all four.

Two scopes, matching the first two the low-stock read offers, because a project's material view showing an organization-wide total would be the same class of misleading figure this replaces. They count different things and the difference is the point: at organization scope `materialCount` is the catalogue, whether or not a material has ever been stocked; at project scope it is what the project carries a balance row for, which is the same set the value is summed over. Counting the catalogue at project scope would report every material against a site that has never held one.

A project that is not this tenant's is a 404, not a summary of zeroes, for the reason `LowStockService` gives: zeroes are an answer, and they must not be the answer to a question asked of a project that is not there.

## Stock with no cost behind it

`updateCurrentStock` adds value only when the movement carries a unit cost:

```java
if (quantityChanged > 0 && unitCost != null) {
    stock.setStockValue(stock.getStockValue().add(incomingValue));
}
```

So a receipt posted with no unit cost adds quantity at no value, and the balance row holds stock worth zero. `stock_value` is `NOT NULL`, so nothing is null by the time a sum sees it: the zero was written by the posting path, not chosen here.

**The sum counts those rows at the zero they hold, and reports how many it counted.** Excluding them would understate silently, which is the failure this issue exists to fix. Refusing the whole total over one of them would throw away a figure that is right about everything else. `unvaluedHoldingCount` is what lets the console say the total understates and by how many holdings, which is the same judgement the tile already makes about the 500-row cap.

A row at zero quantity and zero value is not counted: that is a location that has run out, which is a complete and correct answer, and every seeded balance row starts in exactly that state.

## Tenant scope

Both aggregates carry an explicit `organization_id` predicate **as well as** running under the Hibernate `orgFilter`. That is not the convention on the surrounding per-material reads and is deliberate: the filter is off for a bypassed request, and on a sum that is invisible. A total spanning every tenant is still a number, and nobody reading it can tell.

The filter is proved to reach the aggregate rather than assumed. `valueTotal_runsUnderTheTenantFilter` and `catalogueCount_runsUnderTheTenantFilter` enable `orgFilter` for one organization and ask for another's total: the two scopings intersect to nothing, so the zero they assert is only reachable if the filter clause was appended to the aggregate SQL. Had Hibernate applied filters to entity queries but not to these, the query's own predicate would have handed back the figure. Both tests fail when the `@Filter` is taken off the entity.

## Tests

Twelve new, plus two existing files updated for the new endpoint. Each was run with the code it pins broken, and each failed:

| Test | Broken by | Then |
|---|---|---|
| `valueTotal_excludesAnotherOrganizationsStock` | value total loses its organization predicate | fails |
| `valueTotal_runsUnderTheTenantFilter` | `@Filter` off `CurrentStock` | fails |
| `catalogueCount_runsUnderTheTenantFilter` | `@Filter` off `Material` | fails |
| `catalogueCounts_excludeAnotherOrganization` | unit count loses its `DISTINCT` | fails |
| `emptyHolding_isNotReportedAsUnpriced` | unpriced count stops requiring a quantity on hand | fails |
| `unpricedStock_isCountedAtZeroAndReported` | unpriced count looks for a negative value instead of no value | fails |
| `projectScope_coversWhatTheProjectCarries` | project figures lose their project predicate | fails |
| `projectScope_excludesAnotherOrganization` | project figures lose their organization predicate | fails |
| `emptyOrganization_totalsToZero` | the empty total is no longer coalesced | fails |
| `unknownProject_isRefused` | the project existence check is removed | fails |
| `projectScope_countsWhatTheProjectCarries` | project scope answers with the catalogue counts | fails |
| `EndpointAuthorizationTest`, `MaterialControllerWebAuthzTest` | `@PreAuthorize` off `/summary` | fails |

`MaterialStockSummaryIT` runs on the plain `@DataJpaTest` configuration the other material repository tests use, so it shares their Spring context rather than adding one.

## Not done

No index on `current_stock(organization_id)`. The organization-scope sum reads essentially the whole tenant's balance rows whether or not one exists, and on the per-client production model that is the table. The project-scope variant is bounded by one project. Worth revisiting with a measurement rather than ahead of one.

`docs/openapi.json` was regenerated with `-PupdateOpenApiSnapshot` on the rebased tree and came back byte-identical to the committed file: 151 insertions, no deletions, the new path and schema only.

## Summary by CodeRabbit

* **New Features**
  * Added a materials stock summary endpoint for dashboard metrics.
  * View total stock value, material count, distinct units of measure, and unvalued holdings.
  * Optionally filter summaries by project.
  * Summaries are scoped to the current organization, with validation for inaccessible projects.

* **Tests**
  * Added coverage for organization and project summaries, authorization, tenant isolation, empty data, and stock valuation scenarios.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -10448,6 +10448,44 @@
         },
         "type": "object"
       },
+      "MaterialStockSummaryDto": {
+        "description": "The figures a materials dashboard strip is built from, totalled in the database over the whole scope rather than over the rows a client happens to hold. Every figure is for the same scope: the organization, or one project when projectId is given.",
+        "properties": {
+          "distinctUnits": {
+            "description": "How many distinct units of measure those materials are held in.",
+            "example": 9,
+            "format": "int64",
+            "type": "integer"
+          },
+          "materialCount": {
+            "description": "How many materials the figures cover. At organization scope this is the catalogue size. At project scope it is how many materials the project carries a balance row for, which is the same set the value is summed over.",
+            "example": 743,
+            "format": "int64",
+            "type": "integer"
+          },
+          "projectId": {
+            "description": "The project these figures were totalled within, null when the whole organization was totalled.",
+            "example": 5,
+            "format": "int64",
+            "type": [
+              "integer",
+              "null"
+            ]
+          },
+          "totalStockValue": {
+            "description": "The value of the stock on hand, summed over every balance row in scope at its running weighted-average cost. Zero when nothing is held.",
+            "example": 5.01483E7,
+            "type": "number"
+          },
+          "unvaluedHoldingCount": {
+            "description": "How many balance rows hold a quantity the total could not price, because the receipts behind them carried no unit cost and so added quantity at no value. Those rows are in the total at the zero they actually hold, so a non-zero count here means the total is an understatement and says by how many holdings. Zero means every holding in scope is priced and the total is complete.",
+            "example": 0,
+            "format": "int64",
+            "type": "integer"
+          }
+        },
+        "type": "object"
+      },
       "MaterialUpdateDto": {
         "description": "Payload to partially update a material. Fields left null are left unchanged.",
         "properties": {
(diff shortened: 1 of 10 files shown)


abhijithanandan:
@coderabbitai review

PR #678, opened 2026-09-06
Say which vendor response fields are null, and which never are

abhijithanandan:
Advances #645. Not a fix for the mechanism, which #661 already built: this is the first module through the pass that issue was left open for.

## What I found before building anything

The brief I picked this up from described the document as declaring nothing nullable across 3069 properties, with `@Schema(nullable = true)` silently discarded by swagger-core's `Schema31Mixin`, and asked for a customizer, a mixin override or a pin to 3.0. All of that is already done and merged, so it is worth saying plainly rather than rebuilding it:

- `NullableSchemaCustomizer` (#661, `8a3164c`) already converts the flag into the 3.1 form: a type union for a typed property, an `anyOf` with `{"type": "null"}` for a bare `$ref`. The `OpenApiCustomizer` was the right choice over a `ModelConverter` and I would not change it.
- `docs/openapi.json` on `development` already carries **139** properties admitting null (136 type unions, 3 `anyOf`), not 0.
- The `@Size` overwriting `@NotBlank`'s `minLength: 1` finding was a separate ticket, #657, fixed in #664 by `ValidationConstraintsModelConverter`.
- The eight partial-update surfaces were done in #670.

So the mechanism works and is held by `OpenApiNullabilityTest`. What #645 stayed open for is the per-field reading work, and that is what this is.

## The gap this closes

`OpenApiNullabilityTest` checks one direction: an annotated field says so in the document. It cannot say anything about a field nobody marked, and across the other **1987** response-only properties nothing does. A reader cannot tell a property nobody has looked at from one that is genuinely never null, so the uniform silence carries no information either way.

This makes the silence mean something on one module. Every property of the six vendor response schemas is classified, from the Liquibase column, the mapper, or the query behind it.

| schema | nullable | of | notable |
|---|---|---|---|
| `VendorDto` | 8 | 19 | `paymentTerms` is the inverse side of a `@OneToOne` the service populates only when terms were supplied, and `deletePaymentTerms` removes again |
| `VendorContactDto` | 1 | 6 | `alternatePhone` |
| `VendorBankAccountDto` | 3 | 7 | `ifscCode` / `swift`, since an account is domestic or international and rarely carries both |
| `VendorPaymentTermsDto` | 2 | 4 | `creditLimit` and `creditDays`, where a client writing `?? 0` turns "no limit agreed" into a limit of zero |
| `VendorTaxIdentifierDto` | 0 | 3 | every column `NOT NULL` |
| `VendorSummaryDto` | 0 | 9 | every money total summed under `COALESCE(..., 0)`, so none can come back null the way a bare `SUM` would |

**One thing worth carrying to the next module: the entity's own `@Column(nullable = ...)` is not the authority.** `ddl-auto` is `validate`, and Hibernate's schema validation does not compare nullability, so an entity can disagree with its table and nothing in the build will say so. The Liquibase changelog is what I read. For vendor the two happen to agree, but that was checked rather than assumed.

## What holds it

`ReviewedResponseSchemas` records both halves per schema, and `ResponseNullabilityRatchetTest` requires the table, the annotation on the field and the union in `docs/openapi.json` to agree. Writing the **non-null** half down explicitly, rather than leaving it as the complement of the marked one, is what makes it a ratchet: a property added to a reviewed schema fails the build until someone classifies it, instead of drifting into "non-null" by being unmarked.

Property names in the table are the document's, not the field's. Jackson strips the `is` prefix from a primitive boolean, so `VendorContactDto.isPrimary` publishes as `primary`; the test resolves this through Jackson's own introspection, which is what the document is generated through.

### Failing-test table

Same tests, three trees. Run on `lab-node-116-f`.

| test | unfixed tree | one name removed from the table | this PR |
|---|---|---|---|
| `theAnnotationAgreesWithTheTable` | FAIL, all 14 listed as nullable but unannotated | FAIL, `VendorDto.notes` marked but listed non-null | PASS |
| `theDocumentAgreesWithTheTable` | FAIL, all 14, eg `paymentTerms` still a bare `$ref` | FAIL, `VendorDto.notes` admits null but listed non-null | PASS |
| `everyPublishedPropertyIsClassified` | PASS, property names did not change | FAIL, `VendorDto: [notes] are published but not classified` | PASS |
| `theTableIsNotVacuous` | PASS | PASS | PASS |

The middle column is the drift case demonstrated rather than asserted: it is the code path a newly added property takes.

`OpenApiNullabilityTest`, `PartialUpdateNullContractTest` and `PartialUpdateNullBehaviourTest` stay green, and the full `./gradlew test` suite passes (8m 46s, heap 610 MB live of the 2048 MB cap, 30%).

## Document diff

Audited semantically, not by line count: **14 properties changed, 0 added, 0 removed, `paths` byte-identical.**

```
VendorBankAccountDto: accountHolderName, ifscCode, swift
VendorContactDto:     alternatePhone
VendorDto:            city, country, notes, paymentTerms, pinCode, state, vendorAddress, website
VendorPaymentTermsDto: creditDays, creditLimit
```

## Scoped out, still open on #645

- The request-side properties that are not on a partial-update surface. Lower value: absent and null are the same thing to a bean-bound DTO, and `required` plus the #664 constraints already carry it.
- The response schemas of every other module, about 1930 properties. Per-field reading work with no shortcut, worth taking a module at a time so each pass arrives as a diff someone can review. Vendor went first because its five tables are the whole of it and its mappers copy by name, so every answer is checkable.
- Adding descriptions to the non-null vendor properties. That is the documentation gap, not this one, and doing it here would have buried 14 real changes in 48.

Closes nothing; #645 stays open for the modules above.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -19272,7 +19272,11 @@
       "VendorBankAccountDto": {
         "properties": {
           "accountHolderName": {
-            "type": "string"
+            "description": "Name the account is held in, where it differs from the vendor's own name. Null where it was not recorded.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "accountNumber": {
             "type": "string"
@@ -19288,10 +19292,18 @@
             "type": "integer"
           },
           "ifscCode": {
-            "type": "string"
+            "description": "IFSC code, which only an Indian domestic account has. Null on a foreign account, which carries a SWIFT code instead.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "swift": {
-            "type": "string"
+            "description": "SWIFT code, which only an account reachable internationally has. Null on a domestic account.",
+            "type": [
+              "string",
+              "null"
+            ]
           }
         },
         "type": "object"
@@ -19328,7 +19340,11 @@
       "VendorContactDto": {
         "properties": {
           "alternatePhone": {
-            "type": "string"
+            "description": "Second number for this contact. Null where only one was given.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "contactPerson": {
             "type": "string"
@@ -19435,7 +19451,11 @@
             "type": "array"
           },
           "city": {
-            "type": "string"
+            "description": "City. Optional on creation, so null where none was given.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "contacts": {
             "items": {
@@ -19444,7 +19464,11 @@
             "type": "array"
           },
           "country": {
-            "type": "string"
+            "description": "Country. Optional on creation, so null where none was given.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "goodsReceivedNotes": {
             "items": {
@@ -19457,7 +19481,11 @@
             "type": "integer"
           },
           "notes": {
-            "type": "string"
+            "description": "Free-text notes about the vendor. Null where none were written.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "payables": {
             "items": {
@@ -19466,10 +19494,22 @@
             "type": "array"
           },
           "paymentTerms": {
-            "$ref": "#/components/schemas/VendorPaymentTermsDto"
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/VendorPaymentTermsDto"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "description": "The vendor's single payment-terms record. Null until terms are set, and null again once they are deleted."
           },
           "pinCode": {
-            "type": "string"
+            "description": "Postal code. Optional on creation, so null where none was given.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "purchaseOrders": {
             "items": {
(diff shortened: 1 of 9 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Review status, recorded here because the register cannot take it right now.

CodeRabbit returned **Review rate limited** on the explicit `@coderabbitai review` request, so per the standing policy this merges on green CI plus my own reasoning and goes on the unreviewed register. ClickUp `86d45heqm` is itself rate limited (retry-after is about 23 hours), so the entry is here until it can be copied across:

> **echno-backend #678**, "Say which vendor response fields are null, and which never are". Merged without a CodeRabbit review, rate limited. Green CI, plus the full `./gradlew test` suite run locally on `lab-node-116-f`. Risk is low and bounded: the change is 14 `@Schema(nullable = true)` annotations, three new test methods and a test-side table. No production behaviour changes, no endpoint, mapper, query or entity is touched, and the document diff is 14 properties with none added or removed and `paths` byte-identical.

The one judgement worth a second pair of eyes is not in the code, it is in the table: whether each of the 14 is genuinely nullable and each of the 34 genuinely is not. Every answer is traceable to a named column in `db/changelog/v4.0/baseline-001-level0-tables.xml` or `baseline-002-level1-tables.xml`, or to the `COALESCE` in `VendorSummaryService`, and `ReviewedResponseSchemas` cites which for each schema, so it can be checked without rerunning anything.

PR #679, opened 2026-09-06
Let a project manager read the ledger behind what they approved

### LINKED ISSUE #672: A project manager can approve a stock adjustment but cannot read the ledger rows it wrote
Adjacent gap left out of #666 deliberately, recorded so it is not lost.

#666 widened the reads a project manager needs to *raise* a stock adjustment: the material catalogue, the storage locations, and the balance the document is stamped against. It stopped there. `InventoryTransactionControllerWeb` (`/api/v1/inventory-transactions/web`) is still `system-admin` on all fourteen methods, reads included.

So a project manager can approve a document, and approving is what writes the `ADJUST` rows carrying the reason the stock moved, but cannot then read those rows. The person who signed the correction cannot see what the correction did.

This is the same family as #666 and the same argument would carry, but it is a different surface: the ledger is every movement in the organization, not the one balance the document names, so widening it is a larger decision than widening a stock lookup. Two shapes worth weighing:

- Widen the ledger reads to `system-admin`, `project-manager` the way materials now are, and accept that a project manager sees every receipt, issue and transfer.
- Scope the read instead, so a project manager sees the transactions on projects they are on, and leave the unscoped listing with `system-admin`. More work, and it needs a decision on what "on a project" means here.

Neither is urgent: nothing 403s in the stock-adjustment flow itself after #666. It surfaces the first time somebody wants the ledger view behind the document they just approved.

Related: #650, which asks the wider question of what role the Resources domain should have between member and system-admin.

abhijithanandan:
Closes #672.

## The gap

A project manager raises and approves stock adjustments, and approving one is what writes the `ADJUST` rows carrying the reason the stock moved. #666 gave that role the material, the storage location and the balance the document is stamped against, and deliberately stopped there. `InventoryTransactionControllerWeb` stayed `system-admin` on all fourteen methods, so the person who signed a correction could not read what the correction did.

It is already visible in the product. The material detail page (`/users/dashboard/resources/materials/[id]`) became reachable by a project manager when #666 moved the material reads, and two of its tabs call this controller:

| screen | hook | endpoint |
|---|---|---|
| Material detail page | `useMaterialStock` | `GET /material/{materialId}/stock` |
| Material, movement timeline tab | `useMaterialMovementHistory` | `GET /material/{materialId}/history` |
| Material, stock by location tab | `useInventoryTransactionsByStorageLocationAndMaterial` | `GET /storage-location/{id}/material/{id}/project/{id}` |
| Storage location, stock tab | `useStorageLocationStock` | `GET /storage-location/{id}/stock` |

All four answered 403 to the role the pages were opened for. They are the only endpoints on this controller `echno-web` calls today, and all four are in the widened set below.

## What changed

All fourteen methods are reads, so there is no write to widen. The split is drawn by what a read names.

**Moved to `system-admin`, `project-manager` (10):** `GET /{id}`, `/material/{materialId}`, `/material/{materialId}/history`, `/material/{materialId}/stock`, `/project/{projectId}`, `/project/{projectId}/task-summary`, `/storage-location/{id}`, `/storage-location/{id}/stock`, `/storage-location/{id}/material/{id}/project/{id}`, `/task/{taskId}`.

A project manager can already open every material, storage location, project and task, so a ledger read anchored to one of them shows the movements behind something they can already see.

**Held at `system-admin` (4):** `GET` (unpaginated), `GET /all`, `/type/{transactionType}`, `/date-range`. These name nothing and range over the whole organization.

Nothing is dropped to plain tenant membership.

## The line I did not pretend to draw

Holding those four is about which screen the ledger belongs to, not about confidentiality, and the commit says so. A caller who may list every material may walk the anchored reads material by material and reassemble the same rows. What it buys is that the movement report stays an admin surface until somebody decides otherwise, instead of being widened as a side effect of fixing an approval workflow.

Per-project scoping (the second shape the issue sketches) was considered and rejected for this PR. Roles are Keycloak groups on the organization; there is no per-project authorization anywhere in the codebase to build on, and no `@PostFilter` infrastructure. Inventing the first instance of it is a design decision, not a bug fix.

## Escalated separately

Widening these reads gives a project manager quantities, reference numbers and `unitCost` for movements written by goods receipts, purchase orders, indents, site transfers and material consumptions, while every one of those five document controllers remains `system-admin` on all of its methods. That contradiction is a product decision and is filed as its own issue rather than settled here.

## Tests

New `InventoryTransactionControllerWebAuthzTest`, a `@WebMvcTest` slice on the same pattern as `MaterialControllerWebAuthzTest` and `StorageLocationControllerWebAuthzTest`.

Confirmed red before the guard change (21 tests, 10 failed), green after:

| test | before | after |
|---|---|---|
| `aProjectManagerMayReadTheLedgerBehindAThingTheyCanOpen` (10 paths) | FAILED (403) | PASSED |
| `aProjectManagerIsStillRefusedTheOrganizationWideListings` (4 paths) | PASSED | PASSED |
| `aSystemAdminStillReadsTheWholeLedger` (all 14) | PASSED | PASSED |
| `aMemberHoldingNeitherRoleIsStillRefusedTheAnchoredRead` | PASSED | PASSED |
| `everyMappingOnTheControllerIsClaimedByExactlyOneOfTheTwoLists` | PASSED | PASSED |

The last three are regression guards and are green on both sides by design, so they are reported as such rather than as evidence of the fix. The invariant test exists because a parameterized pair of path lists invites asserting the split twice over nine paths and leaving the tenth governed by neither: it reflects the handler count off the controller and fails if a mapping is added without being placed on one side of the line.

## OpenAPI snapshot

The `@Tag` description is the only part of this change the document carries, and `docs/openapi.json` is updated with the byte-identical string. It is hand-applied rather than produced by `./gradlew openApiSnapshot -PupdateOpenApiSnapshot`, because that task boots Testcontainers and no lab node was reachable while this was written (the subnet router answers but times out dialing port 22 on every host on `10.24.6.0/24`). The Java literal and the JSON value were compared programmatically and are equal; CI regenerating the snapshot is the check that matters.


abhijithanandan:
@coderabbitai review

abhijithanandan:
Escalated the exposure question this opens as #680: a project manager now reads procurement movements through the ledger while every GRN, purchase order, indent, site transfer and material consumption controller stays system-admin on all of its methods. Deliberately not settled in this PR.

PR #682, opened 2026-09-06
Seed the curated Kerala and Tamil Nadu statutory compliance catalogue

abhijithanandan:
## What this is

The `ComplianceRule` catalogue the compliance module has been waiting on since August.

The module itself (backend #398, `echno-core` 1.4.x, `echno-web` #218) has been built and deployed for weeks. The one item recorded as outstanding was "Anand curates the ComplianceRule seed". **That catalogue was delivered on 27 August** as a PDF attachment on ClickUp `86d45jgq3` (*Compliance Document: Legal KL + TN Pilot*), and nobody joined the two up. Section 21 of that document, the Master Compliance Table, is the seed: 30 rules, 12 national (`NAT-*`), 9 Kerala (`KL-*`), 9 Tamil Nadu (`TN-*`).

What is live today is the 037 placeholder set: 21 rules written to exercise state and project-type matching, with no governing Act, no section, no applicability threshold, and an authority no more specific than "Local Body". It was never meant to reach a customer.

## Transcribed, not authored

Every legal fact here is copied from the specification. No threshold was completed from elsewhere and no citation was corrected. A compliance threshold invented in a product that tells a builder they are compliant is a real harm, so where the specification is incomplete this PR leaves it incomplete and flags it.

The specification's Requires Legal Verification section holds **11 open questions, not rules**. Nine of them attach to a rule that is in Section 21 (`NAT-01`, `NAT-02`, `NAT-04`, `NAT-08`, `NAT-11`, `NAT-12`, `KL-04`, `KL-05`, `TN-03`, `TN-04`, `TN-05`, `TN-08`, twelve rules in total); the other two, tree-felling permission and the EPF/ESI thresholds, have no row in Section 21 at all and so are **not seeded in any form**.

Those nine notes are transcribed into the affected rule's own `description`, so the caveat travels with the rule instead of living in a document nobody reads at the point of use. Nothing is seeded as settled that the source does not treat as settled: each of those twelve rules is a confirmed statutory requirement whose *detail* (an exemption position, a categorisation threshold, a renewal cycle) is flagged as unconfirmed. Dropping them would have removed Environmental Clearance and BOCW registration from the checklist, which is the worse error.

## Mapping onto the live schema

The specification's data model has 19 fields; `compliance_rules` has 10. There is no column for governing law, applicability, required evidence, validity, responsible stakeholder or official source. All of them are folded into the free-text `description`, which is also what `OpenAiCompatibleComplianceService` puts in the prompt, so nothing is lost to the model. Nothing queries them, so expiry still cannot be tracked off this table and an applicability threshold still cannot be evaluated in SQL. Both were already true and neither is made worse here.

`default_risk_level` is `NOT NULL` and the specification has **no risk column at all**, so a value had to be chosen. The rule applied uniformly:

| Level | Meaning |
|---|---|
| `CRITICAL` | gates lawful commencement, sale or occupation |
| `HIGH` | a statutory precondition or licence from a named authority |
| `MEDIUM` | a continuing payment, filing, record-keeping or connection obligation |

That mapping is a judgement, not a transcription, and wants confirming by whoever curates the catalogue.

## Shape of the seed

`compliance_rules` is keyed by `(state, project_type, code)` with no national scope and no all-types value, so the 12 national rules are stored once per pilot jurisdiction. Only `RESIDENTIAL` and `COMMERCIAL` are seeded: the specification verified its items for building and construction projects and asserts nothing about industrial, infrastructure or institutional work, and seeding those would claim a verification never done. 30 specification entries become 84 rows.

The 10 placeholders the catalogue supersedes are **deactivated, not deleted**, so generated inspections still resolve the `compliance_rule_ref` they were stamped with. `KL-STRUCT-STAB` and `TN-STRUCT-STAB` stay active on purpose: a structural stability certificate is not in the specification, nothing here supersedes them, and switching off a requirement this catalogue does not cover would quietly narrow what a project is told to obtain. Maharashtra is untouched, being outside the pilot and the only catalogue those projects have.

Idempotent by `ON CONFLICT (state, project_type, code) DO NOTHING`, so it applies cleanly to an environment already holding some of these rows and re-runs without duplicating anything.

## One deliberate consequence worth reading before merge

`effective_from` is set to now rather than back-dated. Per that column's own contract, a back-dated rule reaches nobody who has already been assessed, which is wrong for a rule that is new to us. **So the nightly sweep will find every approved project in every tenant stale against Kerala and Tamil Nadu and re-assess it, one model call each.** That is the intended outcome, because those projects were assessed against the placeholder catalogue and their answers are not worth keeping. It is not free, and it should be a conscious decision rather than a surprise.

Related: candidate rules per jurisdiction go from 6-9 to 21, so a generation run is now 3 batches instead of 1 (roughly 20-35s at the measured 1.7s per rule with concurrent batches).

## Open questions for the catalogue's curator

These are the specification's, not this PR's, and none of them are repaired here:

1. **Three overlapping pairs.** `NAT-01` (RERA registration) vs `KL-02`/`TN-02`; `NAT-04` (pollution consent) vs `KL-04`/`TN-04`; `NAT-06` (BOCW registration) vs `KL-05`/`TN-05`. Each pair is the same obligation, once nationally and once naming the state authority. Both are seeded because the specification's own worked checklist shows a Kerala project carrying the national and the state entry together, but it will read as a duplicate line item.
2. **`TN-03` and `TN-08` name no statute**, citing only a "State Fire Service framework", where the Kerala equivalents cite the Kerala Fire Force Act, 1962.
3. **`KL-05` is used for two different things** in the document: BOCW registration in Section 21, and a local body plinth-level inspection in the Section 20 example checklist.
4. **Cross-references are off throughout the document.** The text says "Section 20" for the master table (it is 21), "Section 22" for the legal-verification list (it is 24), "Section 7" for project attributes (it is 6). Content is unaffected.

## Verification

`ComplianceRuleSeedIT` runs the changelog against a real CockroachDB v26.2.4 via Testcontainers and asserts the catalogue a project actually sees. All 7 pass:

- every pilot jurisdiction carries the 12 national and 9 state rules
- Kerala rules do not leak into Tamil Nadu, and the reverse
- superseded placeholders are no longer candidates, but their rows survive
- the structural-stability placeholders stay active
- Maharashtra is untouched
- every seeded rule carries the fields the generation call reads
- reapplying the seed inserts nothing

Not deployed. `deploy-staging-iitm.yml` on `development` is the step, and it is deliberately left for a human.

### DIFF
--- a/src/main/resources/db/changelog/db.changelog-master.xml
+++ b/src/main/resources/db/changelog/db.changelog-master.xml
@@ -400,4 +400,7 @@
     <!-- v4.0 - Issues: keep the priority the issue form has been submitting all along -->
     <include file="db/changelog/v4.0/088-add-issue-priority.xml"/>
 
+    <!-- v4.0 - Compliance: the curated Kerala and Tamil Nadu statutory catalogue, replacing the 037 placeholders -->
+    <include file="db/changelog/v4.0/089-seed-compliance-rules-kl-tn.xml"/>
+
 </databaseChangeLog>
--- a/src/main/resources/db/changelog/v4.0/089-seed-compliance-rules-kl-tn.xml
+++ b/src/main/resources/db/changelog/v4.0/089-seed-compliance-rules-kl-tn.xml
@@ -1,0 +1,392 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <!--
+        The curated statutory catalogue for the Kerala and Tamil Nadu pilot, transcribed from
+        the Master Compliance Table of the ECHNO Compliance Module product and technical
+        specification (version 1.0 draft, August 2026). Thirty rules: twelve national, nine
+        Kerala, nine Tamil Nadu.
+
+        This is the catalogue the compliance module was built against and has been waiting on.
+        What 037 seeded was a placeholder set written to exercise state and project-type
+        matching: it carries no statutory citation, no governing Act and no issuing authority
+        beyond a rough label. It was never the real catalogue and was never meant to reach a
+        customer.
+
+        TRANSCRIBED, NOT AUTHORED. Every legal fact below (the governing Act and section, the
+        authority, the applicability threshold, the evidence, the validity and renewal position
+        and the source citation) is copied from the specification. Nothing was completed from
+        elsewhere and no citation was corrected. Where the specification records a requirement
+        as only partly verified, that note is carried into the rule's description rather than
+        dropped, so the caveat travels with the rule instead of living in a document nobody
+        reads at the point of use.
+
+        WHAT THE SPECIFICATION SUPPLIES THAT THIS TABLE CANNOT HOLD. The specification's data
+        model has nineteen fields; compliance_rules has ten. There is no column for the
+        governing law, the applicability condition, the required evidence, the validity period,
+        the responsible stakeholder or the official source. All of them are folded into the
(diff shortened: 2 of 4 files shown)


abhijithanandan:
@coderabbitai review

Issue #684, opened 2026-09-06
Notification and leave-calendar endpoints are gated on 'leave:read' and 'leave:admin', authorities this realm has no mechanism to issue


abhijithanandan:
Same phantom-authority shape as #641, in the leave package, on eleven endpoints that are still live.

## Why the authority cannot be issued

`JwtAuthConverter` mints a bare `resource:scope` authority in exactly one place, `extractPermissions`, which reads the `authorization` claim of an RPT. `leave:read` and `leave:admin` therefore need a Keycloak Authorization Services resource named `leave` carrying the matching scope. The multi-tenancy audit of 2026-08-18 traced the identical `organization:admin` string end to end against the live staging Keycloak and found zero authorization scopes realm-wide, one scopeless `Default Resource` on `echno-backend-client`, and one scopeless `Default Permission`. A scopeless permission yields no `resource:scope` authority at all. #641 confirmed the same for `billing:admin` and fixed the billing surface. The same holds unchanged here.

## Affected

`NotificationController` (`/api/v1/notifications`), all five:

- `GET /` , `GET /unread`, `GET /unread-count`, `PATCH /{notificationId}/read`, `POST /mark-all-read`

`LeaveCalendarController` (`/api/v1/leave-calendar`), all six.

Every one refuses every caller and has done since the annotations were written.

## The notification half carries a second defect underneath the first

Four of the five take `employeeId` from the query string and pass it straight to the service, under a guard that reads nothing. That is the shape closed in #589, #599, #607, #631, #635 and #683. Once the guard is repaired, whoever gets through reads any colleague's notifications and marks them read.

`markAsRead(notificationId)` is worse in kind: it takes only the notification id and checks no ownership at all, so it does not even have a parameter to be wrong about. It needs an owner check against the stored notification, not a repaired guard.

## Why this is worth doing rather than deleting

Notifications are how an approver learns that a leave request is waiting on them. `LeaveApprovalService` calls `sendApprovalRequiredNotification` on every routing step and `sendDelegationNotification` on every handover, so the rows are being written and nobody can read them. With mobile approval now a required workflow (ClickUp `86d45jxmn`), the alert channel that feeds it should work.

The calendar half is a plain read surface and needs only the guard corrected.

## Not fixed in the mobile approval work

`feat/mobile-leave-approval` deliberately stops at the approval workflow and its queue. This is a different surface with its own ownership question and its own test surface, and folding it in would have made that change hard to review.

Issue #683, opened 2026-09-06
A leave approver's pending queue is read from a query parameter, so an administrator reads anyone's and the approver cannot read their own


abhijithanandan:
Sixth instance of the family closed in #589, #599, #607, #631 and #635, and the first one where the consequence is that the feature does not work rather than that somebody can reach too far.

## The endpoints

| Endpoint | Guard | Query |
| --- | --- | --- |
| `GET /api/v1/leave-requests/pending-approvals` | `hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin')` | `findByCurrentApproverIdAndStatus(approverId, PENDING_APPROVAL)` |
| `GET /api/v1/leave-requests/pending-approvals/count` | same | `countByCurrentApproverIdAndStatus(approverId, ...)` |
| `GET /api/v1/leave-requests/web/pending-approvals` | same | same |
| `GET /api/v1/leave-requests/web/pending-approvals/count` | same | same |
| `GET /api/v1/leave-requests/web/approver` | same | `findDistinctByApproverParticipation(approverId)` |
| `GET /api/v1/leave-approvals/{web,requests/{id}}/can-approve` | phantom `leave:approve` authority, or the role pair on the web twin | `canApprove(requestId, employeeId)` |

`approverId` and `employeeId` come off the query string. The guard never reads them. So the guard establishes that the caller holds a role and the query answers about whoever the caller named, and nothing connects the two.

## Both halves are wrong, and the second half is the one that matters here

**Too far.** Any holder of system-admin or hr-admin lists any colleague's pending leave queue, and their whole history of requests they were an approver on, by passing that colleague's employee id. `can-approve` lets one employee ask whether another is the current approver of a request.

**Not far enough, which is the bigger problem.** `LeaveApprovalService.resolveApprovalChain` builds the approval chain by walking `employee.getManager()`. The people who hold leave decisions are therefore line managers, and a line manager holds neither `system-admin` nor `hr-admin`. The role gate refuses them their own queue. So the approver cannot see what is waiting on them, and the only people who can see it are the ones who are not deciding.

## Fix

A queue is the caller's own by definition. All six take the approver from the session rather than the query string, and are gated on tenant membership. Dropping the parameter is safe on the deployed client because Spring ignores a query parameter no handler declares, so a call that still sends `approverId` is served the caller's own queue rather than refused.

[...]

PR #685, opened 2026-09-06
Let the person holding a leave decision make it

abhijithanandan:
Anand asked for mobile leave approval to be treated as a required workflow rather than an optional capability (ClickUp `86d45jxmn`, 1 Sep). It was neither: it was refused.

## What was refused, and why

Five of the six endpoints on `LeaveApprovalController` asked for `hasAuthority('leave:approve')`, `'leave:read'` or `'leave:admin'`:

| Endpoint | Guard before | Reachable by |
| --- | --- | --- |
| `POST /leave-approvals/requests/{id}/approve` | `@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin')` | an administrator who also happened to be the current approver |
| `POST /leave-approvals/requests/{id}/reject` | `hasAuthority('leave:approve') or hasAuthority('leave:admin')` | nobody |
| `POST /leave-approvals/requests/{id}/delegate` | `hasAuthority('leave:approve') or hasAuthority('leave:admin')` | nobody |
| `GET /leave-approvals/requests/{id}/history` | `hasAuthority('leave:read') or hasAuthority('leave:admin')` | nobody |
| `GET /leave-approvals/requests/{id}/chain` | `hasAuthority('leave:read') or hasAuthority('leave:admin')` | nobody |
| `GET /leave-approvals/requests/{id}/can-approve` | `hasAuthority('leave:approve') or hasAuthority('leave:admin')` | nobody |

`JwtAuthConverter` mints a bare `resource:scope` authority in exactly one place, `extractPermissions`, which reads the `authorization` claim of an RPT. That needs a Keycloak Authorization Services resource named `leave` carrying the matching scope. The multi-tenancy audit of 2026-08-18 traced the identical `organization:admin` string against the live staging Keycloak and found zero authorization scopes realm-wide, one scopeless `Default Resource` and one scopeless `Default Permission`; a scopeless permission yields no `resource:scope` authority at all. #641 confirmed the same for `billing:admin`. Nothing grants these strings and nothing can.

## Approve had the subtler version of the same problem, and so did the web twin

`approve` was guarded on the system-admin or hr-admin role, while `LeaveApprovalService.validateApprover` requires the caller to be the request's **current approver**, and `resolveApprovalChain` builds that chain by walking `employee.getManager()`. A site manager is an approver and holds neither role, so the guard refused them. An administrator who was not in the chain got past the guard and was refused by the service. The two halves could only both be satisfied by the accident of an approver who also happened to be an administrator.

`LeaveApprovalControllerWeb` carried that same role gate on all six of its endpoints, so this was never only a mobile problem: the web console had it too, and the approvals screen worked only for administrators.

## The fix

Every endpoint on both twins is gated on `@orgSecurity.isMemberOfCurrentTenant()`, which is what the annotation can actually evaluate, with who may act settled in the service against the record. **This is narrower than the role gate it replaces, not wider.** Before, any holder of system-admin or hr-admin reached the handler; now the current approver acts and nobody else does.

The approval trail needed its own answer rather than a role. An approver at level two has no account of what level one said unless they can read the trail, so refusing them the trail refuses them the decision, which is the shape #666 was. It is readable by the employee the leave belongs to, everybody named anywhere in the chain (approver or `delegatedFromId`), and the leave administrators.

## The queue was another instance of the family closed in #589, #599, #607, #631 and #635

`GET /leave-requests/pending-approvals`, its `/count` and the web `/approver` listing took the approver as a query parameter under a guard that only asked for a role. The guard checked a role, the query read a number the caller chose, and nothing tied the two together: an administrator listed any colleague's queue by asking for it, and the managers who actually hold these decisions could not list their own. A queue is the caller's own, so it comes from the session now. `can-approve` was the same shape, letting one employee probe another's place in a chain; it answers about the caller.

Filed separately as #683.

## The approver also could not open the request

Reading one leave request, `GET /leave-requests/requestId/{id}` and its web twin, was gated on the system-admin and hr-admin roles. That is the read the request detail screen is built on, and the approve, reject and delegate buttons live on that screen. Repairing the actions and leaving this would have repaired the endpoints that return 403 and left the approver unable to reach the form, which is what #666 did.

Same rule, same code: `requireMayReadTrail` is reachable as `requireMayReadRequest` and `LeaveRequestService.getRequest` calls it, so there is one answer to who takes part in a request rather than two that drift. A draft has no approval rows, so only its owner and the administrators read it.

Worth noting for anyone tracing an old bug report: this also means a `project-manager` has been getting a 403 from that endpoint all along.

## Removing those parameters is safe on the deployed client

Spring drops a query parameter no handler declares, so a call that still sends `approverId` or `employeeId` is served the caller's own answer rather than refused. That is what lets this ship without waiting on a frontend release, and two tests pin it.

## Tests

Twenty-one assertions were **measured** failing against the old code, by running the new tests on `origin/development` in a throwaway worktree. The request-read work landed after that run, so its cases are reasoned rather than measured, and the reasoning is separated out below rather than folded into the count.

| Test | Cases failing on `origin/development` | How they fail |
| --- | --- | --- |
| `LeaveApprovalWorkflowAuthzTest` (as `BaselineApproverReachTest`) | 17 | a site manager holding no administrative role gets 403 on all 11 read paths and all 6 action paths |
| `LeaveApprovalEndpointAuthorityTest.noLeaveApprovalEndpointIsGuardedByAnAuthorityTheRealmCannotIssue` | 1 | names the 5 phantom-guarded endpoints |
| `LeaveApprovalEndpointAuthorityTest.actingOnAnApprovalIsGatedOnMembership...` | 1 | the guard expressions are the old ones |
| `LeaveApprovalEndpointAuthorityTest.anApprovalQueueTakesNoApproverParameter` | 1 | five handlers still declare `approverId` |
| `LeaveApprovalEndpointAuthorityTest.theCanApproveCheckAsksAboutTheCaller...` | 1 | both twins still declare `employeeId` |

`everyLeaveApprovalEndpointIsStillGuardedBySomething` pins the endpoint count so the first rule cannot pass by a guard having been deleted rather than corrected. It passes on both sides by design, which is the point of it.

Added after that baseline run, so reasoned rather than measured: the two request-read paths in the allow list fail on the old code, because the old guard was the system-admin and hr-admin pair and the site manager stub answers false to it. The two request-read paths in the refusal list would have passed on the old code as well, since a non-member was refused either way, and they are there to stop the widening going too far rather than to demonstrate the fix. The two service cases in `LeaveApproverQueueTest` cannot be compiled against the old code, because `requireMayReadRequest` did not exist.

`LeaveApprovalTrailAccessTest` (12 cases) and `LeaveApproverQueueTest` (7 cases) cover the service-side rules. They cannot be compiled against the old code, because the old signatures took the id from the caller, which is the defect itself.

## The API contract

`docs/openapi.json` is regenerated and committed. Seven endpoints lose a query parameter and the operation, tag and 403 descriptions across the approval workflow are rewritten: 54 insertions, 127 deletions.

There is no container runtime on this workstation and the lab node that has one was unreachable all session, so the document came from the build itself. The snapshot test writes one canonical string and either commits it or writes it to `build/openapi.json` for diffing, and the workflow uploads that file, so what the run served is what is committed here. CI re-verifies it, which is the check that matters.

## Left alone deliberately

`NotificationController` and `LeaveCalendarController` in the same package still carry the phantom authority on eleven endpoints, and the notification endpoints have the caller-supplied `employeeId` shape on top of it, plus a `markAsRead` that checks no ownership at all. Separate surface, separate ownership question, filed as #684.

https://claude.ai/code/session_01YWbRuHXKpFT6jL3mzYNhhD

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -60907,7 +60907,7 @@
     },
     "/api/v1/leave-approvals/requests/{requestId}/approve": {
       "post": {
-        "description": "Records an approval for the given request at its current approval level and advances the workflow. Returns the request with its updated status and approval trail.",
+        "description": "Records an approval for the given request at its current approval level and advances the workflow. The approval is recorded against the signed-in caller, who must be the request's current approver. Returns the request with its updated status and approval trail.",
         "operationId": "approve_1",
         "parameters": [
           {
@@ -60949,7 +60949,7 @@
                 }
               }
             },
-            "description": "The approval action payload failed validation"
+            "description": "The approval action payload failed validation, or the caller is not the request's current approver, or the request is not pending approval"
           },
           "402": {
             "content": {
@@ -60969,7 +60969,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant, or has no employee record in it, so the record would name nobody"
+            "description": "Caller is not a member of the current tenant, or has no employee record in it, so the decision would name nobody"
           },
           "404": {
             "content": {
@@ -61030,21 +61030,12 @@
     },
     "/api/v1/leave-approvals/requests/{requestId}/can-approve": {
       "get": {
-        "description": "Returns whether the given employee is the current approver for the given leave request, as a single boolean flag.",
+        "description": "Returns whether the signed-in caller is the current approver for the given leave request, as a single boolean flag. This is the check a client makes before drawing an approve button. It used to take the employee to ask about as a query parameter, which let one employee probe another's place in a chain; a caller that still sends employeeId is answered for themselves, because a query parameter no handler declares is ignored.",
         "operationId": "canApprove_1",
         "parameters": [
           {
             "in": "path",
             "name": "requestId",
-            "required": true,
-            "schema": {
-              "format": "int64",
-              "type": "integer"
-            }
-          },
-          {
-            "in": "query",
-            "name": "employeeId",
             "required": true,
             "schema": {
               "format": "int64",
@@ -61094,7 +61085,7 @@
                 }
               }
             },
-            "description": "Caller lacks the leave approve or admin authority"
+            "description": "Caller is not a member of the current tenant"
           },
           "404": {
             "content": {
@@ -61147,7 +61138,7 @@
             "description": "Bad Gateway"
           }
         },
-        "summary": "Check whether an employee can approve a request",
+        "summary": "Check whether you can approve a request",
         "tags": [
           "Leave Approvals"
         ]
@@ -61210,7 +61201,7 @@
                 }
               }
             },
-            "description": "Caller lacks the leave read or admin authority"
+            "description": "Caller is not a member of the current tenant, or takes no part in this request and holds neither the system-admin nor the hr-admin role"
           },
           "404": {
             "content": {
@@ -61271,7 +61262,7 @@
     },
     "/api/v1/leave-approvals/requests/{requestId}/delegate": {
       "post": {
-        "description": "Reassigns the current approval step for the given request to the delegate named in the payload. Returns the request with its updated approver and approval trail.",
+        "description": "Reassigns the current approval step for the given request to the delegate named in the payload, recording who handed it over. Only the request's current approver may delegate it. Returns the request with its updated approver and approval trail.",
         "operationId": "delegate_1",
         "parameters": [
           {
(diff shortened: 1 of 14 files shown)


abhijithanandan:
@coderabbitai review

PR #694, opened 2026-09-06
Let a stock adjustment name the document it was raised to answer

abhijithanandan:
Backend half of tornotron/echno-web#398. The web half is a separate PR and depends on this contract.

## Why

A cross-project site transfer received short leaves an open variance. The sending site is down the full sent quantity, the receiving site is up what arrived, and the difference is unaccounted for. The transfer deliberately writes no loss movement for it (#660), because a loss written automatically is a stock correction nobody authorised. The document that decides what became of the difference is a stock adjustment, and until now it had nowhere to say which transfer it answered. So the correction and its cause could not be read from either end: the transfer's variance stayed amber for ever, and the adjustment read as an unexplained count variance.

web#392 shows and names the variance but offers no route to the adjustment, and that was left out on purpose: a link built against the web mock shape's `transferId` would have produced a prefilled form that silently dropped the reference on submit.

## The shape, and why not a foreign key

`sourceDocumentType` (enum) plus `sourceDocumentId` (long), both nullable, written together or not at all.

A transfer variance is the first cause with a document behind it and it will not be the last: a goods receipt and a consumption are the obvious next two. A nullable foreign key per cause ends as a row of mutually exclusive columns with nothing enforcing that only one of them is set. The schema already carries a type-and-id key twice, in `status_transition` and `attachment`, and for the further reason those give: a posted adjustment is a ledger fact and has to outlive the document that caused it, so deleting a transfer must not take the record of the correction raised against it.

The departure from those two is that this is an enum rather than a free string. What the pair gives up is the database refusing an id that names nothing, so the write path does it instead: every value has a resolver that loads the named document within the caller's organization, and a value with no resolver is refused rather than stored. Adding a source type therefore means adding its resolver, which is deliberate.

## The reference is not taken on trust

The id the caller sent is the id that is looked up, organization scoped, and the id that is written. There is no second key for a guard and a write to disagree about, which is the defect shape this repo keeps finding. The reverse read is scoped in the query for the same reason: matching on a caller supplied id against a column with no foreign key behind it would otherwise reach across tenants.

A cancelled transfer is refused as a cause. Cancelling is reachable only from `PENDING`, which is the state in which nothing has been received, and it returns the whole sent quantity to the sending site, so it left no variance to close. Its lines still read as in transit, and that figure is history rather than a live shortage.

## Authorization

`GET /api/v1/stock-adjustments/web/by-source-document` reads under the same guard as the other stock adjustment reads (any tenant member), so whoever received the transfer can also see the adjustment they are being sent to raise. Receiving a transfer needs `system-admin`, and creating an adjustment needs `system-admin` or `project-manager`, so the whole path holds for the person who is actually offered the link.

One mismatch found while tracing that path is filed separately rather than fixed here: a `project-manager` may raise and approve a stock adjustment but may not read a site transfer at all, so they cannot follow a source reference back to its transfer. Widening site transfer reads is a decision of its own and does not belong in this change.

## Tests

New `StockAdjustmentSourceDocumentTest`, 11 cases, plus three cases on the controller authorization slice. Each was confirmed red by reverting the specific line it covers.

`docs/openapi.json` is regenerated on a runner: this workstation has no container runtime, so the snapshot test cannot be run locally.

https://claude.ai/code/session_01YWbRuHXKpFT6jL3mzYNhhD

## Summary by CodeRabbit

- **New Features**
  - Stock adjustments can now reference the source document they address, currently supporting site transfers.
  - Added an endpoint to retrieve stock adjustments associated with a specific source document.
  - Source documents are validated for organization access and eligible status when creating or updating adjustments.
  - API responses now include source document type and ID when available.

- **Bug Fixes**
  - Prevented cross-organization source-document matches and invalid document references.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -16405,6 +16405,26 @@
             "format": "int64",
             "type": "integer"
           },
+          "sourceDocumentId": {
+            "description": "Id of the document named by sourceDocumentType. Required when that field is set, and refused when it is not.",
+            "example": 31,
+            "format": "int64",
+            "type": [
+              "integer",
+              "null"
+            ]
+          },
+          "sourceDocumentType": {
+            "description": "The kind of document this adjustment was raised to answer, where it answers one. Send it together with sourceDocumentId or leave both out: a type naming no document, or a document with nothing saying what kind it is, is refused rather than half-stored. The named document is looked up in the caller's organization when the adjustment is written, so an id that names nothing there is a 404 and not a dangling reference.",
+            "enum": [
+              "SITE_TRANSFER"
+            ],
+            "example": "SITE_TRANSFER",
+            "type": [
+              "string",
+              "null"
+            ]
+          },
           "status": {
             "description": "State of the adjustment document. Optional, and draft is the only value accepted, so leaving it out is the same as sending draft. A document reaches its posted state through POST /{id}/approve, which is what refuses an approval by whoever raised it and writes the stock-ledger entries; a document created or edited into that state would carry neither, and would read as approved with nobody on record as having approved it and no movement behind it.",
             "enum": [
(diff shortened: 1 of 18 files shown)


abhijithanandan:
On the two review findings.

**The Liquibase guard is fixed.** Each column now has its own change set guarded on itself. The point stands: guarding both columns on the presence of one of them marks the change set as ran on a database holding only the first, and leaves the entity mapped to a column that is not there.

**The nullable enum is real but cannot be fixed here.** `"type": ["string", "null"]` alongside `"enum": ["SITE_TRANSFER"]` does refuse null under JSON Schema, so the two disagree. But this file is generated, and CI verifies it byte for byte against the document the application actually serves, so hand-adding `null` to the enum would fail the build rather than fix anything.

It is also not something this change introduced. Every nullable-typed enum schema in the document has this shape:

```
nullable-typed enum schemas in the whole document: 9
  of which the enum also lists null: 0
```

The other seven are on `EmployeeUpdateFieldsDto.status`, `IssueUpdateFieldsDto.priority`, `LeaveRequestUpdateFieldsDto.startHalfDayType` and `endHalfDayType`, `ProjectUpdateFieldsDto.projectType` and `status`. It is springdoc's uniform output for a nullable enum under OpenAPI 3.1, so the fix belongs in a schema customiser applied across all nine, not in one hand edit to a generated file that the next regeneration would undo.

PR #696, opened 2026-09-06
Mint an invite code into the caller's own organization, not one they name

### LINKED ISSUE #687: An invite code is minted into whichever organization the caller names, and Organization is the one entity the tenant filter does not cover
Found while sweeping for further instances of the family closed in #589, #599, #607, #631, #635, #683 and #684. This is the one where the consequence leaves the tenant.

## The endpoint

`POST /api/v1/invitation/web/generateCode/organizationId/{organizationId}`, `ProjectInviteCodeController` line 69.

```java
public ResponseEntity<ProjectInviteCodeDto> createInviteCode(
        @Valid @RequestBody InviteCodeGenerationDto dto, @PathVariable Long organizationId)
```

The guard asks whether the caller holds a role **in their own tenant**. It never reads `#organizationId`. `ProjectInviteCodeService.generateInviteCode` then does:

```java
Organization organization = organizationRepository.findById(organizationId)
        .orElseThrow(() -> new ResourceNotFoundException(...));
```

## Why the tenant filter does not catch this one

The usual answer to a caller-supplied `organizationId` in this codebase is that it is redundant rather than dangerous, because the Hibernate `orgFilter` scopes the query to the current tenant anyway. That is true of `ProjectInviteCode` itself, which is a `TenantScopedEntity` carrying `@Filter(name = "orgFilter", ...)`, and it is why the sibling read endpoint is a much smaller problem: `readAllProjectInviteCodes(organizationId)` against a foreign id returns empty rather than that organization's codes.

`Organization` is the exception. It is the tenant root, so it is **not** a `TenantScopedEntity` and carries no `@Filter`. `findById` returns any organization in the deployment. `TenantIsolationLoadListener` does not help either: it short-circuits on `!(event.getEntity() instanceof TenantScopedEntity)`, and that is the first line of the method.

So the invite code is constructed against organization B and saved with `organization_id = B`, by an administrator of organization A. The code value comes back in the `201` response body, so the caller does not need to read it back through the filtered listing.

The redemption path is separately reachable: `POST /api/v1/invitation/web/validate/userId/{userId}` is guarded by `@orgSecurity.isSelfUser(#userId)`, which the same caller satisfies for their own user id.

## What I verified and what I did not

Verified by reading the source: the guard expression, that it does not reference `#organizationId`, that the service resolves the organization by the caller's id with no `TenantContext` comparison, that `Organization` carries neither `TenantScopedEntity` nor `@Filter`, and that the load listener returns early for it.

Not verified: whether an invite code minted into another organization actually completes redemption end to end. That needs a run against a two-organization fixture, and it is the thing to establish first, because it decides whether this is a privilege-escalation path or only a data-integrity one.

## Rider in the same method

```java
employeeRepository.existsByIdAndOrgRolesIn(dto.getManagerId(), OrgRole.getManagerRoles())
```

has no organization predicate, so a `managerId` from another tenant satisfies the check whose error message claims it looked "with a manager role in this organization". `Employee` is filtered, so this is probably neutered in practice; the message is wrong either way.

## Fix direction

The organization is settled by the session. Take it from `TenantContext.getCurrentOrgId()` and drop the path segment, the way #683 and #684 did. The read endpoint at line 49 has the same shape and should move with it. `patchInviteCode` at line 118 already gets this right, using `findByIdAndOrganization_Id(..., TenantContext.getCurrentOrgId())`, so the correct pattern is in the same class.

Worth deciding separately: whether anything else in the codebase reaches `organizationRepository.findById` with a caller-supplied id, since `Organization` being outside the filter makes that a general hazard rather than a local one.

abhijithanandan:
Closes #687.

## The defect

`POST /api/v1/invitation/web/generateCode/organizationId/{organizationId}` took the organization from the path. Its guard, `@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin')`, establishes that the caller holds a role in the organization **their session is scoped to**, and never reads `#organizationId`. Those are different questions, and the second one was never asked. The service then resolved the named organization and bound a working invite code to it, returning the code value in the 201 body.

## Why nothing underneath caught it

Confirmed against the current code, since the fix only makes sense if this is right.

Tenant isolation here is a Hibernate `@Filter(name = "orgFilter")` on entities implementing `TenantScopedEntity`, enabled per `@Transactional` method by `HibernateFilterConfig` whenever `TenantContext.getCurrentOrgId()` is set, plus `TenantIsolationLoadListener`, which fail-closes on any loaded tenant-scoped row belonging to another organization. That pairing is why the other findings from the same sweep degrade to an empty list rather than a leak.

`Organization` sits outside both, and necessarily so, because it is the tenant root:

- it does not implement `TenantScopedEntity` and carries no `@Filter`, so `findById` resolves any organization in the deployment;
- `TenantIsolationLoadListener.onPostLoad` returns on its first line for anything that is not a `TenantScopedEntity`.

Both defences are therefore absent at exactly the point where a caller names an organization by id. This is a class of endpoint that has to check entitlement itself, not an annotation someone forgot. `TenantIsolationIT` now asserts that property against a real database, beside the cases that are covered, so the reasoning is checked rather than asserted in a comment.

## The fix

**A guard that establishes entitlement.** New `@orgSecurity.isCurrentTenant(organizationId)` compares the named organization against `TenantContext`. That is a real check rather than a restatement: `TenantFilter` only ever sets the current organization to one the caller holds an `ORG_MEMBER_` authority for, refusing the request outright otherwise. It fails closed on a null tenant and on a null argument. Paired with the existing role check, one answers where and the other answers what.

**A service that reads the same key.** `generateInviteCode` and `readAllProjectInviteCodes` drop their organization parameter and read `TenantContext`. The recurring shape in this repository is a guard checking one id while the service loads by another, so leaving the parameter in place would have left the next edit free to reintroduce it. The route keeps its path segment, so `echno-core` call sites are unchanged; the segment is now binding rather than decorative.

The listing moves for its own reason, not by symmetry: it hands back code values, so it decides who may read a credential, and that decision should not be resting on the entity's filter happening to cover it.

## Two more of the same shape

- `POST /api/v1/employee/web/joinOrganization/userId/{userId}/organizationId/{orgId}` carried the identical guard over an identical caller-supplied id, and its write is the worse one: it creates the employee row and adds the user to the Keycloak group for the named organization, which is the `ORG_MEMBER_` authority the whole tenant layer is built on. Same guard fix, applied here.
- The manager check inside generation refused with a message claiming the manager was looked for "in this organization" while the query carried no organization predicate. It now carries one, so the message is true wherever the method is used next rather than only where the Hibernate filter happens to be enabled.

The mobile twin of `joinOrganization` is guarded by a bare realm authority bound to no organization at all. Tightening it needs a decision about the mobile onboarding contract, so it is filed rather than changed here.

## Returning the code value

Kept, deliberately. The listing endpoint already returns code values to the same role, so withholding it from the 201 removes nothing an entitled caller cannot read a second later, while breaking the console flow that shows an admin the code they just created. The weaker part of this credential is its shape rather than where it is returned, and that is filed separately since changing it is a contract change across `echno-core` and the mobile app.

## Tests

|---|---|---|
| `ProjectInviteCodeTenantGuardTest.generate_forAnotherOrganization_isForbiddenAndNeverReachesTheService` | a role holder naming another organization is refused, before the service runs | **measured** (was 201) |
| `ProjectInviteCodeTenantGuardTest.list_forAnotherOrganization_isForbiddenAndNeverReachesTheService` | same for the listing | **measured** (was 200) |
| `ProjectInviteCodeTenantGuardTest.generate_forTheCallersOwnOrganization_isStillAllowed` | the ordinary case is unaffected | passed before and after, by design |
| `ProjectInviteCodeTenantGuardTest.list_forTheCallersOwnOrganization_isStillAllowed` | same for the listing | passed before and after, by design |
| `OrganizationSecurityServiceTest.isCurrentTenant_*` (3 cases) | matches only the organization in force; fails closed on a null tenant and a null argument | new surface, nothing to measure |
| `ProjectInviteCodeServiceTest.generateInviteCode_bindsTheCodeToTheSessionOrganization` | the persisted code carries the session's organization | reasoned; does not compile against the old signature |
| `ProjectInviteCodeServiceTest.generateInviteCode_withNoTenantInContext_mintsNothing` | no tenant means no organization load and no save | reasoned, same reason |
| `ProjectInviteCodeServiceTest.readAllProjectInviteCodes_*` (2 cases) | the listing is keyed on the session and refuses without one | reasoned, same reason |
| `ProjectInviteCodeServiceTest.generateInviteCode_checksTheManagerWithinTheSessionOrganization` | the manager query carries the organization its message claims | reasoned, same reason |
| `TenantIsolationIT.findById_onTheTenantRootItself_isNotCoveredByEitherMechanism` | the tenant root loads across organizations, against a real database | documents the mechanism; passes either way |

The two measured reds were run against the unmodified controller with only the new `isCurrentTenant` helper added, so the test file as committed is the one that was measured. The remaining service cases cannot compile against the old two-argument signature and are not counted as measured.

`TenantIsolationIT` needs Testcontainers, which this workstation has no Docker for; CI runs it.

## OpenAPI

Only three operation descriptions and their 403 descriptions changed, and no structure. There is no Docker here to run `openApiSnapshot`, so the three blocks in `docs/openapi.json` were edited in place to the exact strings the annotations now produce, leaving the canonical form untouched. CI re-verifies against the served document.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -33940,7 +33940,7 @@
     },
     "/api/v1/employee/web/joinOrganization/userId/{userId}/organizationId/{orgId}": {
       "post": {
-        "description": "Creates an employee record linking the given user to the given organization, with the employment details supplied in the request body.",
+        "description": "Creates an employee record linking the given user to the organization named in the path, which must be the organization the caller's session is scoped to. Uses the employment details supplied in the request body.",
         "operationId": "joinOrganization",
         "parameters": [
           {
@@ -34011,7 +34011,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller lacks the required role in the current tenant, or named an organization that is not it"
           },
           "404": {
             "content": {
@@ -55702,7 +55702,7 @@
     },
     "/api/v1/invitation/web/generateCode/organizationId/{organizationId}": {
       "post": {
-        "description": "Generates a new invite code for the given organization.",
+        "description": "Generates a new invite code for the organization named in the path, which must be the organization the caller's session is scoped to.",
         "operationId": "createInviteCode",
         "parameters": [
           {
@@ -55764,7 +55764,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller lacks the required role in the current tenant, or named an organization that is not it"
           },
           "404": {
             "content": {
@@ -55825,7 +55825,7 @@
     },
     "/api/v1/invitation/web/organizationId/{organizationId}": {
       "get": {
-        "description": "Returns every invite code generated for the given organization.",
+        "description": "Returns every invite code generated for the organization named in the path, which must be the organization the caller's session is scoped to.",
         "operationId": "readAllInviteCodes",
         "parameters": [
           {
@@ -55880,7 +55880,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller lacks the required role in the current tenant, or named an organization that is not it"
           },
           "404": {
             "content": {
--- a/src/main/java/org/tornotron/echno_backend/common/service/OrganizationSecurityService.java
+++ b/src/main/java/org/tornotron/echno_backend/common/service/OrganizationSecurityService.java
@@ -189,6 +189,37 @@ public class OrganizationSecurityService {
         return currentUserId != null && currentUserId.equals(userId);
     }
 
+    /**
+     * Checks that the organization the caller named is the one their session is scoped to.
+     *
+     * <p>For the endpoints that take an organization id from the caller and then load an
+     * {@code Organization} by it. That load is the one place tenant isolation cannot help:
+     * {@code Organization} is the tenant root, so it is not a {@code TenantScopedEntity},
+     * carries no {@code orgFilter}, and {@code TenantIsolationLoadListener} returns on its
+     * first line for anything that is not tenant-scoped. A caller-supplied id therefore
+     * resolves to any organization in the deployment unless a guard says otherwise.
+     *
+     * <p>Comparing against {@code TenantContext} is what makes it a check rather than a
+     * formality: {@code TenantFilter} only ever sets the current organization to one the
+     * caller holds an {@code ORG_MEMBER_} authority for, so equality with it establishes
+     * entitlement. Pair it with a role check, which answers what the caller may do; this
+     * answers where.
+     *
+     * <p>Usage in {@code @PreAuthorize}:
+     * {@code @orgSecurity.isCurrentTenant(#organizationId) and @orgSecurity.hasAnyOrgRoleForCurrentTenant('hr-admin')}
+     *
+     * @param organizationId the organization named by the caller
+     * @return true if it is the organization in force for this request
+     */
+    public boolean isCurrentTenant(Long organizationId) {
+        Long orgId = TenantContext.getCurrentOrgId();
+        boolean result = orgId != null && orgId.equals(organizationId);
+        if (!result) {
+            log.debug("Current-tenant check refused organization {} against tenant {}", organizationId, orgId);
+        }
+        return result;
+    }
+
     public boolean isMemberOfCurrentTenant() {
         Long orgId = TenantContext.getCurrentOrgId();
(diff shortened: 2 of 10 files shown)


abhijithanandan:
@coderabbitai review

Issue #691, opened 2026-09-06
The leave-policy listing takes the organization from the path and checks only that it exists


abhijithanandan:
Found while sweeping for further instances of the family closed in #589, #599, #607, #631, #635, #683 and #684. Smallest of the batch, filed on its own so the family keeps one entry per site.

## The endpoint

`GET /api/v1/leave-policies/organization/{organizationId}`, `LeavePolicyController:90`.

```java
@PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin')")
```

The guard establishes a role in the caller's own tenant and never reads `#organizationId`. `LeavePolicyService.getPoliciesByOrganization` (`:150`) and `getAllPoliciesByOrganization` (`:170`, reached with `includeInactive=true`) both do:

```java
if (!organizationRepository.existsById(organizationId)) throw ...;
return policyRepository.findByOrganizationIdAndIsActiveTrueOrderByDisplayOrderAsc(organizationId);
```

The only thing checked about the organization is that it exists.

## Scope of the consequence

`LeavePolicy` is a `TenantScopedEntity` carrying `@Filter(name = "orgFilter", ...)`, so a foreign id returns empty rather than another tenant's accrual rates, carry-forward caps and encashment rules. I checked the entity rather than reading `existsById` as a cross-tenant hole. So this is the redundant-parameter form of the family rather than the dangerous one, and it sits below #687 and #689 in priority.

It is still worth closing. The parameter serves no purpose, the `existsById` check reads as though it were doing tenant work and is not, and every one of these that stays makes the next audit slower.

## Note for whoever fixes it

The web twin, `LeavePolicyControllerWeb`, has this handler **commented out** at lines 84 to 95. So there is no live counterpart to keep in sync, and the choice is between scoping the mobile one to `TenantContext` and deleting it. Check what `echno-core` calls before deleting: the twin being commented out suggests the console reaches leave policies some other way, and that route is the one that should survive.

PR #701, opened 2026-09-06
Give the attendance approvers somewhere to find what is waiting on them

### LINKED ISSUE #692: There is no endpoint that lists the attendance records waiting on a caller, so the geofence routing has no inbox
Found while closing tornotron/echno-web#402.

Since #681 an attendance day can be routed to a named person: an employee who marks from outside the site boundary gives a reason, `AttendanceGeofenceService.resolveApprover` names who decides, and `AttendanceService.approve` accepts that person's decision through `canDecideApproval`. The decision works. Nothing can find it.

## What is missing

There is no query for "the attendance records waiting on me". The two list endpoints are:

- `GET /api/v1/attendance/web/project/{projectId}` — needs a `projectId` **and** a required `date`, filters on `AttendanceStatus`, and has no `approvalStatus` parameter.
- `GET /api/v1/attendance/web/employee/{employeeId}` — one employee, a date range.

So an approver can reach a record only by already knowing it exists, or by guessing a project and a date. An approver with people on several sites has to guess once per site per day. A record from three days ago is invisible to anyone not looking for it.

Confirmed against `docs/openapi.json` on `development`, not inferred from a status code: the paths under `/api/v1/attendance` are check-in, clock-event, employee/{employeeId}, mark-absent, project/{projectId}, summary/{employeeId}, {id} and {id}/approve, plus the `/web` twins of each. There is no pending-approvals path, and `@tornotron/echno-core` 8.1.0's `attendanceService` has no method for one either.

## The shape it should take

The leave module already has this, and it was corrected recently, so the shape is settled rather than open:

- `GET /api/v1/leave-requests/web/pending-approvals`
- `GET /api/v1/leave-requests/web/pending-approvals/count`

Both declare no parameters. #683 took the `approverId` query parameter off them deliberately: an approval queue is "what is waiting on me", and reading the approver from a parameter meant an administrator could read any colleague's queue while the line managers a chain is actually built from could not read their own. The attendance pair should be built that way from the start.

Suggested:

- `GET /api/v1/attendance/pending-approvals` and its `/web` twin, no parameters, the caller resolved from the session.
- `GET /api/v1/attendance/pending-approvals/count` and its `/web` twin, for a sidebar badge.

## What the query has to honour

Both of the rules that already exist on the write path, so that the queue and the approve call agree:

1. **The whole fallback chain, not the direct manager.** A caller sees a record when `geofenceApproverId` names them, and a caller holding the record-management roles sees the pending records they manage. `resolveApprover` leaves `geofenceApproverId` null when neither the reporting manager nor a project manager on the site resolves, and those records fall to the record managers, so a queue keyed only on the id would drop them. Worth stating in the issue because the reporting-manager field is set on 4 of 30 employees on staging, so the null and project-manager branches are the common cases today rather than the edge.

2. **Never the caller's own geofence record.** `requireActorMayApprove` refuses a self-approval on a record flagged for a geofence decision whatever roles the caller holds, and puts that refusal ahead of the role check. A record manager who marked away from site must not find their own day in their queue as something they can act on, or the queue is offering an action the API answers with a 403.

A reasonable read of both together is that the queue is exactly the set `canDecideApproval` would accept, restricted to `approvalStatus = PENDING`, minus the caller's own geofence records. Keeping it derived from that rather than restated would stop the two drifting.

## Why it is worth doing rather than leaving

Routing a decision to a named person and giving that person no way to find it means the decision does not get made. The web side of this is half-fixed already: tornotron/echno-web#402 now offers Approve and Reject to the approver the record names, where before it read a job title and gave a supervisor no buttons at all. Finding the record is the half that needs this endpoint.

Downstream: `echno-core` needs the service method and hooks, and `echno-web` needs the queue surface, which will sit next to the existing regularizations one at `/users/dashboard/attendance/regularizations` rather than becoming a second pattern.

abhijithanandan:
Closes #692.

## The gap

Since #681 an attendance day marked from outside a project's geofence is routed to a named person: the employee gives a reason, `AttendanceGeofenceService.resolveApprover` names who decides, and `AttendanceService.approveAttendance` accepts that person's decision. The decision has worked from the first day. Finding the record has not.

The two listings are `attendance/project/{projectId}`, which requires a `date` and has no approval filter, and `attendance/employee/{employeeId}`, which is one person over a range. So an approver reaches a held day only by already knowing it is there. tornotron/echno-web#402 gave the named approver the Approve and Reject buttons; this is the half that lets them find the record to press them on.

## What is added

Four endpoints, none of which takes a parameter:

| Endpoint | Answers |
|---|---|
| `GET /api/v1/attendance/pending-approvals` | The held days waiting on the caller, newest first, capped at 500 with `X-Total-Count` and `X-Result-Capped` |
| `GET /api/v1/attendance/pending-approvals/count` | `{"count": n}` for the sidebar badge |
| `GET /api/v1/attendance/web/pending-approvals` | The web twin |
| `GET /api/v1/attendance/web/pending-approvals/count` | The web twin |

The twins are the convention on this controller and every endpoint on it has one, so both were written rather than one; an earlier fix in this area shipped a twin short and left half a feature dead.

The caller comes from the session. Reading an approver off the query string under a guard that only asks for a role is the defect closed in #589, #599, #607, #631, #635 and #683, and again in the sweep findings #687 to #691: the guard checks a role, the query reads a number the caller sent, and the two never meet, so an administrator reads a colleague's queue while the line managers a chain is actually built from read none of their own. A caller that still sends `approverId` is served their own queue, a query parameter no handler declares being ignored.

## How the query agrees with the write path

`AttendanceService.requireActorMayApprove` runs two rules in order, and `AttendanceApprovalQueueSpecifications.waitingOn` is built to be the same two:

1. **Self-approval is refused ahead of the role check.** A record manager who marked from off site is refused their own day whatever roles they hold. The queue excludes `employeeId = caller` unconditionally, so it never offers an action the API answers with a 403.
2. **`canDecideApproval` is the record-management roles plus the approver the record names.** A caller holding the roles sees every other held day in the tenant, which is what reaches the records `resolveApprover` could name nobody for. Anyone else sees the days that name them. The id is read from the stored `geofenceApproverId`, so the chain is not walked again and the queue cannot drift from the record.

A null `geofenceApproverId` widens rather than strands: those records are the record managers' by the first branch. That is the common case rather than the edge, the reporting-manager field being set on a small minority of employees today.

## The one place the queue is narrower than `canDecideApproval`

The issue suggested the queue be "the set `canDecideApproval` would accept, restricted to `approvalStatus = PENDING`". Taken literally that is the entire attendance table: `checkIn` builds every record with `approvalStatus = PENDING` and nothing moves it until somebody decides the day, so pending is an ordinary record's resting state rather than a request for attention. An HR admin's queue would be every attendance row the organization has.

So the queue also filters `requiresGeofenceApproval = true`, which is what that flag was added for and what the entity documents it as ("what an approver's queue filters on"). The queue is therefore a strict subset of what the approve endpoint accepts, which is the safe direction: it never offers a record the server would refuse.

**Nothing else disagrees.** The web gate `features/attendance/lib/approval-gate.ts` mirrors the same two rules in the same order, including the fail-closed treatment of an unresolved viewer id, and this query agrees with it on every held record. The gate additionally allows a record manager to decide an ordinary, unflagged pending record, which the backend also allows and this queue simply does not list, for the reason above.

## Also

- `AttendanceRepository` gains `JpaSpecificationExecutor`, so the listing and the count read one predicate. A separate JPQL pair would have been two statements to keep in step, and the badge disagreeing with the list it labels is a bug users report as "the number is wrong".
- The count is `count(spec)`, not a page total. Spring's `PageableExecutionUtils` skips the count query when the first page comes back short, so a badge derived from a listing is right only while the queue is shorter than a page.
- Liquibase `090` indexes `(organization_id, requires_geofence_approval, approval_status, geofence_approver_id)`. Neither geofence column was ever a search term before; both are now read on every visit to the approvals screen, over the table that grows with every employee on every day on every project. A prefix of the index answers the record manager's three-column question, the whole of it the named approver's four.

## Tests

Two classes, and the split is the same one `RegularizationRegisterQueryIT` settled: who the caller is belongs to the service, what the SQL does belongs against a real database.

`AttendanceApprovalQueueTest` (Mockito, no context) captures the specification the service hands the repository and runs it against a recording criteria builder, so the predicate set is asserted rather than described.

| Test | Fails without the change because |
|---|---|
| `theQueueIsReadForTheSignedInCallerAndTakesNoApproverArgument` | the queue is narrowed to the caller's own employee id, resolved from the session |
| `theCountIsCountedForTheSameCallerTheListIsServedFor` | the badge and the list resolve the caller identically |
| `theCountDoesNotComeFromTheListing` | the count runs a count query and never touches `findAll` |
| `aCallerWithNoEmployeeRecordHereIsRefusedAQueueRatherThanServedSomebodyElses` | a caller with no employee record here is refused, not served |
| `aCallerWithNoEmployeeRecordHereIsRefusedTheCount` | same, on the count |
| `theQueueIsTheHeldDaysThatArePendingAndNeverTheCallersOwn` | the three predicates are `PENDING`, held, and not the caller's own |
| `aRecordManagerSeesTheHeldDaysThatNameNobody` | the approver predicate is dropped for a record manager, and the self-exclusion is not |
| `theQueueIsOrderedAndCapped` | an unordered page walk can show one record twice and never reach another |

`AttendanceApprovalQueueIT` (`@DataJpaTest` against CockroachDB) seeds ten rows across two tenants, each excluded row excluded for exactly one reason, so a dropped filter fails a different test.

| Test | Fails without the change because |
|---|---|
| `theNamedApproverSeesTheHeldDaysThatNameThemAndNothingElse` | the decided, the unheld, the other approver's and the other tenant's rows all stay out |
| `aRecordManagerSeesEveryHeldDayIncludingTheOnesNamingNobody` | a queue keyed only on the id drops the null-approver rows |
| `nobodySeesTheirOwnHeldDay` | the self-exclusion applies to the record manager, not only to the ordinary employee |
| `theCountIsTheWholeQueueAndNotTheFirstPage` | a page of two is asked for against a queue of four and the count is read separately; a page total would say two, an unscoped count seven |
| `theQueueIsScopedToTheTenant` | the same approver id is seeded in both tenants |
| `theQueueIsOrderedNewestDayFirst` | the order is total, with the primary key underneath the date |

**Verification path: CI.** This laptop has no container runtime, so the integration test cannot run here, and it is under load, so nothing was built on it. Every red above is reasoned from the code rather than measured against `development`, which is the honest description for a feature whose tests cannot compile against a branch where the endpoints do not exist. No red is claimed as measured.

`docs/openapi.json` is regenerated from this branch's own `openapi-served` build artifact rather than locally, and CI re-verifies it.

Downstream: `@tornotron/echno-core` gains the two methods and their hooks in a paired PR, since the contract there is hand-maintained and nothing would fail to compile without it.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -25237,6 +25237,217 @@
         ]
       }
     },
+    "/api/v1/attendance/pending-approvals": {
+      "get": {
+        "description": "Returns the attendance days held for a geofence decision that the signed-in caller may decide: the days that name them as approver, and, for a holder of the attendance record-management roles, every other held day in the tenant, which is how the days no approver could be resolved for are reached. A caller's own days are never in it, because nobody approves their own absence from site whatever roles they hold. Newest day first, up to a fixed ceiling; the response carries X-Total-Count with the true number waiting, and X-Result-Capped when there were more than one response can hold.",
+        "operationId": "getPendingApprovals_3",
+        "responses": {
+          "200": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "items": {
+                    "$ref": "#/components/schemas/AttendanceResponseDto"
+                  },
+                  "type": "array"
+                }
+              }
+            },
+            "description": "Records waiting on the caller returned"
+          },
+          "400": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetail"
+                }
+              }
+            },
+            "description": "Bad Request"
+          },
+          "402": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/SubscriptionErrorResponse"
+                }
+              }
+            },
+            "description": "Payment Required"
+          },
+          "403": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetail"
+                }
+              }
+            },
+            "description": "Caller is not a member of the current tenant, or has no employee record in it, so there is no queue to serve"
+          },
+          "404": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetail"
+                }
+              }
+            },
+            "description": "Not Found"
+          },
+          "409": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetail"
+                }
+              }
+            },
+            "description": "Conflict"
+          },
+          "422": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetail"
+                }
+              }
+            },
+            "description": "Unprocessable Entity"
+          },
+          "500": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetail"
+                }
+              }
+            },
+            "description": "Internal Server Error"
+          },
+          "502": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetail"
+                }
+              }
+            },
+            "description": "Bad Gateway"
+          }
+        },
+        "summary": "List the attendance days waiting on you",
+        "tags": [
+          "Attendance"
+        ]
+      }
+    },
+    "/api/v1/attendance/pending-approvals/count": {
+      "get": {
+        "description": "Returns how many attendance days are awaiting a decision from the signed-in caller, for the badge a client draws on the menu. Counted over the same set the listing serves.",
+        "operationId": "getPendingApprovalCount_3",
+        "responses": {
+          "200": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "additionalProperties": {
+                    "format": "int64",
+                    "type": "integer"
+                  },
+                  "type": "object"
+                }
+              }
+            },
+            "description": "Count returned"
+          },
+          "400": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetail"
+                }
+              }
+            },
+            "description": "Bad Request"
+          },
+          "402": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/SubscriptionErrorResponse"
+                }
+              }
+            },
+            "description": "Payment Required"
+          },
+          "403": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetail"
+                }
+              }
+            },
+            "description": "Caller is not a member of the current tenant, or has no employee record in it, so there is no queue to count"
+          },
+          "404": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetail"
+                }
+              }
+            },
+            "description": "Not Found"
+          },
+          "409": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetail"
+                }
+              }
+            },
+            "description": "Conflict"
+          },
+          "422": {
+            "content": {
+              "*/*": {
+                "schema": {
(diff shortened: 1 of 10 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #702, opened 2026-09-06
Hold the attendance and leave-balance reads to the policy they were meant to have

### LINKED ISSUE #688: Attendance and movement records are readable by id by any member of the tenant, which is the check canViewEmployeeRecords was written to make
Found while sweeping for further instances of the family closed in #589, #599, #607, #631, #635, #683 and #684. Not the caller-supplied-id shape exactly: here the id names a record rather than a person, and the fault is that nothing connects the record to the caller at all. The effect is the same, and the policy being bypassed already exists a few lines away.

## The endpoints

| Endpoint | Guard | Service |
| --- | --- | --- |
| `GET /api/v1/attendance/{id}` (`AttendanceController:99`) | `@orgSecurity.isMemberOfCurrentTenant()` | `getAttendanceById(id)` → `findByIdAndOrganization_Id(id, TenantContext.getCurrentOrgId())` |
| `GET /api/v1/attendance/web/{id}` (`AttendanceControllerWeb:97`) | same | same |
| `GET /api/v1/movement-records/{id}` (`MovementRecordController:60`) | same | `getMovementById(id)` → `findByIdAndOrganization_Id(...)` |
| `GET /api/v1/movement-records/web/{id}` (`MovementRecordControllerWeb:61`) | same | same |
| `GET /api/v1/movement-records/attendance/{attendanceId}` (`MovementRecordController:75`) | same | `findByAttendanceIdOrderByStartTimeAsc(attendanceId)` |
| `GET /api/v1/movement-records/web/attendance/{attendanceId}` (`MovementRecordControllerWeb:76`) | same | same |

Every one is tenant-scoped and none is employee-scoped. Any member of the organization walks integer ids and reads every colleague's record.

## Why this matters more than an ordinary over-broad read

`AttendanceSecurityService.canViewEmployeeRecords(#employeeId)` exists, is documented, and gates the sibling route `GET /attendance/employee/{employeeId}` on the same controller (line 113) to self-or-manager. Enumerating `/{id}` returns the identical `AttendanceResponseDto`: `employeeId`, `employeeName`, check-in and check-out coordinates, geofence status and the attachment, which for this product is a selfie and a GPS fix. So the policy is written, applied on one route, and bypassable on the neighbouring one.

The write paths in the same package are careful in exactly the way the reads are not. `addMovement` calls `requireActorMayRecordFor(employeeId)`, the approval path re-reads the approver off the stored record with `canDecideApproval(attendance.getGeofenceApproverId())`, and `AttendanceSecurityService`'s own javadoc says the `@PreAuthorize` guard "only sees the caller's word". The reads never got the same treatment.

## The one repository query with no organization predicate

`MovementRecordRepository:33`

```java
List<MovementRecord> findByAttendanceIdOrderByStartTimeAsc(Long attendanceId);
```

Its sibling `getMovementById` uses `findByIdAndOrganization_Id`. `MovementRecord` is a `TenantScopedEntity` carrying `@Filter(name = "orgFilter", ...)`, so a derived query on it is scoped by the session filter and this does not in fact cross tenants; I checked the entity rather than assuming from the method name. It is still the only read in the class relying on the filter alone rather than saying what it means, which is the condition the 2026-08-18 audit flagged as a latent trap, so it is worth an explicit predicate as defence in depth.

## Fix direction

Resolve the record first, then check the employee on it, the way the approval path already resolves the approver off the stored row. `canViewEmployeeRecords(attendance.getEmployee().getId())` is the existing answer and should be reachable from the by-id reads on both twins. The movement reads should reach it through the attendance record they hang off, so that a movement trail is readable by exactly the people who may read the attendance it belongs to, rather than by a separate rule that will drift.

Check both twins: each of these three routes exists on the mobile and the `/web` controller with the id moving between a path segment and a query parameter, and fixing one twin is how #684 nearly got missed.

### LINKED ISSUE #689: An employee cannot see how many leave days they have left, and every HR administrator can see everyone's
Found while sweeping for further instances of the family closed in #589, #599, #607, #631, #635, #683 and #684. Seventh instance, and it carries both halves in one controller pair.

## The endpoints

All on `LeaveBalanceController` and `LeaveBalanceControllerWeb`, the id moving from a path segment to a query parameter between the twins.

| Endpoint | Guard | Service |
| --- | --- | --- |
| `GET /leave-balances/employee/{employeeId}` (`:50`) | `hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin')` | `getAllBalancesForEmployee(employeeId, ...)` |
| `GET /leave-balances/employee/{employeeId}/policy/{policyId}` (`:70`) | same | `getOrCalculateBalance(employeeId, ...)` |
| `POST /leave-balances/employee/{employeeId}/recalculate` (`:111`) | same | `recalculateBalances(employeeId, ...)` |
| `GET /leave-balances/employee/{employeeId}/summary` (`:92`) | `@orgSecurity.isMemberOfCurrentTenant()` | `getBalanceSummary(employeeId, ...)` |
| `GET /leave-balances/web?employeeId=` (`Web:49`) | the role pair | `getAllBalancesForEmployee` |
| `GET /leave-balances/web/specific?employeeId=` (`Web:68`) | the role pair | `getOrCalculateBalance` |
| `POST /leave-balances/web/recalculate?employeeId=` (`Web:107`) | the role pair | `recalculateBalances` |
| `GET /leave-balances/web/summary?employeeId=` (`Web:89`) | `isMemberOfCurrentTenant()` | `getBalanceSummary` |

None of the guards reads `#employeeId`. The services resolve the employee under `TenantContext.getCurrentOrgId()`, so this stays inside the tenant, and answers about whoever the caller named.

## Both halves

**Not far enough, and this is the one a user would report.** An ordinary employee cannot read their own leave balance. Entitlement, consumed and remaining are behind the system-admin and hr-admin roles, and the only balance-adjacent route an employee can reach for themselves is `/transactions`, guarded on `isSelfInCurrentTenant`, which returns the ledger and not the balances. "How many days do I have left" is the first question anybody asks a leave module, and the product cannot answer it for the person asking.

**Too far.** Every `hr-admin` and `system-admin` reads every colleague's balances by walking ids, and on `recalculate` triggers a write against them. Separately, the `summary` pair is gated on bare tenant membership, so for that one view **any** employee reads any colleague's entitlement and remaining days by passing their id.

## The correct expression is already in the file

`getTransactionHistory` on the same controller is `@orgSecurity.isSelfInCurrentTenant(#employeeId)`, and `LeaveRequestController:48` uses `@orgSecurity.isSelfOrHasAnyOrgRole(#employeeId, 'system-admin', 'hr-admin')`. The latter is what these want: the employee reads their own, the administrators read anybody's.

`recalculate` is a write and may deserve to stay with the administrators; that is a judgement to make deliberately rather than by copying the read.

## Also worth deciding, not folded in

`POST /leave-balances/adjust` on both twins is role-gated with the `employeeId` inside `LeaveBalanceAdjustmentDto`. A manual HR adjustment is an administrative function by definition, so that one looks correct as it stands. Recording it here so a fix does not sweep it up by pattern-matching.

## Trace the whole workflow

#666 was the case where the endpoint named in the issue was repaired and the caller was still unable to complete the workflow, because a read one tier up was left refusing them. Before closing this, check the balance the leave request form displays when an employee fills it in, and the balance the approval screen shows an approver deciding on a request, and confirm both callers can reach what those screens read.

abhijithanandan:
Closes #688. Closes #689.

## What was verified, and where it is smaller than filed

Both issues were checked against the code rather than taken on trust, because the sweep that filed them had already corrected its own cross-tenant claims once.

**#688 stands as written, and it is the more serious of the two.** `Attendance` and `MovementRecord` both implement `TenantScopedEntity` and carry `@Filter(name = "orgFilter", ...)`, so nothing here crosses a tenant. That was never the claim. The claim is that within one organization the by-id reads were guarded by membership alone while `@attendanceSecurity.canViewEmployeeRecords` gated the employee-scoped route on the same controller, and both return the same `AttendanceResponseDto`. So the policy was written, documented, applied on one route, and bypassable on the one next to it by counting.

**#689's two halves are not equally severe, and the issue text overstates one of them.**

- The read that is too narrow is real: an employee could not read their own entitlement, consumed or remaining days, and the only balance-adjacent route they could reach for themselves returned the ledger rather than the balances. `features/leave/components/dashboard/employee-dashboard.tsx` and the leave-request form both call these.
- "Every HR administrator can see everyone's" is **not** a defect. An hr-admin reading any employee's leave balance inside their own organization is the job. That half of the title should be read as describing the summary route only, which was on bare tenant membership, so **any** employee could read a colleague's entitlement by passing their id. That one is real and is fixed.
- One correction to the issue's endpoint table: `POST /recalculate` does not call `recalculateBalances`. Both twins call `getAllBalancesForEmployee`, the same method the GET beside them calls, so the route is a read wearing a POST and there is no bulk recalculate in the service at all. It is left role-gated and filed separately rather than folded in here.

**The `@Transactional` note that matters for review:** these reads are not pure. `getOrCalculateBalance` persists a balance row through `initializeBalance` and recomputes accrual when a month has passed. Widening them to the employee therefore lets an employee cause their own row to be initialised, which is how the figures come to exist at all. No other employee's row is reachable.

## The fix

**Attendance and movements (#688).** The check is taken in the service against the loaded record, not in an annotation, because the id on these requests names a record and so identifies no person. That is exactly where the writes in the same package already take theirs. New `AttendanceSecurityService.canViewAttendanceRecord(employeeId, designatedApproverId)`: `canViewEmployeeRecords` plus the approver the record names.

The approver branch is load-bearing rather than decorative. A geofence exception is decided by the employee's reporting manager, who usually holds none of the organization-wide roles, and `canDecideApproval` already lets them decide it. Without the branch this change would have produced a caller who may approve a record and may not look at it, which is #666 in a new place. There is a test for it.

The movement reads reach the policy through the attendance they hang off, so a movement trail is readable by exactly the people who may read the day it belongs to, rather than by a second rule that would drift from the first. `findByAttendanceIdOrderByStartTimeAsc` also gains an explicit organization predicate: the filter did cover it, so this is defence in depth and not a repair, but it was the one read in the class that said nothing about the tenant it meant.

**Leave balances (#689).** The balance listing, the single-policy balance, the summary and the transaction ledger all land on `@orgSecurity.isSelfOrHasAnyOrgRole(#employeeId, 'system-admin', 'hr-admin')`, which is what `LeaveRequestController` already uses for the request these balances are spent by. That is a widening for the first two, a narrowing for the summary, and a widening for the ledger.

The ledger moved because of the workflow trace the issue asked for. `app/users/dashboard/workforce/leaves/manage/balance/page.tsx` reads the summary and the transaction history side by side, and adjusts balances from `balance-adjustment-dialog.tsx`. The ledger was self-only, so an administrator could adjust a balance and not see the transaction their adjustment produced. It grants nothing new in kind: `/{balanceId}/transactions` already hands those roles the same rows.

`recalculate` and `adjust` stay with the administrators. Both are written as commands, and an employee triggering a recalculation of their own figures is a different decision from an employee being told what they are.

## Both twins

Checked on all six routes. The mobile and `/web` controllers were byte-identical in guard on every one, and both are changed.

## Tests

The first commit is the tests alone. It was run on CI against the unmodified code (run [34050549515](https://github.com/tornotron/echno-backend/actions/runs/34050549515)); the `Compile` step passed and `Test` failed on exactly the eight cases below, so every red here is **measured** and none is reasoned. This workstation has no container runtime and is under load, so CI was the verification path throughout.

|---|---|---|
| `AttendanceRecordReadAuthorizationTest.attendanceById_isRefused_toAMemberWhoIsNeitherTheEmployeeNorAManager` | an ordinary member cannot read a colleague's day by id | **measured** |
| `AttendanceRecordReadAuthorizationTest.movementById_isRefused_toAMemberWhoMayNotReadTheAttendanceItHangsOff` | same for one movement | **measured** |
| `AttendanceRecordReadAuthorizationTest.movementsByAttendance_areRefused_toAMemberWhoMayNotReadTheAttendance` | same for the trail | **measured** |
| `AttendanceRecordReadAuthorizationTest.attendanceById_isAllowed_toTheEmployeeItBelongsTo` | the employee still reads their own | passed before and after, by design |
| `AttendanceRecordReadAuthorizationTest.attendanceById_isAllowed_toTheApproverTheRecordNames` | the designated approver is not shut out | passed before and after, by design |
| `AttendanceRecordReadAuthorizationTest.movementsByAttendance_areAllowed_toTheEmployeeTheAttendanceBelongsTo` | same for the trail | passed before and after, by design |
| `LeaveBalanceReadGuardTest.anEmployeeCanReadTheirOwnBalances` | the narrow half, mobile | **measured** |
| `LeaveBalanceReadGuardTest.anEmployeeCanReadTheirOwnBalancesOnTheWebTwin` | the narrow half, web | **measured** |
| `LeaveBalanceReadGuardTest.aColleaguesSummaryIsRefusedToAnOrdinaryEmployee` | the wide half, mobile | **measured** |
| `LeaveBalanceReadGuardTest.aColleaguesSummaryIsRefusedOnTheWebTwinToo` | the wide half, web | **measured** |
| `LeaveBalanceReadGuardTest.anAdministratorCanReadTheLedgerBehindABalanceTheyAdjusted` | the workflow half | **measured** |
| `LeaveBalanceReadGuardTest.anEmployeeCanReadTheirOwnSummary` | the summary an employee could already reach still answers | passed before and after, by design |
| `LeaveBalanceReadGuardTest.anAdministratorCanStillReadAnyEmployeesBalances` | the administrators lose nothing | passed before and after, by design |
| `LeaveBalanceReadGuardTest.recalculateStaysWithTheAdministrators` | the command route did not move | passed before and after, by design |

The attendance cases drive the real `AttendanceSecurityService` over a mocked `OrganizationSecurityService`, so they exercise the policy composition rather than a stub of it.

## OpenAPI

Description text only, twenty lines, no structure. Taken from the build's own `openapi-served` artifact rather than regenerated here, since this workstation has no container runtime; the artifact is written from the identical canonical string the update flag produces, and CI re-verifies it.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -26899,7 +26899,7 @@
         ]
       },
       "get": {
-        "description": "Returns a single attendance record including its clock events, movements, regularizations and approval state.",
+        "description": "Returns a single attendance record including its clock events, movements, regularizations and approval state. Readable by the employee it belongs to, by a holder of an attendance record-management role, or by the approver the record names.",
         "operationId": "getById_5",
         "parameters": [
           {
@@ -26951,7 +26951,7 @@
                 }
               }
             },
-            "description": "Caller is not a member of the current tenant"
+            "description": "Caller is neither the employee the record belongs to, nor a holder of an attendance record-management role, nor the approver the record names"
           },
           "404": {
             "content": {
@@ -27246,7 +27246,7 @@
         ]
       },
       "get": {
-        "description": "Returns a single attendance record including its clock events, movements, regularizations and approval state.",
+        "description": "Returns a single attendance record including its clock events, movements, regularizations and approval state. Readable by the employee it belongs to, by a holder of an attendance record-management role, or by the approver the record names.",
         "operationId": "getById_4",
         "parameters": [
           {
@@ -27298,7 +27298,7 @@
                 }
               }
             },
-            "description": "Caller is not a member of the current tenant"
+            "description": "Caller is neither the employee the record belongs to, nor a holder of an attendance record-management role, nor the approver the record names"
           },
           "404": {
             "content": {
@@ -62979,7 +62979,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller is neither the employee identified by the id nor a holder of the system-admin or hr-admin role in the current tenant"
           },
           "404": {
             "content": {
@@ -63110,7 +63110,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller is neither the employee identified by the id nor a holder of the system-admin or hr-admin role in the current tenant"
           },
           "404": {
             "content": {
@@ -63357,7 +63357,7 @@
                 }
               }
             },
-            "description": "Caller is not a member of the current tenant"
+            "description": "Caller is neither the employee identified by the id nor a holder of the system-admin or hr-admin role in the current tenant"
           },
           "404": {
             "content": {
@@ -63473,7 +63473,7 @@
                 }
               }
             },
-            "description": "Caller is not the employee identified by the id"
+            "description": "Caller is neither the employee identified by the id nor a holder of the system-admin or hr-admin role in the current tenant"
           },
           "404": {
             "content": {
@@ -63598,7 +63598,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller is neither the employee identified by the id nor a holder of the system-admin or hr-admin role in the current tenant"
           },
           "404": {
             "content": {
@@ -63966,7 +63966,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller is neither the employee identified by the id nor a holder of the system-admin or hr-admin role in the current tenant"
           },
           "404": {
             "content": {
@@ -64088,7 +64088,7 @@
                 }
               }
             },
-            "description": "Caller is not a member of the current tenant"
+            "description": "Caller is neither the employee identified by the id nor a holder of the system-admin or hr-admin role in the current tenant"
           },
           "404": {
             "content": {
@@ -64204,7 +64204,7 @@
                 }
               }
             },
-            "description": "Caller is not the employee identified by the id"
+            "description": "Caller is neither the employee identified by the id nor a holder of the system-admin or hr-admin role in the current tenant"
           },
           "404": {
             "content": {
@@ -76170,7 +76170,7 @@
     },
     "/api/v1/movement-records/attendance/{attendanceId}": {
       "get": {
-        "description": "Returns every movement logged against a given attendance record.",
+        "description": "Returns every movement logged against a given attendance record. Readable by whoever may read that attendance record.",
         "operationId": "getByAttendance_1",
         "parameters": [
           {
@@ -76225,7 +76225,7 @@
                 }
               }
             },
-            "description": "Caller is not a member of the current tenant"
+            "description": "Caller may not read the attendance record these movements belong to"
           },
           "404": {
             "content": {
@@ -76409,7 +76409,7 @@
     },
     "/api/v1/movement-records/web/attendance/{attendanceId}": {
       "get": {
-        "description": "Returns every movement logged against a given attendance record.",
+        "description": "Returns every movement logged against a given attendance record. Readable by whoever may read that attendance record.",
         "operationId": "getByAttendance",
         "parameters": [
           {
@@ -76464,7 +76464,7 @@
                 }
               }
             },
-            "description": "Caller is not a member of the current tenant"
+            "description": "Caller may not read the attendance record these movements belong to"
           },
           "404": {
             "content": {
@@ -76525,7 +76525,7 @@
     },
     "/api/v1/movement-records/web/{id}": {
       "get": {
-        "description": "Returns a single movement record including its verification state.",
+        "description": "Returns a single movement record including its verification state. Readable by whoever may read the attendance record it hangs off.",
         "operationId": "getById_3",
         "parameters": [
           {
@@ -76577,7 +76577,7 @@
                 }
               }
             },
-            "description": "Caller is not a member of the current tenant"
+            "description": "Caller may not read the attendance record this movement belongs to"
           },
           "404": {
             "content": {
@@ -76751,7 +76751,7 @@
     },
     "/api/v1/movement-records/{id}": {
       "get": {
-        "description": "Returns a single movement record including its verification state.",
+        "description": "Returns a single movement record including its verification state. Readable by whoever may read the attendance record it hangs off.",
         "operationId": "getById_2",
         "parameters": [
           {
@@ -76803,7 +76803,7 @@
                 }
               }
             },
-            "description": "Caller is not a member of the current tenant"
+            "description": "Caller may not read the attendance record this movement belongs to"
           },
           "404": {
             "content": {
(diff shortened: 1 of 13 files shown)


abhijithanandan:
@coderabbitai review

PR #703, opened 2026-09-06
Make the directory and leave-policy reads check the id they are given

### LINKED ISSUE #690: Two employee-directory reads take the organization or the manager from the caller, and the line manager they are for holds no role that reaches them
Found while sweeping for further instances of the family closed in #589, #599, #607, #631, #635, #683 and #684. Two endpoints on `EmployeeControllerWeb`, same shape, different consequences.

## `GET /api/v1/employee/web/managers/organizationId/{organizationId}` (`:329`)

```java
@PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin','project-manager')")
```

The guard checks a role in the caller's own tenant and never reads `#organizationId`. `EmployeeHierarchyService.readAllTheManagersByOrganizationId` (`:141`) calls `employeeRepository.findEmployeesByOrganizationIdAndOrgRoles(organizationId, OrgRole.getManagerRoles())`, with no tenant predicate of its own.

`Employee` is a `TenantScopedEntity` carrying `@Filter(name = "orgFilter", ...)`, so a foreign id returns empty rather than another tenant's managers; I checked the entity rather than reading the missing predicate as a cross-tenant hole. What is left is a parameter that does nothing except let the query answer about something the caller chose, with the tenant filter doing the work the check should be doing.

The adjacent `GET /managers` (`:315`) takes no id, which suggests the id-bearing one is simply redundant and can go.

Second point on this one: the response is a full `EmployeeDto`, which carries `salary`, `dateOfBirth`, `address`, `bloodGroup` and `phoneNumber`. A `project-manager` reading the manager directory gets all of it. Whether a project manager should see colleagues' salaries is a product question, not a bug, but it is worth answering while this endpoint is open.

## `GET /api/v1/employee/web/managerId/{managerId}/subordinates` (`:300`)

```java
@PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin','project-manager')")
```

`EmployeeHierarchyService.getDirectSubordinates` (`:106`) tenant-checks the manager with `existsByIdAndOrganization_Id(managerId, TenantContext.getCurrentOrgId())` and then returns `findByManager_Id(managerId)`.

**Not far enough, and this is the half that matters.** Reporting lines are set by `assignManager` on `Employee.manager`, which is independent of the Keycloak-derived `OrgRole` set. A site supervisor who is somebody's manager in the data but holds no `project-manager` org role gets a 403 on their own direct reports. This is the same manager-versus-role mismatch that `AttendanceSecurityService.canDecideApproval` was written to solve, and that #685 found in the leave approval chain, where the approver is whoever `employee.getManager()` walks to and holds neither administrative role.

**Too far.** Any `project-manager` passes any other manager's id and receives their whole team as full `EmployeeDto`s.

## Fix direction

Both want a guard that reads the id it is given. For subordinates that is self-or-roles on `#managerId`; `@orgSecurity.isSelfOrHasAnyOrgRole` already exists and is used at `EmployeeControllerWeb:195`. For the manager directory the tenant is settled by the session, so the id should come from `TenantContext` and the path segment go, unless the endpoint is simply redundant against `/managers`, in which case removing it is cleaner.

## Weaker, recorded here rather than filed separately

- `GET /api/v1/employee/web/{id}` (`:159`), role-gated, so an employee cannot read their own record even though the `PATCH` at `:195` is correctly `isSelfOrHasAnyOrgRole(#id, ...)`. A fix that landed on the write and missed the read. `UserControllerWeb:130` gives the employee an alternative self-route, which is the only reason this is not filed on its own. The mobile twin `EmployeeController:126` has the same shape under a phantom `hasAuthority('employee:read')`, so it refuses everybody, which is #684's mechanism in another module.
- `GET /api/v1/employee/web/{employeeId}/roles` (`:375`), `isMemberOfCurrentTenant()`, lets any member enumerate who holds `system-admin` and `hr-admin`. Reconnaissance rather than personal data.
- `ProjectControllerWeb:373`, `isMemberOfCurrentTenant() or hasAnyOrgRole(...)`, where the `or` makes the role clause dead and any member lists any colleague's project assignments.

### LINKED ISSUE #691: The leave-policy listing takes the organization from the path and checks only that it exists

abhijithanandan:
Closes #690. Closes #691.

Uses the `@orgSecurity.isCurrentTenant` guard #696 introduced for #687. That has since merged, so this now sits directly on `development`.

## What was verified, and where it is smaller than filed

**Both organization-id findings are the redundant-parameter form of this family, not the dangerous one.** Checked on the entities rather than inferred from the missing predicate: `Employee` and `LeavePolicy` both implement `TenantScopedEntity` and carry `@Filter(name = "orgFilter", ...)`, and `HibernateFilterConfig` enables that filter on every `@Transactional` method with a tenant in force. A foreign organization id therefore returns an empty list rather than another tenant's managers or another tenant's accrual rates and encashment caps. Neither is #687, where `Organization` itself is loaded and the tenant root sits outside both the filter and the load listener.

What is left in both is real but modest: a parameter that decides what is answered with nothing establishing the caller is entitled to it, and in the leave-policy case an `existsById` that reads as though it were the tenant check and is not.

**#691 is not a duplicate of #687.** Different site, different repository call: #687 was `organizationRepository.findById` producing a credential bound to whatever was named; this is `existsById` in front of a filtered query. It is the general hazard #687's body flagged for separate handling, and this is that handling for one instance of it.

**#690's subordinate half is the one that matters, and it is as filed.** Reporting lines are set by `assignManager` on `Employee.manager`, which is independent of the Keycloak-derived `OrgRole` set, so the guard and the data were answering different questions in both directions at once: a line manager holding no `project-manager` role got a 403 on their own direct reports, and any holder of that role could name another manager and receive their whole team as full `EmployeeDto`s, carrying `salary`, `dateOfBirth`, `address` and `phoneNumber`.

## The fix

`GET /employee/web/managerId/{managerId}/subordinates` moves to `@orgSecurity.isSelfOrHasAnyOrgRole(#managerId, 'system-admin', 'hr-admin')`. Both halves close at once: the manager reads their own team through the self branch, and the administrative roles read anybody's. `project-manager` stops being a blanket key to every manager's team and keeps its own team through the same self branch. Neither `echno-core` nor `echno-web` calls this route (`useSubordinates` is defined in core and consumed nowhere), so the narrowing costs no screen.

`GET /employee/web/managers/organizationId/{organizationId}` and `GET /leave-policies/organization/{organizationId}` keep their path segment and gain `@orgSecurity.isCurrentTenant(#organizationId)` alongside the role check, so the segment becomes binding rather than decorative. One answers where, the other what. Deleting the manager route as redundant against `GET /managers` was the alternative the issue offered and is defensible: the path string appears nowhere in `echno-core` or `echno-web`, only a stale `employeeKeys.managersByOrg` query-key factory with no service method behind it. It is left in place because removing a route from a published contract is a separate decision from repairing its guard, and this PR should not make both.

`LeavePolicyService.getPoliciesByOrganization` gains a note saying why its `existsById` is not the tenant check it reads like, so the next reader does not take it for one.

## The read that was traced rather than filed

`GET /leave-policies/employee/{employeeId}` and its web twin were `isSelfInCurrentTenant`, while `LeaveRequestController:48` lets a system or HR admin raise a leave request **on an employee's behalf** with `isSelfOrHasAnyOrgRole`. The request form at `app/users/dashboard/workforce/leaves/manage/requests/new/page.tsx` calls exactly this read to populate itself. So the caller entitled to finish the job could not read what the form needed: #666 again, in the file #691 names. The two guards are now the same expression.

This grants nothing new. The same roles already read every policy in the organization, with full accrual and encashment rules, through the role-gated `GET /leave-policies/web`; this returns a subset of that filtered by one employee's gender and service months.

## Twins

- Subordinates and the manager directory exist only on `EmployeeControllerWeb`. There is no mobile twin to keep in step.
- The leave-policy organization listing's web twin is **commented out** at `LeavePolicyControllerWeb` lines 84 to 95, so there is no live counterpart. The console reaches leave policies through `GET /leave-policies/web` instead, which takes no id and is correct as it stands. That confirms the issue's guess about which route survives.
- The applicable-policies read exists on both twins and both are changed.

## Tests

The first commit is the tests alone. It was run on CI against the unmodified code (run [34050748852](https://github.com/tornotron/echno-backend/actions/runs/34050748852)); `Compile` passed and `Test` failed on exactly the six cases below, so every red here is **measured** and none is reasoned. This workstation has no container runtime and is under load, so CI was the verification path.

|---|---|---|
| `EmployeeDirectoryReadGuardTest.aLineManagerCanReadTheirOwnDirectReports` | a manager holding no organization-wide role reaches their own team | **measured** |
| `EmployeeDirectoryReadGuardTest.anotherManagersTeamIsRefusedToARoleHolderWhoIsNotThatManager` | a `project-manager` no longer walks manager ids | **measured** |
| `EmployeeDirectoryReadGuardTest.theManagerDirectoryIsRefusedForAnOrganizationTheCallerIsNotEntitledTo` | the path segment binds | **measured** |
| `EmployeeDirectoryReadGuardTest.theManagerDirectoryStillAnswersForTheCallersOwnOrganization` | the ordinary case is unaffected | passed before and after, by design |
| `LeavePolicyReadGuardTest.theListingIsRefusedForAnOrganizationTheCallerIsNotEntitledTo` | same for leave policies | **measured** |
| `LeavePolicyReadGuardTest.theListingStillAnswersForTheCallersOwnOrganization` | the ordinary case is unaffected | passed before and after, by design |
| `LeavePolicyReadGuardTest.anAdministratorCanReadThePoliciesTheRequestFormWillOffer` | the workflow half, mobile | **measured** |
| `LeavePolicyReadGuardTest.anAdministratorCanReadThemOnTheWebTwinToo` | the workflow half, web | **measured** |
| `LeavePolicyReadGuardTest.anEmployeeStillReadsTheirOwnApplicablePolicies` | the employee loses nothing | passed before and after, by design |
| `LeavePolicyReadGuardTest.anEmployeeStillCannotReadAColleaguesApplicablePolicies` | the employee branch is still self and only self | passed before and after, by design |

## OpenAPI

`@PreAuthorize` expressions and response-description text only, no route or DTO change.

## Recorded in #690, deliberately not fixed here

Three weaker observations the issue lists rather than files. None is folded in, because each is a decision rather than a guard with an obvious right answer:

- `GET /employee/web/{id}` is role-gated while the `PATCH` beside it is correctly `isSelfOrHasAnyOrgRole(#id, ...)`, so an employee cannot read the record they may edit. `UserControllerWeb:130` gives them another self-route, which is the only thing keeping this benign. The mobile twin sits under a phantom `hasAuthority('employee:read')` and so refuses everybody, which is #684's mechanism in another module.
- `GET /employee/web/{employeeId}/roles` on bare membership lets any member enumerate who holds `system-admin` and `hr-admin`.
- `ProjectControllerWeb:373` reads `isMemberOfCurrentTenant() or hasAnyOrgRole(...)`, where the `or` makes the role clause dead and any member lists any colleague's project assignments.

Whether a `project-manager` should read colleagues' salaries through `EmployeeDto` is the same kind of question and is likewise left open.

## Review

CodeRabbit reviewed this and raised one nitpick, which was valid and is applied in the last commit: the suite proved the administrators could now reach the applicable-policies read and did not prove an ordinary employee still could not reach a colleague's, which would leave the change indistinguishable from dropping the guard to bare membership. `anEmployeeStillCannotReadAColleaguesApplicablePolicies` holds that side.

Its second suggestion, an administrator case naming an employee of another organization, is not applicable at this level. `isSelfOrHasAnyOrgRole` answers such a call on the role branch, and what refuses it afterwards is the service's own `findByIdAndOrganizationId` plus the `orgFilter` on `Employee`. Neither is exercised in a web slice where `@orgSecurity` is a mock, so a test there would assert the mock rather than the property. `TenantIsolationIT` is where that belongs.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -34659,7 +34659,7 @@
     },
     "/api/v1/employee/web/managerId/{managerId}/subordinates": {
       "get": {
-        "description": "Returns every employee who reports directly to the given manager.",
+        "description": "Returns every employee who reports directly to the given manager. Callable by that manager, or by a system or HR admin for anybody.",
         "operationId": "getDirectSubordinates",
         "parameters": [
           {
@@ -34714,7 +34714,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller is neither the manager named nor a system or HR admin"
           },
           "404": {
             "content": {
@@ -34880,7 +34880,7 @@
     },
     "/api/v1/employee/web/managers/organizationId/{organizationId}": {
       "get": {
-        "description": "Returns every employee holding a manager organization role within the given organization.",
+        "description": "Returns every employee holding a manager organization role within the given organization, which has to be the caller's own. Identical to the list beside it, which takes the organization from the session instead.",
         "operationId": "getAllManagersForAnOrganization",
         "parameters": [
           {
@@ -34935,7 +34935,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller is not entitled to the organization named, or lacks the required role in it"
           },
           "404": {
             "content": {
@@ -66291,7 +66291,7 @@
                 }
               }
             },
-            "description": "Caller is not the employee identified by the id"
+            "description": "Caller is neither the employee identified by the id nor a holder of the system-admin or hr-admin role in the current tenant"
           },
           "404": {
             "content": {
@@ -66416,7 +66416,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller is not entitled to the organization named, or lacks the required role in it"
           },
           "404": {
             "content": {
@@ -67095,7 +67095,7 @@
                 }
               }
             },
-            "description": "Caller is not the employee identified by the id"
+            "description": "Caller is neither the employee identified by the id nor a holder of the system-admin or hr-admin role in the current tenant"
           },
           "404": {
             "content": {
--- a/src/main/java/org/tornotron/echno_backend/employee/EmployeeControllerWeb.java
+++ b/src/main/java/org/tornotron/echno_backend/employee/EmployeeControllerWeb.java
@@ -300,14 +300,15 @@ public class EmployeeControllerWeb {
      * @return A {@link ResponseEntity} containing a list of employee DTOs who report to the manager and HTTP status 200 (OK).
      */
     @GetMapping("/managerId/{managerId}/subordinates")
-    @PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin','project-manager')")
+    @PreAuthorize("@orgSecurity.isSelfOrHasAnyOrgRole(#managerId, 'system-admin', 'hr-admin')")
     @Operation(
             summary = "List a manager's direct subordinates",
-            description = "Returns every employee who reports directly to the given manager."
+            description = "Returns every employee who reports directly to the given manager. Callable "
+                    + "by that manager, or by a system or HR admin for anybody."
     )
     @ApiResponses({
             @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "200", description = "Subordinates returned"),
-            @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "403", description = "Caller lacks the required role in the current tenant"),
+            @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "403", description = "Caller is neither the manager named nor a system or HR admin"),
             @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "404", description = "No manager with the given id")
     })
     public ResponseEntity<List<EmployeeDto>> getDirectSubordinates(@PathVariable Long managerId) {
@@ -329,14 +330,17 @@ public class EmployeeControllerWeb {
     }
 
     @GetMapping("/managers/organizationId/{organizationId}")
-    @PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin','project-manager')")
+    @PreAuthorize("@orgSecurity.isCurrentTenant(#organizationId) "
+            + "and @orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin','project-manager')")
     @Operation(
             summary = "List managers for an organization",
-            description = "Returns every employee holding a manager organization role within the given organization."
+            description = "Returns every employee holding a manager organization role within the given "
+                    + "organization, which has to be the caller's own. Identical to the list beside it, "
+                    + "which takes the organization from the session instead."
     )
     @ApiResponses({
             @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "200", description = "Managers returned"),
-            @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "403", description = "Caller lacks the required role in the current tenant"),
+            @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "403", description = "Caller is not entitled to the organization named, or lacks the required role in it"),
             @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = "404", description = "No organization with the given id")
     })
     public ResponseEntity<List<EmployeeDto>> getAllManagersForAnOrganization(@PathVariable Long organizationId) {
--- a/src/main/java/org/tornotron/echno_backend/leave/LeavePolicyController.java
+++ b/src/main/java/org/tornotron/echno_backend/leave/LeavePolicyController.java
@@ -88,7 +88,8 @@ public class LeavePolicyController {
     }
 
     @GetMapping("/organization/{organizationId}")
-    @PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin')")
+    @PreAuthorize("@orgSecurity.isCurrentTenant(#organizationId) "
+            + "and @orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin')")
     @Operation(
             summary = "List an organization's leave policies",
             description = "Returns the leave policies configured for the organization. Active policies "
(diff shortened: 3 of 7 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Applied. `anEmployeeStillCannotReadAColleaguesApplicablePolicies` holds the side the suite was missing: proving only that the administrators can now reach this read leaves the change indistinguishable from dropping the guard to bare tenant membership, which is exactly what the leave-balance summary route carried before #702.

The second half of the suggestion, an administrator case naming an employee of another organization, is not applicable here. `isSelfOrHasAnyOrgRole` answers such a call on the role branch, and what refuses it afterwards is the service's own `findByIdAndOrganizationId` together with the `orgFilter` on `Employee`. Neither runs in a web slice where `@orgSecurity` is a mock, so a test at this level would be asserting the mock. That property belongs in `TenantIsolationIT`, which asserts it against a real database.

PR #705, opened 2026-09-06
Bound how many invite codes may be offered for redemption

### LINKED ISSUE #699: An invite code is a five-digit bearer credential redeemable without an organization qualifier or a rate limit
Raised while fixing #687 in #696. That issue asked separately whether the invite code should come back in the 201 body at all. The answer there was yes, because the listing endpoint already returns code values to the same role and withholding it from the creation response removes nothing an entitled caller cannot read a second later. Judging that question is what surfaced this one, which is about the credential rather than about where it is returned.

## What the credential is

`ProjectInviteCodeService.generateSecureFiveDigitNumber` mints a five-digit integer, so the whole space is 90,000 values. The column is `@Column(nullable = false, unique = true)`, unique across the deployment rather than per organization, so codes cannot be reused and the space is shared by every tenant.

Redemption is `POST /api/v1/invitation/web/validate/userId/{userId}`, guarded by `@orgSecurity.isSelfUser(#userId)`. That guard is correct and deliberately tenant-independent, because the caller is a user who does not belong to an organization yet. `validateAndUseInviteCode` then resolves the code with `inviteCodeRepository.findByCode(...)`, which carries no organization qualifier. It cannot carry one: the redeemer has no tenant, which is the whole point of the flow. So a code is a bearer credential, and the only thing standing between a value and membership of the organization it belongs to is that the value was not guessed.

Three things compound:

- The space is small enough that exhausting it is not a serious obstacle.
- Uniqueness across tenants means a value that redeems anywhere is a value that redeems somewhere useful.
- There is no rate limit on the redemption endpoint. Cloudflare has one on the Keycloak token endpoint, not on this.

Expiry and `maxUses` bound how long a *known* code stays useful. Neither bounds how many values may be tried.

## What is worth changing

The obvious shape is a longer opaque code (a random string of enough entropy that guessing is not a strategy) rather than a five-digit number. `findByCodeAndOrganization_Id` already exists on the repository and is unused, which suggests someone once intended a per-organization lookup; that would let uniqueness become per-tenant, though it does not by itself help the redeeming user who has no tenant to name.

A rate limit and a failed-redemption counter are worth having regardless of the code's shape, and they are cheaper.

## Why this is not a quick fix

The code value is on the wire in three places: the creation response, the listing, and the mobile redemption form. `echno-core` types it as a number (`invitation-create.ts`, `invitation-service.ts`), and the Flutter app reads it too. Widening it is a contract change across all three, so it wants a decision and a coordinated release rather than a backend patch.

Worth deciding alongside: whether the listing should return code values at all, or only a reveal-on-request for a single code, once the code is expensive enough to be worth protecting.

abhijithanandan:
Closes #699.

## The property, and which half of it this changes

An invite code is a bearer credential: whoever presents a live value joins the organization it belongs to. Redemption resolves it with `findByCode` and **no organization qualifier**, and it cannot have one, because the person redeeming holds no membership yet. That is the state the flow exists to end, and the guard, `@orgSecurity.isSelfUser(#userId)`, is deliberately tenant-independent for exactly that reason. Both are correct.

What is left standing between a value and membership is that the value has to be **known**. That property only holds while the number of values that can be offered is bounded. Nothing bounded it.

There are two ways to restore the bound: make the credential wide enough that knowing it is the only way to have it, or limit how fast values can be offered. The first is a contract change and is scoped below, not made here. **This PR does the second**, and it should not be read as having settled the question: a credential worth guessing is still worth guessing more slowly.

## What already existed

`bucket4j` is a compile dependency with a filter block in `application.yml`, and it has never run: `bucket4j.enabled: false`. Switching it on as written would not have helped, and would have hurt. Its key is `getRemoteAddr()`, which behind nginx, the Cloudflare Tunnel and Cloudflare is the proxy, so one bucket of 20 requests a minute would be shared by the whole deployment; and its scope is `^/api/.*`, below ordinary console use by one person. That is filed separately as #706.

At the edge, `echno-deployment` `tofu/envs/cloudflare` carries a rate limit on the Keycloak token path only, keyed on `ip.src` and `cf.colo.id`. So the pattern exists there, but not on this endpoint.

## What this adds, and where

An `InviteCodeRedemptionLimiter` in the redemption path, not a filter. The service is the only place that knows whether a code was accepted, and the endpoint is the only one that resolves a bearer credential without a tenant, so a limit written there can be exact where a path-matched filter cannot.

**An attempt is charged before the lookup and refunded when the code is accepted.** So a person redeeming the invitation they were actually sent spends nothing however often they do it, and the allowance is in practice a budget for wrong codes. Charging first is what makes it a limit on attempts rather than on the fraction of them that has already finished failing: the token is taken before anything is looked up, so there is no window in which an unbounded number of lookups is already in flight. The charge lives in the bucket rather than the database, so a rejected code rolls the transaction back and leaves the attempt counted, which is the point.

**Two allowances, answering different questions.**

- *Per caller*, keyed on the authenticated principal. The endpoint requires authentication, so there is a principal to key on, and it is a far better key than an address here: for the reason above, an address-keyed limit at the origin would be one shared allowance wearing the name of a per-client one.
- *Per deployment*. A per-caller limit is worth exactly what another account costs, and where self-registration is open that is close to nothing. This ceiling does not move when someone brings more accounts.

The deployment ceiling has a cost worth stating rather than burying: **someone who burns it delays everyone else's redemptions until it refills.** That is why it is set well above any believable volume of genuine wrong codes rather than tight, why it is configurable per deployment, and why zero disables it. It is a deliberate trade, not an oversight.

**Where the counts live** follows the same single switch as every other piece of cross-replica state, `echno.cache.provider`. In process by default; in the shared Redis when it is `redis`, without which a multi-replica deployment multiplies every ceiling by its replica count. `RedisDistributedStateIT` already held the underlying property that two proxy managers over one key draw on one allowance.

**An unreachable bucket store leaves the limit unavailable, not the endpoint refused.** Failing closed would turn an outage in something advisory into an outage in onboarding, and this limit reduces a rate over an already-authenticated endpoint; it is not what decides whether a code is valid.

## Two smaller things in the same path

- **A five-character code that was not digits reached `Integer.parseInt` and answered 500.** The request body constrains the code to exactly five characters and never said they were digits. So a malformed code was met with a server error rather than a refusal, and was not judged as an attempt at all. It is now an `InvalidInviteCodeException`, and it costs an attempt like any other wrong code.
- **`maxUses` and `validityDays` were unbounded ints.** That is how a credential ends up with an expiry far enough out to be no expiry, and a large enough `validityDays` overflows the date arithmetic outright. Bounded to 1..1000 and 1..365. The console sends 7 to 90 days and never sends `maxUses` at all, so the bounds refuse only values that were already nonsense.

## Expiry and single use: both already hold, by default

The issue asked whether a code that never expires or can be redeemed repeatedly is a separate weakness. Checked rather than assumed:

- **Expiry holds.** `validityDays` defaults to 5, `expiryDate` is set from it at generation, and redemption refuses a code whose expiry has passed. The console offers 7 to 90 days.
- **Single use holds.** `maxUses` defaults to 1, redemption refuses once `currentUses >= maxUses`, and the counter is incremented in the same transaction as the join. The shipped default is a single-use invitation that dies in under a week. `ProjectInviteCodeServiceTest.aGeneratedCodeIsSingleUseAndShortLivedByDefault` pins that, so a later edit cannot quietly turn the default code into a permanent one.

So neither is the weakness. Two riders, neither addressed here:

- The use counter carries no `@Version` and no lock, so two concurrent redemptions of one code both read the same count. It holds today because CockroachDB's default isolation is serializable and one of the two transactions retries. That is a property of the database rather than of the code, and it would not hold on a Postgres deployment at read committed. **Reasoned from the source and the engine's documented default, not measured.**
- `isActive` is never flipped false when a code is exhausted. Harmless, since the count check covers it, but the flag then means less than its name suggests.

## Widening the code: scope, cost, and one thing the issue assumed that is not so

The issue treated widening as blocked by shipped clients typing the code as a number. **They do not. Both clients already type it as a string.**

| | invite code type |
|---|---|
| `echno-backend` `ProjectInviteCode.code` / `ProjectInviteCodeDto.code` | `int` |
| `echno-core` `Invitation.inviteCode` | `string` |
| `echno-core` `InvitationResponseSchema.code` | `z.string().nullish()` |
| Flutter `echno` `Invitation.inviteCode` | `String?` |
| Flutter `echno` `validateInviteCode({required String code})` | `String` |

The backend is the odd one out, and its numeric response is what `echno-core`'s own schema rejects today; filed as tornotron/echno-core#97. So widening the backend's code to an opaque string moves it **towards** its clients.

Two further findings that change the cost:

- The Flutter app posts to `/invitation/generate` and `/invitation/validate`. The backend publishes `/api/v1/invitation/web/generateCode/organizationId/{id}` and `/api/v1/invitation/web/validate/userId/{userId}`. That surface is already off-contract and cannot be working against the current backend; last push 2025-11.
- `echno-web`'s join form already uppercases the input and calls the code case-insensitive, which is meaningless for a five-digit integer and would become meaningful.

What it would actually take, in this repository:

1. One Liquibase changeset in the next version folder: `project_invite_code.code` from `int` to `varchar`, keeping the unique constraint, casting existing rows. `ddl-auto` is `validate`, so the changeset is the whole migration.
2. `ProjectInviteCode.code` and `ProjectInviteCodeDto.code` to `String`; `findByCode` and `findByCodeAndOrganization_Id` likewise.
3. `InviteCodeValidationDto`: replace `@Size(min=5,max=5)` with the new shape's constraint. `parseCode` then disappears.
4. Replace `generateSecureFiveDigitNumber` with a generator over an unambiguous alphabet, **and add the collision retry that is missing today**: nothing retries a duplicate now, so a collision on the unique constraint is a 500, and at 90,000 values that stops being hypothetical once a deployment holds a few hundred live codes.
5. `docs/openapi.json` regenerates.

`echno-core` needs no type change; it needs #97 either way. The open decisions are the shape (length, alphabet, whether it stays human-typeable, since it is read aloud and typed on a phone) and whether the listing should keep returning code values or move to reveal-on-request once the code is expensive enough to protect. Those are for you, not for this PR.

## Recommended follow-up at the edge

Worth mirroring the Keycloak rule in `echno-deployment` onto `/api/v1/invitation/web/validate/`, keyed on `ip.src` as that one is. The edge is the one place a real client address is available, and it stops volume before the origin. It is a backstop rather than the control: it cannot tell an accepted code from a refused one, and it does not exist for a deployment that is not behind Cloudflare, which is why the load-bearing limit is the one in this PR.

## Tests

|---|---|---|
| `InviteCodeRedemptionLimiterTest.aCallerIsRefusedOnceTheirAllowanceIsSpent` | the per-caller allowance actually runs out | new surface, nothing to measure |
| `InviteCodeRedemptionLimiterTest.theRefusalCarriesHowLongToWait` | the refusal says when it ends, so a client need not poll | new surface |
| `InviteCodeRedemptionLimiterTest.theRefusalSaysNothingAboutTheAllowanceItself` | the message does not pace the next run | new surface |
| `InviteCodeRedemptionLimiterTest.oneCallerRunningOutDoesNotRefuseAnother` | one account cannot lock out onboarding through the per-caller bucket | new surface |
| `InviteCodeRedemptionLimiterTest.theDeploymentAllowanceRefusesEvenACallerWithBudgetLeft` | the ceiling holds across accounts, not only within one | new surface |
| `InviteCodeRedemptionLimiterTest.aDeploymentRefusalDoesNotSpendTheCallersOwnAllowance` | a queue nobody caused does not also spend their budget | new surface |
| `InviteCodeRedemptionLimiterTest.anAcceptedCodeCostsNothing` | a real redemption never approaches the limit | new surface |
| `InviteCodeRedemptionLimiterTest.aRefundOnAnUntouchedAllowanceGrantsNothingExtra` | a refund cannot manufacture allowance | new surface |
| `InviteCodeRedemptionLimiterTest.zeroDisablesAnAllowanceEntirely` | the documented off switch is really off | new surface |
| `InviteCodeRedemptionLimiterTest.anUnreachableBucketStoreLetsTheRequestThrough` | an unavailable limit is not an onboarding outage | new surface |
| `ProjectInviteCodeServiceTest.validateAndUse_pastTheAttemptAllowance_isRefusedWithoutLookingTheCodeUp` | the third attempt never reaches the repository | new surface |
| `ProjectInviteCodeServiceTest.validateAndUse_acceptedCode_costsNoAttempt` | five successful redemptions on an allowance of one | new surface |
| `ProjectInviteCodeServiceTest.validateAndUse_everyKindOfRefusal_spendsTheAttempt` | expired, inactive and exhausted all cost an attempt, so none is a free probe | new surface |
| `ProjectInviteCodeServiceTest.validateAndUse_aCodeThatIsNotANumber_isRefusedRatherThanFailing` | a five-letter code is a refusal, not a 500 | **reasoned**; the old path threw `NumberFormatException` out of the service, which the test would have caught as the wrong type |
| `ProjectInviteCodeServiceTest.validateAndUse_aCodeThatIsNotANumber_stillSpendsTheAttempt` | a malformed code is judged like any other | new surface |
| `ProjectInviteCodeServiceTest.aGeneratedCodeIsSingleUseAndShortLivedByDefault` | expiry and single use are the shipped default | passes before and after; it exists to keep that true |
| `GlobalExceptionHandlerTest.tooManyAttempts_isA429CarryingRetryAfter` | 429 with `Retry-After`, and a body that says nothing about the allowance | new surface |
| `GlobalExceptionHandlerTest.tooManyAttempts_neverAsksTheClientToRetryImmediately` | a sub-second wait still asks for one second | new surface |
| `InviteCodeRedemptionRefusalTest` (2 cases) | the status and the header actually reach the wire through the controller | new surface |
| `RedisAttemptBucketsLazinessTest` (3 cases) | construction does not dial Redis; a request against an absent Redis fails there and the limiter lets it through | **measured** (run 34051839238: the eager version failed `RedisUnreachableBootIT.contextLoadsWithoutAReachableRedis` with a `RedisConnectionException` during context refresh) |

The one measured red is the Redis regression, caught by an existing test on the first CI run and fixed in the second commit. The rest is new surface, where there is no earlier behaviour to measure against: an unlimited endpoint has no failing case to record, only an absent limit.

## OpenAPI

Three deltas, each taken byte for byte from the build's own `openapi-served` artifact and verified against it rather than typed: the bounds now published on `maxUses` and `validityDays`, the validate operation's description, and its `429`. The file was not copied wholesale, because the artifact is generated from the pull request's **merge ref** against whatever `development` was at the time, so it predates two merges since and taking it entire would have reverted them.

Producing it also caught something worth having caught. The served `429` documented its body as an `OrganizationDto`. Springdoc fills in an undeclared error body from the advice method's return type, and this handler returns a `ResponseEntity` so it can carry `Retry-After`, which tells springdoc nothing; it fell back to the operation's own return type. The schema is now spelt out on the annotation. The published contract is what `echno-core` checks its call sites against, so a wrong schema there is worse than an absent one.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -56505,7 +56505,7 @@
             "content": {
               "*/*": {
                 "schema": {
-                  "$ref": "#/components/schemas/OrganizationDto"
+                  "$ref": "#/components/schemas/ProblemDetail"
                 }
               }
             },
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -7182,6 +7182,8 @@
           },
           "maxUses": {
             "format": "int32",
+            "maximum": 1000,
+            "minimum": 1,
             "type": "integer"
           },
           "phone": {
@@ -7202,6 +7204,8 @@
           },
           "validityDays": {
             "format": "int32",
+            "maximum": 365,
+            "minimum": 1,
             "type": "integer"
           }
         },
@@ -56403,7 +56407,7 @@
     },
     "/api/v1/invitation/web/validate/userId/{userId}": {
       "post": {
-        "description": "Validates the supplied invite code and joins the user to the organization it grants access to.",
+        "description": "Validates the supplied invite code and joins the user to the organization it grants access to. Attempts are limited: one is charged before the code is looked up and given back when the code is accepted, so a successful redemption costs nothing and repeated wrong codes are answered with 429 and a Retry-After header.",
         "operationId": "validateInviteCode",
         "parameters": [
           {
@@ -56496,6 +56500,16 @@
               }
             },
             "description": "Unprocessable Entity"
+          },
+          "429": {
+            "content": {
+              "*/*": {
+                "schema": {
+                  "$ref": "#/components/schemas/OrganizationDto"
+                }
+              }
+            },
+            "description": "The attempt allowance is spent; Retry-After says when it refills"
           },
           "500": {
             "content": {
--- a/src/main/java/org/tornotron/echno_backend/common/exception/GlobalExceptionHandler.java
+++ b/src/main/java/org/tornotron/echno_backend/common/exception/GlobalExceptionHandler.java
@@ -6,8 +6,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.dao.CannotAcquireLockException;
 import org.springframework.dao.DataIntegrityViolationException;
+import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ProblemDetail;
+import org.springframework.http.ResponseEntity;
 import org.springframework.security.authorization.AuthorizationDeniedException;
 import org.springframework.security.authorization.ExpressionAuthorizationDecision;
 import org.springframework.validation.FieldError;
@@ -255,6 +257,25 @@ public class GlobalExceptionHandler {
     public ProblemDetail handleInvalidInviteCodeException(InvalidInviteCodeException ex, WebRequest request) {
         logger.error("Invalid Invite Code: ", ex);
         return problem(HttpStatus.BAD_REQUEST, "Invalid Invite Code", ex.getMessage(), request);
+    }
+
+    /**
+     * An operation whose attempt allowance is spent.
+     *
+     * <p>Returned as a {@link ResponseEntity} rather than a bare {@link ProblemDetail} so it can
+     * carry {@code Retry-After}, which is the one thing a well-behaved client needs and the one
+     * thing a wrong number in the body cannot supply. The detail is deliberately uninformative
+     * about how much allowance remains, since that would pace whoever provoked it; the header is
+     * not, because it says only when the refusal ends, which the client is entitled to know.
+     */
+    @ExceptionHandler(TooManyAttemptsException.class)
+    public ResponseEntity<ProblemDetail> handleTooManyAttempts(TooManyAttemptsException ex, WebRequest request) {
+        logger.warn("Attempt allowance exhausted: {}", ex.getMessage());
+        ProblemDetail body = problem(HttpStatus.TOO_MANY_REQUESTS, "Too Many Requests", ex.getMessage(), request);
+        long retryAfterSeconds = Math.max(1, ex.getRetryAfter().toSeconds());
+        return ResponseEntity.status(HttpStatus.TOO_MANY_REQUESTS)
+                .header(HttpHeaders.RETRY_AFTER, Long.toString(retryAfterSeconds))
+                .body(body);
     }
 
     /**
--- a/src/main/java/org/tornotron/echno_backend/common/exception/TooManyAttemptsException.java
+++ b/src/main/java/org/tornotron/echno_backend/common/exception/TooManyAttemptsException.java
@@ -1,0 +1,26 @@
+package org.tornotron.echno_backend.common.exception;
+
+import java.time.Duration;
+
+/**
+ * Raised when a caller has used up the attempts allowed for an operation within its window.
+ *
+ * <p>Carries how long the caller should wait, which becomes the {@code Retry-After} header. The
+ * message deliberately says only that the allowance is spent, never how many attempts remain or
+ * which of the allowances refused, because that would tell whoever provoked it how to pace the
+ * next run.
+ */
+public class TooManyAttemptsException extends RuntimeException {
+
+    private final transient Duration retryAfter;
+
+    public TooManyAttemptsException(String message, Duration retryAfter) {
+        super(message);
+        this.retryAfter = retryAfter;
+    }
+
+    /** How long until the allowance has refilled enough for one more attempt. */
+    public Duration getRetryAfter() {
+        return retryAfter;
+    }
+}
--- a/src/main/java/org/tornotron/echno_backend/common/ratelimit/AttemptBuckets.java
+++ b/src/main/java/org/tornotron/echno_backend/common/ratelimit/AttemptBuckets.java
@@ -1,0 +1,34 @@
+package org.tornotron.echno_backend.common.ratelimit;
+
+import io.github.bucket4j.Bucket;
+import io.github.bucket4j.BucketConfiguration;
+
+/**
+ * Supplies a token bucket for a named key, so a caller counting attempts does not have to know
+ * whether the count lives in this process or in the shared Redis.
+ *
+ * <p>Both implementations hand back a {@link Bucket}, which is the interface a local bucket and a
+ * Redis-backed {@code BucketProxy} have in common. Which one is in use follows the same single
+ * switch as every other piece of cross-replica state, {@code echno.cache.provider}: the default
+ * {@code caffeine} keeps the counts in process, {@code redis} shares them across pods. A limit
+ * held in process is still a limit, but on a multi-replica deployment it is one allowance per
+ * replica rather than one allowance, so the effective ceiling is multiplied by the replica count.
+ * That is the reason the redis path exists rather than a preference.
+ *
+ * <p>The key namespace is the caller's to choose and is expected to be prefixed, because on the
+ * redis path the keys of every limiter share one keyspace.
+ */
+public interface AttemptBuckets {
+
+    /**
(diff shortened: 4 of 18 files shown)


abhijithanandan:
@coderabbitai review

Issue #710, opened 2026-09-06
Two employee-directory reads sit on bare membership, and one contradicts the write beside it


abhijithanandan:
Two observations from tracing #690, recorded together because both are about `EmployeeControllerWeb` and neither was in scope for the guard repairs that shipped in #703.

## Any member can enumerate who holds which role

`GET /employee/web/{employeeId}/roles` is on bare tenant membership. So any signed-in member can walk employee ids and build a map of who holds which organization role. Nothing crosses a tenant, since `Employee` carries the `orgFilter`, and no role is granted by reading. What it gives away is the shape of the organization's permissions, which is reconnaissance rather than access.

Worth deciding rather than reflexively narrowing: some of this is legitimately public inside a company, and a directory that hides who the site manager is would be worse than one that shows it. The question is whether the full role set per person is the right granularity.

## An employee cannot read the record they are allowed to edit

`GET /employee/web/{id}` is role-gated while its `PATCH` is `isSelfOrHasAnyOrgRole`. So a person may edit their own record and not read it back. That is the same shape as the defect fixed in `eec6c37`, where reads used membership and writes used roles, leaving the bootstrap admin able to write a project and then refused reading it.

The mobile twin is worse and is a different problem: it refuses everybody, under an authority the realm has no mechanism to issue. That is the phantom-guard mechanism established in #684, so it should be repaired the same way rather than deleted.

## Not in scope here

The question of whether a `project-manager` should see colleagues' salary, date of birth, address and phone number through `EmployeeDto` is a product decision and is filed separately.

abhijithanandan:
Both halves answered. The read/write asymmetry is repaired in #713; the roles enumeration is closed as a decision and stays on tenant membership.

## The roles enumeration stays open, deliberately

`GET /employee/web/{employeeId}/roles` keeps `isMemberOfCurrentTenant()`. Knowing who the site manager or the HR admin is, is how a person finds who to route a request to, and a directory that hides it is worse than one that shows it. The answer crosses no tenant, because `Employee` carries the org filter, and it grants nothing: reading a role is not holding one. The employee ids it is walked with are already enumerable through the lookup list the pickers use, so narrowing this endpoint alone would cost the directory something real and buy back only the ordering of a list somebody can otherwise assemble.

The reasoning is recorded on the method rather than only here, so the next reader finds it where the guard is. The 403 text was saying the caller lacked a role while the guard checked membership, and that is corrected with it.

[...]

Issue #709, opened 2026-09-06
An or-clause makes the role check dead in ProjectControllerWeb


abhijithanandan:
`ProjectControllerWeb:373` combines a membership check and a role check with `or`. Membership is the weaker of the two and is satisfied by everyone the role check would accept, so the role clause can never change the outcome. It reads as a two-part guard and behaves as a one-part one.

That is worth fixing whether or not the endpoint should be role-gated, because the next person to read it will believe the role is enforced. Either the `or` is wrong and it should be `and`, or the role clause is redundant and should be deleted so the guard says what it does.

Decide which by asking what the endpoint is for rather than by making the expression tidy. Note that the read-versus-write asymmetry in this controller has bitten before: commit `eec6c37` fixed reads that used `isMemberOfCurrentTenant()` while writes used `hasAnyOrgRole()`, so a bootstrap admin could write a project and then be refused reading it.

Found while tracing #690.

abhijithanandan:
The premise holds. The reason is one level below the expression, and it is worth stating because it changes what the right fix is.

## Why the role clause cannot decide anything

Both clauses open on `TenantContext.getCurrentOrgId()` and return false when it is null. `TenantFilter` sets it in exactly two branches and both confirm an `ORG_MEMBER_{orgId}` authority first: the `X-Organization-Id` branch refuses with 403 without it, and the header-less branch reads the id out of the `ORG_MEMBER_` authorities themselves. `JwtAuthConverter` mints `ORG_MEMBER_{id}` from the flat group `/org-{id}` and `ORG_{id}_ROLE_{role}` from the subgroup `/org-{id}/{role}`, independently, so a user placed only in a subgroup really does arrive holding the role and not the membership. That caller reaches the controller with no organization in force and fails both clauses.

So `hasAnyOrgRoleForCurrentTenant` implies `isMemberOfCurrentTenant` on every request that reaches a controller. Pinned in `TenantFilterTest.theRoleGuardCannotSucceedWhereTheMembershipGuardFails`, run against the real `OrganizationSecurityService` inside the filter chain rather than against a mock.

## Which of the two resolutions

The clause is removed, not turned into an `and`. An `and` would narrow these reads to role-holders and shut out the ordinary members they exist for, which is the opposite of what the endpoints are for: every write on this controller is already `system-admin, project-manager`, and the reads are the tenant-wide half.

## The thing worth knowing that came out of it

[...]

Issue #695, opened 2026-09-06
A project-manager can raise the adjustment that closes a transfer variance but cannot read the transfer it names


abhijithanandan:
Found while tracing the whole workflow for the stock adjustment source reference (#694, raised from web#398), not while fixing an endpoint that returned 403.

## The mismatch

Every read on `SiteTransferControllerWeb` is gated on `system-admin` alone:

```
@PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin')")
```

That covers `GET /{id}`, the listings, the by-project lookups, the status history, and the receive and cancel actions.

`StockAdjustmentControllerWeb` is gated differently: reads are open to any tenant member, and writes and decisions are open to `system-admin` **or** `project-manager`. So a project-manager may raise, approve and reject a stock adjustment.

A stock adjustment can now name the site transfer whose receipt variance it closes. That produces a workflow a project-manager can start but cannot see:

- they may raise the adjustment naming transfer 31, and the server will accept it, because the reference is resolved server side against their own organization
- they cannot open `GET /api/v1/site-transfers/web/31` to check what they are closing
- they cannot follow the reference back from the adjustment to the transfer
- a prefill built from the transfer fails for them, because the fetch of the transfer is refused rather than the adjustment being refused

The immediate blast radius is small, because the link that starts this is offered on the transfer detail screen, which a project-manager cannot open either. It is the shape that matters: the same shape as #666, where a project-manager could raise and approve a stock adjustment but not fill in the form, and #650 and #672 in the same family.

## The wider question underneath it

`system-admin` for every read of a site transfer looks tighter than the module needs. A project manager cannot see transfers into or out of their own project, which is material arriving on their site. Whether that is deliberate or a leftover is worth deciding rather than inheriting.

## Not fixed in #694 on purpose

Widening site transfer reads is an authorization decision with its own blast radius and does not belong inside a contract change. Filing it so it is decided rather than discovered later.

Suggested shape if it is widened: match the stock adjustment reads, so a tenant member may read a transfer while receiving and cancelling stay on the elevated role, since those move stock. `SiteTransferControllerWebAuthzTest` pins the current guard and would need to move with it.

abhijithanandan:
Traced this alongside #698, #709 and #710. The finding stands and the workflow gap is real, but it cannot be closed without settling #680, so it is left open here rather than fixed.

## Why it cannot be separated from #680

[...]

PR #713, opened 2026-09-06
Repair the employee read, the leave-policy duplicate target, and the project read guards

### LINKED ISSUE #698: Duplicating a leave policy into another organization checks employment there, not a role there
Found while sweeping for the siblings of #687 (fixed in #696). This one is not the same defect, which is why it is filed rather than folded into that fix: there **is** an entitlement check here, but it establishes something weaker than what the guard claims.

## The endpoint

`POST /api/v1/leave-policies/{policyId}/duplicate?targetOrganizationId=`, `LeavePolicyController:186`, and its web twin `POST /api/v1/leave-policies/web/duplicate`, `LeavePolicyControllerWeb:169`. Both:

```java
```

`LeavePolicyService.duplicatePolicy:334` then resolves the target at `:339`:

```java
Organization targetOrg = organizationRepository.findByIdAndUserEmail(targetOrganizationId, userContextService.getCurrentUserEmail())
```

and writes the copy into it at `:351` and `:372`.

## The mismatch

Two different organizations are in play and the two checks are each applied to the wrong one:

- The **role** check (`system-admin` or `hr-admin`) is evaluated against the caller's *current* tenant.
- The **entitlement** check, `findByIdAndUserEmail`, asserts only that the caller has an `Employee` row in the *target* organization. Any employee satisfies it, whatever their role.

So a user who is an hr-admin in organization A and an ordinary employee in organization B can write a leave policy into B while holding no administrative role there. Unlike #687 this is not reachable by someone with no relationship to the target, which is why it is a lesser finding, but the guard reads as though the role applied to the target and it does not.

A second, smaller thing in the same method: the duplicate-code check at `:343` is keyed on `targetOrganizationId` and runs against `LeavePolicy`, which is filtered by `orgFilter`. For any organization other than the current tenant the predicate becomes unsatisfiable, so the uniqueness rule is silently skipped on exactly the path that needs it.

## The decision this needs

**Is duplicating a leave policy into another organization a supported feature at all?**

- If yes, the role check has to move to the target: `@orgSecurity.hasAnyOrgRole(#targetOrganizationId, 'system-admin', 'hr-admin')`. The duplicate-code check then has to run outside the org filter, or it stays broken.
- If no, drop `targetOrganizationId` and duplicate within the current tenant, which makes both problems disappear.

Tightening the current-tenant guard alone changes nothing and would be the wrong reading of it.

## Not yet established

Whether Hibernate applies `orgFilter` to the `JOIN o.employees e` inside `OrganizationRepository.findByIdAndUserEmail:51`. If it does, the query can never resolve a non-current organization and the endpoint is dead rather than exploitable, which would change this from a security finding to a dead-feature one. That needs a run against a database rather than a reading, and it is worth doing first because it decides which of the two options above is even reachable today.

abhijithanandan:
Closes #698, #709, #710. #695 is answered in its own thread rather than in code.

Four authorization issues taken together because they are all guard changes on the same layer. The tests were pushed as their own commit ahead of the fix, so what each guard decided before it changed is measured rather than assumed.

## #710, the read/write asymmetry — repaired

`GET /employee/web/{id}` was gated on the system-admin, hr-admin or project-manager role while the `PATCH` beside it gates on `isSelfOrHasAnyOrgRole`, so a person could edit their own record and then be refused reading it back. It now carries the same self branch and the same three roles, which widens it by exactly one caller: the employee themselves. A member holding no elevated role still cannot open a colleague's record.

The mobile twin asked for `employee:read` or `employee:admin`, bare `resource:scope` authorities that `JwtAuthConverter` mints only from an RPT's `authorization` claim, which this realm has no scopes to fill. It refused every caller from the day it was written. Same mechanism as #684, repaired the same way rather than deleted. The six other phantom guards on that controller are left alone and filed as #716: four are writes, and moving a phantom write guard from refusing everybody to admitting a role is a widening decision of its own.

## #710, the roles enumeration — closed as a decision, not narrowed

`GET /employee/web/{employeeId}/roles` stays on tenant membership. Knowing who the site manager is, is how work gets routed; the answer crosses no tenant and grants nothing; and the ids it is walked with are already enumerable through the lookup list. Narrowing it would cost the directory something real and buy back only the ordering of a list somebody can otherwise assemble. The reasoning is recorded on the method. Its 403 text said "role" while the guard checked membership, and that is corrected.

## #698, the duplicate target — repaired, and the endpoint is worse than filed

The role was checked against the caller's current tenant while entitlement to the target came from `findByIdAndUserEmail`, which establishes an `Employee` row there and nothing about what the caller may do. Both twins now require the role in the current tenant **and** in the target.

The issue left open whether the target is even reachable. `OrganizationLookupUnderTheOrgFilterIT` settles it against a real database: **the org filter does not reach an entity joined explicitly in HQL**, and the joined `Employee` is never selected so the fail-closed load listener gets no post-load to judge either. Neither mechanism scopes that query, and a foreign organization comes back silently. So this was a live cross-organization write path, and the repair closes it rather than tidying a dead one.

Not fixed here, filed as #718: the duplicate-code check beside it runs against `LeavePolicy`, which *does* carry the filter as a query root, so it is unsatisfiable for any target other than the current tenant and the uniqueness rule is skipped on exactly the path that needs it.

## #709, the dead `or` — clause removed

Seven read guards read "member OR system-admin/project-manager". The second clause could never decide anything: both sides open on `TenantContext.getCurrentOrgId()`, and `TenantFilter` sets it only after confirming an `ORG_MEMBER_` authority, so a role-holder who is not a member arrives with no organization in force and fails both. Removed rather than turned into an `and`, which would have shut out the ordinary members these reads exist for.

The invariant is pinned in `TenantFilterTest`, against the real `OrganizationSecurityService` inside the filter chain. The branch `ProjectControllerWebAuthzTest` used to assert goes with it: it passed only because `@orgSecurity` is mocked there, which lets the two guards be set into a state no request can reach. That also means `eec6c37`'s intent, letting a role-holder read what they may already write, was never achieved; making it work is a change to tenant resolution rather than to a controller, and is left as a decision. The same expression stands on 36 more guards across ten controllers, filed as #717 and sequenced after that decision.

## #695 — not changed

Widening site-transfer reads to `project-manager` is one row of #680's table and settles its direction, so it is left to that decision. Reasoning, the verified severity, and the one option that does not touch #680 are in the #695 thread.

## Tests

| test | what it covers |
|---|---|
| `EmployeeReadAuthzTest` | reads on both employee twins: self, a plain member refused a colleague, a role-holder allowed one |
| `LeavePolicyDuplicateAuthzTest` | the duplicate guard on both leave-policy twins, each side of the pair and both together |
| `OrganizationLookupUnderTheOrgFilterIT` | whether the org filter reaches an explicit HQL join, with an unfiltered control |
| `TenantFilterTest` | a role without membership resolves no tenant, so the role guard cannot outrun the membership guard |

The OpenAPI document is regenerated for the sixteen response and operation descriptions that moved with the guards.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -35262,7 +35262,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller is not a member of the current tenant"
           },
           "404": {
             "content": {
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -35197,7 +35197,7 @@
     },
     "/api/v1/employee/web/{employeeId}/roles": {
       "get": {
-        "description": "Returns the set of organization roles held by the given employee.",
+        "description": "Returns the set of organization roles held by the given employee. Open to any member of the organization, since knowing who holds which role is how work is routed.",
         "operationId": "getOrgRoles",
         "parameters": [
           {
@@ -35685,7 +35685,7 @@
         ]
       },
       "get": {
-        "description": "Returns a single employee's full details.",
+        "description": "Returns a single employee's full details. Callable by the employee themselves or by a system admin, HR admin or project manager.",
         "operationId": "readAnEmployee_1",
         "parameters": [
           {
@@ -35737,7 +35737,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller is neither the employee nor a system admin, HR admin or project manager"
           },
           "404": {
             "content": {
@@ -36030,7 +36030,7 @@
         ]
       },
       "get": {
-        "description": "Returns a single employee including their personal details, roles and status.",
+        "description": "Returns a single employee including their personal details, roles and status. Callable by the employee themselves or by a system admin, HR admin or project manager.",
         "operationId": "readAnEmployee",
         "parameters": [
           {
@@ -36082,7 +36082,7 @@
                 }
               }
             },
-            "description": "Caller lacks the employee read or admin authority"
+            "description": "Caller is neither the employee nor a system admin, HR admin or project manager"
           },
           "404": {
             "content": {
@@ -66974,7 +66974,7 @@
     },
     "/api/v1/leave-policies/web/duplicate": {
       "post": {
-        "description": "Copies the policy's quota, accrual and eligibility rules into a new policy owned by the target organization. Returns the newly created policy.",
+        "description": "Copies the policy's quota, accrual and eligibility rules into a new policy owned by the target organization. Returns the newly created policy. The caller must hold the system-admin or hr-admin role in their own organization and in the target.",
         "operationId": "duplicatePolicy_1",
         "parameters": [
           {
@@ -67035,7 +67035,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller lacks the required role in the current tenant or in the target organization"
           },
           "404": {
             "content": {
@@ -67906,7 +67906,7 @@
     },
     "/api/v1/leave-policies/{policyId}/duplicate": {
       "post": {
-        "description": "Copies the policy's quota, accrual and eligibility rules into a new policy owned by the target organization. Returns the newly created policy.",
+        "description": "Copies the policy's quota, accrual and eligibility rules into a new policy owned by the target organization. Returns the newly created policy. The caller must hold the system-admin or hr-admin role in their own organization and in the target.",
         "operationId": "duplicatePolicy",
         "parameters": [
           {
@@ -67967,7 +67967,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller lacks the required role in the current tenant or in the target organization"
           },
           "404": {
             "content": {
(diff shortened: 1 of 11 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Three findings, taken one at a time.

**The roles endpoint's 403 text.** Valid, and mine. The operation description said any member may call it while the 403 beside it still named a role. Fixed, with the OpenAPI document regenerated to match.

**The two remaining phantom guards on `EmployeeController`.** The observation is right and it is deliberately out of scope here. `readAllEmployees`, `readEmployeesByOrganizationId`, create, both patches and delete are all still on `employee:*` authorities that this realm cannot issue, and all six refuse every caller. They are filed as #716 rather than folded in, because four of the six are writes and repairing a phantom write guard moves it from refusing everybody to admitting a role, which is a widening decision rather than a substitution. The shape suggested for `readEmployeesByOrganizationId`, pairing `isCurrentTenant(#id)` with the role check, is the right one for an endpoint that takes an organization id from the caller, and has been added to that issue.

**`*IT` classes not running.** Not correct. Gradle's JUnit Platform detection selects test classes by their annotations, not by a name pattern, so no `*IT` include is needed and none is configured. Run 34060437075 on this branch executed 364 test cases in `*IT` classes, including all three in `OrganizationLookupUnderTheOrgFilterIT`, all passing. The build-enforced coverage the finding asks for is already there.

PR #714, opened 2026-09-06
Remove the rate-limit filter that never ran

### LINKED ISSUE #706: The configured Bucket4j filter has never run, and switching it on as written would refuse the whole API
Found while adding a targeted attempt limit to invite-code redemption in #699. That PR deliberately did not switch this on, and the reasons are worth recording rather than rediscovering.

## What is configured

`application.yml` carries a Bucket4j filter block:

```yaml
bucket4j:
  enabled: false
  filters:
    - cache-name: rate-limit-buckets
      url: ^/api/.*
      rate-limits:
        - cache-key: "getRemoteAddr()"
          bandwidths:
            - capacity: 20
              time: 1
              unit: minutes
```

The dependency is real (`com.giffing.bucket4j.spring.boot.starter`), the cache is declared under `spring.cache.cache-names`, `logback-spring.xml` sets the starter to DEBUG, and `EchnoCacheEnvironmentPostProcessor` already switches it to `redis-lettuce` when `echno.cache.provider=redis`. It is a complete piece of wiring that has never run.

## Two things wrong with switching it on as it stands

**The key is the proxy.** `getRemoteAddr()` is the socket peer. Every deployment we run puts something in front of the application: the compose staging is behind nginx, IITM is behind a Cloudflare Tunnel, and the DO shape is behind Cloudflare. So the value is the proxy's address for every request, and one bucket of 20 requests a minute would be shared by the entire deployment. There is no `server.forward-headers-strategy` set, so `X-Forwarded-For` is not being honoured either.

**The scope is everything.** `^/api/.*` at 20 a minute is below ordinary use of the console by one person, never mind a site.

Together: enabling it as written would take the whole API down rather than limit anything.

## What to decide

1. **Set `server.forward-headers-strategy`** (or configure the Tomcat `RemoteIpValve`) so the application sees the real client address at all. This is worth doing on its own merits, since every access log line currently records the proxy.
2. **Then decide whether a blanket per-IP ceiling is wanted**, and at what number. If it is, it belongs on the unauthenticated surface (`/auth/register` is `permitAll()` and unthrottled) rather than uniformly across `/api/**`, where the authenticated principal is a better key than any address.
3. **Or delete the block**, if the intent is that volumetric limiting lives at the edge. Cloudflare already carries one on the Keycloak token path, codified in `echno-deployment` `tofu/envs/cloudflare`, and adding rules there is cheap.

Leaving it configured-but-off is the one option worth avoiding: it reads as a control that exists, and it is not one.

For contrast, the limit added in #699 is keyed on the authenticated principal, scoped to the one endpoint that resolves a bearer credential, and charged per attempt with a refund on success. It does not depend on any of the above.

abhijithanandan:
Closes #706.

## Why it never ran

One line, and only one. The starter's servlet-filter auto-configuration carries `@ConditionalOnBucket4jEnabled`, which is `@ConditionalOnProperty(prefix = "bucket4j", name = "enabled", matchIfMissing = true)`. `application.yml` sets `bucket4j.enabled: false`, so the configuration class is never considered and no filter is ever registered. Everything else in the wiring is real and complete: the dependency, the `rate-limit-buckets` cache name, the DEBUG logger, and the environment post-processor that pointed the buckets at Redis. That is what made it read as a control.

The flag is also load bearing in the other direction, which matters for what "delete the block" can mean: because the condition matches when the property is missing, removing the line would **enable** the starter rather than leave things as they are. So the block goes and the flag stays.

## Why switching it on would have refused the API

The issue gave one reason. Checking it turned up a correction and a stronger reason underneath.

**The correction.** The issue says no `server.forward-headers-strategy` is set. It is: `application.yml` has carried `forward-headers-strategy: FRAMEWORK` since March 2025. So `getRemoteAddr()` is not the socket peer. Spring Boot registers `ForwardedHeaderFilter` at `Integer.MIN_VALUE` and the starter's filter defaults to `Integer.MIN_VALUE + 10`, so the forwarding filter runs first and the bucket4j filter would have seen the wrapped request, whose `getRemoteAddr()` comes from `X-Forwarded-For`.

**The stronger reason.** That does not rescue it, because the console does not put a browser address in that header. `echno-web` is a server-side BFF: the browser calls the Next route, and `app/api/v1/[...path]/route.ts` opens its own connection to `backend.echno.xyz`, forwarding exactly three headers, `accept`, `accept-language` and `x-request-id`. None is a client address. Whatever the hop in front of the backend writes into `X-Forwarded-For` therefore names the Next server, identically for every signed-in person. **One bucket of 20 requests a minute for the whole console**, not one per client. And an address arriving in a header is only as trustworthy as the hop that wrote it, which is a general property of keying on an address at an origin rather than something specific to this deployment.

**And the number is below one person's use.** `^/api/.*` at 20 a minute with `refill-speed: interval`, so the allowance returns only at the minute boundary rather than trickling back. A single console screen spends more than that.

Together: the whole API, one shared allowance, twenty a minute. Not a limit, an outage.

## Repair or delete

Deleted. The reason is the position, not the numbers, so a repaired version of the same thing would have to solve a problem it is not placed to solve. The starter's filter runs before Spring Security, so it has no principal and an address is the only key available to it, and at this origin an address is not a client.

## Where rate limiting belongs

Two layers, and this was the third.

**Volumetric limiting at the edge.** Cloudflare is the only place a real client address exists, and the only place that stops load before it reaches the origin. `echno-deployment` `tofu/envs/cloudflare` already carries the Keycloak token rule and #75 proposes the same shape for invite-code redemption. Its stated limitation is the right one: the edge cannot tell an accepted request from a refused one, and a deployment not behind Cloudflare gets nothing.

**Limiting by meaning in the service.** Where there is a principal to key on and an outcome to charge for. `InviteCodeRedemptionLimiter` is the worked example: keyed on the authenticated principal, scoped to the one endpoint that resolves a bearer credential, charged per attempt and refunded on success, so an ordinary redemption spends nothing.

**Nothing in between.** A path-matched filter at the origin sits between the two and can do neither job. That is what is being removed.

The one gap this leaves is worth naming rather than implying it away: `POST /api/v1/auth/register` is `permitAll()` and has no limit at either layer. It is the only unauthenticated write surface, so it is exactly where the edge layer earns its place, and it cannot be covered from here for the same reason as everything above. Recorded on echno-deployment#75, which is where that layer is being built.

## What went with the block

- `rate-limit-buckets` under `spring.cache.cache-names`, which backed the filter and nothing else. There are no live `@Cacheable` sites, and naming caches there fixes the set the manager will serve and refuses any other, so an empty list is both accurate and friendlier to the first one that arrives.
- The `com.giffing.bucket4j.spring.boot.starter` DEBUG logger, which now has nothing to log. The surviving limiter is built on `io.github.bucket4j`, not the starter.
- `EchnoCacheEnvironmentPostProcessor` and its `spring.factories` entry. Its only job was defaulting `bucket4j.cache-to-use` to `redis-lettuce`, a property read solely by the starter's auto-configuration. The attempt buckets reach Redis through `AttemptBucketsConfig`, on the same `echno.cache.provider` switch, so the behaviour it existed to produce is unchanged.

Kept: the `bucket4j` dependency, which the surviving limiter is built on, and `bucket4j.enabled: false` for the reason above. `RedisConfig`'s javadoc said the `RedisClient` bean was for the starter's auto-configuration; it is for `AttemptBucketsConfig`, and now says so.

## Tests

| Test | What it holds | Red before |
|---|---|---|
| `PathMatchedRateLimitAbsenceTest.noFilterBlockIsConfiguredOverTheApi` | no path-matched limit is configured over `/api` | **measured** (run 34056129923, `AssertionError` at line 58, on the commit before the removal) |
| `PathMatchedRateLimitAbsenceTest.theStarterIsShippedOff` | the starter's filter machinery stays out | passes before and after; it exists to keep that true |
| `PathMatchedRateLimitAbsenceTest.theOffSwitchIsLoadBearingRatherThanDecoration` | the flag defaults to on, so it cannot go with the block, and this fails if an upgrade changes that default | passes before and after; it is why the line stays |
| `PathMatchedRateLimitAbsenceTest.theApplicationIsConfiguredToReadTheForwardedClientAddress` | forwarding is on, which the issue assumed it was not | passes before and after; it records the correction |
| `PathMatchedRateLimitAbsenceTest.theAddressTheApplicationSeesIsWhateverTheNearestHopAsserts` | the address is asserted by a header rather than observed from the socket | passes before and after; it is the premise the decision rests on |

Only the first has a red to measure. The rest hold the reasoning, which is the part a later reader will not have.

The requirement to prove the ordinary path stays open is met by the shape of the change rather than by a new test: nothing is added in front of `/api/**`, the first test holds that, and the full suite exercises the API through the same chain it did before. The property is already covered for the mechanism that survives, by `InviteCodeRedemptionLimiterTest.anAcceptedCodeCostsNothing` and `ProjectInviteCodeServiceTest.validateAndUse_acceptedCode_costsNoAttempt`.

## OpenAPI

No change. Nothing here touches a controller, a DTO or a response.


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #715, opened 2026-09-06
Clear the remainder of the #687 sweep

### LINKED ISSUE #691: The leave-policy listing takes the organization from the path and checks only that it exists

### LINKED ISSUE #700: Remaining caller-supplied organization ids from the #687 sweep: a dead load, a check that cannot fire, and three ignored path segments
The remainder of the sweep for #687 siblings (fixed in #696). None of these leaks anything: the entity behind each is a `TenantScopedEntity` carrying `orgFilter`, so a foreign id yields an empty result. They are recorded because each one reads as though it were doing tenant work while the filter is what actually carries it, and every one that stays makes the next audit of this family slower.

The genuine findings from the same sweep are #697, #698 and #699. #691 is a separate entry in the same family and is not duplicated here.

## 1. A dead unfiltered organization load

`AttachmentService:339`:

```java
case "organization" -> attachment.setOrganization(organizationRepository.findById(entityId).orElse(null));
```

`entityId` is a caller-supplied `@PathVariable` and `Organization` is outside both tenant defences, so on the face of it this is the #687 shape. It is not, because all three write paths overwrite the field one line later with the tenant's own organization (`:117`, `:147`, `:439`, each `setOrganization(tenantEntityHelper.resolveCurrentOrganization())`). The load happens, its result is discarded, and `.orElse(null)` means the outcome does not even differ by whether the id exists.

Delete the line. The sibling `case` arms load `TenantScopedEntity` types by id and are fail-closed by `TenantIsolationLoadListener`, so this is the only arm that does not belong.

## 2. A duplicate-code check that can never fire

`LeavePolicyService:75`:

```java
policyRepository.existsByOrganizationIdAndLeaveTypeCode(dto.getOrganizationId(), ...)
```

The policy itself is written into `TenantContext.getCurrentOrgId()` at `:71`, so there is no cross-tenant effect. But `LeavePolicyCreationDto.organizationId` is a caller-supplied `@NotNull` field, and because `LeavePolicy` is filtered, any value other than the current tenant makes the predicate unsatisfiable and the uniqueness rule silently skipped. A client can create duplicate leave-type codes inside its own organization by sending an id that is not its own.

This is a correctness bug, not a security one. The field is part of the published DTO and clients send it today, so removing it is a breaking change; re-keying the check to `TenantContext` fixes the behaviour without touching the wire format, and is the safer of the two.

## 3. Path segments that are ignored or redundant

- `GET /api/v1/leave-requests/organizationId/{organizationId}`, `LeaveRequestController:124`. The segment is in the mapping and the handler binds no parameter for it at all; `getRequestsByOrganization()` reads `TenantContext`. Whatever a client puts there is ignored, which is worse for a caller than being refused. The web twin (`LeaveRequestControllerWeb:125`) is already `GET /organization` with no segment.
- `GET /api/v1/employee/web/managers/organizationId/{organizationId}`, `EmployeeControllerWeb:329`. Guard reads `TenantContext`, service reads `#organizationId` (`EmployeeService:494` into `EmployeeHierarchyService:141`). The classic mismatch, neutered by the filter on `Employee`. Foreign id yields an empty list.
- `GET /api/v1/leave-policies/organization/{organizationId}` is the same shape but has an existence oracle on top of it, so it is filed on its own as #691.

For each of these the decision is whether the route may lose the segment. The `echno-core` contract is hand-maintained with no codegen, so a route change breaks the web app at runtime rather than at compile time. Scoping the parameter to `TenantContext` and refusing a mismatch is the change that can be made now; retiring the segments belongs in a deliberate core release.

## Confirmed non-defects, recorded so the next sweep does not re-derive them

- `AttendanceService:227/369/618/672`, `ShiftTimingService:37`, `AttendanceSettingsService:56`, `MovementRecordService:92`, `AttendanceRegularizationService:99`, `ProjectService:164`, `LeavePolicyService:71`, `TenantEntityHelper:20` all read `TenantContext.getCurrentOrgId()` on the line above the lookup. They load the caller's own tenant root. Nothing to change.
- `OrganizationController:142/200` and `OrganizationWebController:146/169/201` bind the guard to the caller-supplied id (`isMember(#id)`, `hasAnyOrgRole(#id, ...)`, `hasOrgRole(#id,'system-admin')`). This is the correct pattern for the family.
- `OrganizationService.batchUpdateOrganization:419` already carries an explicit per-id `isMemberOrAdmin` loop with a comment explaining that `Organization` is guarded by neither the filter nor the load listener. It is the precedent worth copying.
- `EmployeeController:149` (`GET /employee/organization/{id}`) binds `@orgSecurity.isMember(#id)`, so it is the correctly-written member of the family.

abhijithanandan:
Closes #700.

The remainder of the #687 sweep. None of the three leaks anything, and each was still costing the next pass over this family the same work. Two of them are also wrong in ways a caller can see, and one of those is wrong differently from how the issue recorded it.

## 1. A dead unfiltered organization load

`AttachmentService.linkToEntity` loaded an `Organization` by the entity id in the request path, and all three write paths overwrote the field on the next line with `tenantEntityHelper.resolveCurrentOrganization()`. The load could not change the row, and `.orElse(null)` meant the outcome did not differ by whether the id existed either.

It still mattered, because `Organization` is the tenant root and carries neither the `orgFilter` nor the fail-closed load listener, so a caller-supplied id reaching an unfiltered `findById` is the shape of #687. Establishing that this one was inert is exactly the work the sweep did. The arm goes and `OrganizationRepository` goes with it, so there is nothing left in the service that can reach an organization by an id a caller chose.

Asserted on the constructor rather than on behaviour, because a load whose result is discarded has none. That is the whole reason the line went.

## 2. A uniqueness check that asked about the wrong row, twice

`LeavePolicyService.createPolicy` writes the policy into `TenantContext.getCurrentOrgId()` and checked for a duplicate against `dto.getOrganizationId()`. `LeavePolicy` carries the `orgFilter`, so any other value made the predicate unsatisfiable and the rule was skipped rather than applied.

Checking it turned up a second way past the same guard, not in the issue: the code is stored uppercased (`setLeaveTypeCode(dto.getLeaveTypeCode().toUpperCase())`) and the check ran against the casing as it arrived. So a lower-case code was checked against a value the table never holds, from the caller's own organization, with no foreign id involved at all.

**And a correction to what either one costs.** The issue says a client can create duplicate leave-type codes inside its own organization. It cannot. `uk_leave_policy_org_type` is a real unique constraint on `(organization_id, leave_type_code)`, declared on the entity and in the changelog, so the insert is refused whichever way the check was skipped. What is lost is the answer. Instead of the 409 the service was written to give, which names the code that clashed, `GlobalExceptionHandler` catches the `DataIntegrityViolationException` and answers a 409 reading "Database operation failed, Data Integrity violation", which names nothing. That is a smaller defect than recorded and still worth fixing: the check exists so the answer says which field was wrong.

Both arguments now carry the values the row will actually hold. `dto.organizationId` stays in the payload, because clients send it and removing it is a contract change rather than a repair.

## 3. Path segments that are ignored

Three were listed. **Two are already fixed on `development`**, by 50a7da2 ("Make the directory and leave-policy reads check the id they are given"), which landed the same evening #700 was filed and closed #691. `EmployeeControllerWeb` managers-by-organization and `LeavePolicyController` policies-by-organization both carry `@orgSecurity.isCurrentTenant(#organizationId)` beside the role check. Nothing to do; recorded rather than re-derived.

The third is live and is the worst of the three. `GET /api/v1/leave-requests/organizationId/{organizationId}` published the segment and **the handler bound no parameter for it at all**, so a caller who named another organization was handed their own leave requests as though the id had been honoured. Silently answering the wrong question is worse for a caller than a refusal, and the published contract makes a claim the handler does not keep.

It takes the same pair the other two took: `isCurrentTenant(#organizationId)` and the role check, which makes the segment binding rather than decorative. The service still reads `TenantContext`, which the guard has just established is the same organization.

**Retiring the segment is a separate decision and is not made here.** Measured while deciding: **no client calls any of the three routes.** `echno-core` and `echno-web` reach the `/web` twins, which take no segment. `echno-core` `leave-service.ts` `getOrganizationRequests()` hits `/leave-requests/web/organization` with no argument; `employee-service.ts` `getManagers()` hits `/employee/web/managers` with no argument; every policy call goes to `/leave-policies/web`. There is even a `managersByOrg` query key in `echno-core` that no hook or service uses, which suggests the org-scoped variant was planned and abandoned. So the segments could be retired cheaply, but the `echno-core` contract is hand-maintained with no code generation, so removing a published route breaks a client at runtime rather than at compile time. That belongs in a deliberate core release, not in a commit repairing a guard.

## The negative results

The sweep's confirmed non-defects move into `docs/MULTI_TENANCY_FILTER_GUIDE.md`, as a new section: the order to check a caller-supplied organization id in, the table of sites already cleared and why, the table of repairs and which to copy, and three traps. They are the expensive half of this work and an issue is not where they survive. The guide is the file someone reads before touching this family, which is where they will be looked for.

The middle trap is the one that generalises furthest, and it is the casing bypass above written as a rule rather than as an instance: **a check on the request as received is not a check on the row that will be written.** An organization id is only the most obvious argument that can be wrong. Any normalisation applied on the way in, case here but equally trimming or a canonical form, opens the same gap wherever the guard runs before it, and nothing about tenancy would surface it. That belongs somewhere a later reader will look, not in a test name.

## Tests

| Test | What it holds | Red before |
|---|---|---|
| `AttachmentOrganizationLinkTest.theServiceCannotReachAnOrganizationByACallerSuppliedId` | the repository the dead arm needed is gone, so the question cannot recur | **measured** (run 34056379146, `AssertionError` at line 37) |
| `LeavePolicyUniqueCodeCheckTest.aSecondPolicyForTheSameCodeIsRefused` | the ordinary duplicate is still refused, and refused by the service | passes before and after; it is the case the other two must not break |
| `LeavePolicyUniqueCodeCheckTest.namingSomeoneElsesOrganizationInTheBodyDoesNotSkipTheCheck` | the id in the body cannot switch the rule off | **measured** (same run, line 92) |
| `LeavePolicyUniqueCodeCheckTest.sendingTheCodeInLowerCaseDoesNotSkipTheCheckEither` | the check runs against the value the row will hold | **measured** (same run, line 100) |
| `LeaveRequestOrganizationSegmentTest.theListingStillAnswersForTheCallersOwnOrganization` | the ordinary read is not narrowed by the repair | passes before and after |
| `LeaveRequestOrganizationSegmentTest.namingAnotherOrganizationIsRefusedRatherThanIgnored` | the segment is refused rather than ignored, and the service is never reached | **measured** (same run, line 84) |

Four measured reds on the commit before the fixes, which is why the tests are their own commit.

## OpenAPI

One operation changed, `getOrganizationRequests_1`. Hand-applied rather than copied wholesale: the `parameters` block is taken byte for byte from `/api/v1/employee/web/managers/organizationId/{organizationId}`, which springdoc generated from an identical `@PathVariable Long organizationId`, and the summary, description and 403 text follow the annotations in this diff. CI's `openApiSnapshot` step is the check that it matches what the build serves.

### DIFF
--- a/docs/MULTI_TENANCY_FILTER_GUIDE.md
+++ b/docs/MULTI_TENANCY_FILTER_GUIDE.md
@@ -22,7 +22,8 @@ Data-level organization isolation using Hibernate filters that automatically app
 14. [Database Migrations](#database-migrations)
 15. [Entities Covered](#entities-covered)
 16. [Edge Cases](#edge-cases)
-17. [Troubleshooting](#troubleshooting)
+17. [Caller-Supplied Organization Ids: Already Checked](#caller-supplied-organization-ids-already-checked)
+18. [Troubleshooting](#troubleshooting)
 
 ---
 
@@ -682,6 +683,70 @@ User, Organization, Plan, Feature, PlanFeature, Subscription, SubscriptionItem, 
 | **Native SQL query** | Filter does NOT apply automatically. Add `AND organization_id = :orgId` manually |
 | **New entity without `@Filter` annotation** | Not filtered at all — visible to all orgs. Only add filter to org-scoped entities |
 | **Duplicate `@FilterDef` on entity** | **Application fails to start** with error: `Multiple '@FilterDef' annotations define a filter named 'orgFilter'`. Remove from entity — it should only exist in `package-info.java` |
+
+---
+
+## Caller-Supplied Organization Ids: Already Checked
+
+A handler that takes an organization id from the caller and looks something up with it is the
+shape worth auditing, because it can mean the caller chose which tenant's data was read. Most
+occurrences are not that. They are a redundant parameter beside a `TenantContext` read, or a
+parameter the entity's `orgFilter` renders inert.
+
+Telling the two apart is slow, and it is the same work every time. This section records the
+answers already arrived at, so a later pass can skip them and spend its time on what is new.
+Sources: issues #687, #691, #696, #697, #698, #699, #700, and the commits that closed them.
+
+### How to tell them apart, in order
+
+1. **Is the entity a `TenantScopedEntity`?** If it carries `orgFilter`, a foreign id yields an
+   empty result, and `TenantIsolationLoadListener` refuses a load by id outright. The parameter
+   then decides nothing, and the finding is tidiness rather than exposure.
+2. **Is it `Organization` itself?** The tenant root carries neither defence. A caller-supplied id
+   reaching an unfiltered `Organization` lookup is the dangerous form. `OrganizationService`
+   `batchUpdateOrganization` is the worked example: an explicit per-id `isMemberOrAdmin` loop,
+   with a comment saying why it has to be there.
+3. **Does the guard read the same id the service reads?** A guard on `TenantContext` beside a
+   service on `#organizationId` is a mismatch even when the filter neutralises it. The repair is
+   `@orgSecurity.isCurrentTenant(#organizationId)` alongside the role check, which makes the
+   segment binding instead of decorative.
+4. **Trace the whole workflow, not the call that refused.** A read gated a tier above the write
+   beside it strands the caller who is entitled to finish the job (#666).
+
+### Confirmed non-defects
+
+Recorded so they are not re-derived. Each was traced to the line above the lookup.
+
+| Site | Why it is fine |
+|------|----------------|
+| `AttendanceService` 227, 369, 618, 672 | reads `TenantContext.getCurrentOrgId()` on the line above; loads the caller's own tenant root |
+| `ShiftTimingService:37`, `AttendanceSettingsService:56` | same shape |
+| `MovementRecordService:92`, `AttendanceRegularizationService:99` | same shape |
+| `ProjectService:164`, `LeavePolicyService.createPolicy`, `TenantEntityHelper:20` | same shape |
+| `OrganizationController` 142, 200 | guard bound to the caller-supplied id (`isMember(#id)`, `hasAnyOrgRole(#id, ...)`); the correct pattern for the family |
+| `OrganizationWebController` 146, 169, 201 | same, including `hasOrgRole(#id,'system-admin')` |
+| `OrganizationService.batchUpdateOrganization:419` | explicit per-id `isMemberOrAdmin` loop, with the reason written down; the precedent to copy |
+| `EmployeeController:149` (`GET /employee/organization/{id}`) | binds `@orgSecurity.isMember(#id)`; the correctly-written member of the family |
+
+### Repaired, with the repair to copy
+
+| Site | What it was | What it became |
+|------|-------------|----------------|
+| `EmployeeControllerWeb` managers by organization | guard on `TenantContext`, service on `#organizationId` | `isCurrentTenant(#organizationId)` and the role check |
+| `LeavePolicyController` policies by organization (#691) | plus an existence check that read as a tenant check | same pair; the service now says why its `existsById` is not one |
+| `LeaveRequestController` requests by organization (#700) | segment published and bound to nothing at all | same pair; the segment now means what the contract says |
+| `LeavePolicyService.createPolicy` (#700) | uniqueness checked against `dto.organizationId` and the raw casing | checked against the tenant and the code as stored |
+| `AttachmentService.linkToEntity` (#700) | an `Organization` arm loading by a caller-supplied id, overwritten one line later | arm removed, and the repository with it |
+
+### Two traps worth keeping in mind
+
+- **An empty result is not proof of a defence.** It can equally be an id that matches nothing.
+  Check the entity for `orgFilter` rather than inferring from an empty list.
+- **Retiring a route or a path segment is a separate decision from repairing its guard.** The
+  `echno-core` contract is hand-maintained with no code generation, so a route change breaks the
+  web app at runtime rather than at compile time. Repair the guard now; retire the segment in a
+  deliberate release. As of #700, no client calls any of the three org-segmented routes above:
+  `echno-core` and `echno-web` both reach the `/web` twins, which take no segment.
 
 ---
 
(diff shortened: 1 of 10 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #720, opened 2026-09-06
Make the normalisation that decides a key locale-independent

### LINKED ISSUE #719: toUpperCase without a Locale in LeavePolicyService is locale-sensitive
`LeavePolicyService` normalises the leave-type code with `toUpperCase()` and no `Locale` argument. That uses the JVM's default locale, which is environment-dependent rather than a property of the data.

The classic case is Turkish: `"i".toUpperCase()` yields `"İ"` (dotted capital I) under `tr-TR` and `"I"` everywhere else. So the same code entered by the same person normalises differently depending on the locale the JVM happens to start in, and a container inheriting a different locale after a base-image change is enough to shift it.

**This is pre-existing and is not a regression.** It was noticed while re-keying the uniqueness check onto the values the row will actually hold (#700, PR #715). That change made the check and the store use the same call in the same JVM, so it introduces no new inconsistency and there was no reason to widen its scope. Filing separately rather than folding it in.

Why it is still worth fixing: the two calls agreeing within one JVM is not the same property as the stored value being stable across environments. A row written under one locale and matched under another would not agree, and the uniqueness constraint is the thing that would surface it, as a duplicate that should not have been possible.

The fix is `toUpperCase(Locale.ROOT)`. Worth a sweep for other unqualified `toUpperCase`/`toLowerCase` calls on values that get persisted or compared, rather than fixing only this one.

abhijithanandan:
Closes #719.

## What was wrong

`LeavePolicyService` uppercases the leave-type code with no `Locale`, so the casing the row is stored under is a property of the JVM the request reached rather than of the value. The sweep the issue asked for found the same shape in two more places that reach storage, and ten more that reach a comparison.

## The sweep, and what it is worth fixing

Twenty unqualified `toUpperCase()` / `toLowerCase()` calls in `src/main/java`. Sorted by whether the result is ever stored or matched against:

**Keys: stored, and then compared against the stored value (3)**

| Call | Why it matters |
|---|---|
| `LeavePolicyService:88` | The leave-type code. Written to the row and checked for uniqueness. The issue's own case. |
| `CategoryNormalizer:8` | `normalizedName`, persisted and used as the uniqueness key via `existsByNormalizedNameAndOrganization_Id`. Sharper than the leave-policy case: the lowering is followed by a `[^a-z0-9\s]` strip, and the dotless small i Turkish produces is outside that set, so it is deleted rather than folded. "MIX" normalises to "mx" and "Iron Fittings" to "ron fttngs". |
| `AttachmentOwner.folder():52` | The object-store folder a file is written under. Files written under one locale would be looked for under a prefix a different locale never produces. |

**Comparisons: a search term lowered in Java and matched against SQL `LOWER()` (10)**

`ExpenseService:142`, `SubContractService:183`, `TaskService:251`, `SearchService:96`, `IssueService:187`, `ReceiptService:155`, `EmployeeService:248`, `EmployeeService:275`, `AttendanceService:545`, `ProjectService:312`.

Nothing is stored, so no data is at risk, but the two sides of the comparison must fold the same way and only one of them is the JVM's. The database's `LOWER()` does not consult the JVM locale, so under `tr-TR` a search for "I" is sent as the dotless "ı" and matches nothing the database lowered to "i". `AssetService:505` already passes `Locale.ROOT` for exactly this, so this brings the rest in line with the convention already in the tree.

**Display and messages: left alone (4)**

`NotificationService:77` (a notification title), `AttachmentService:392`, `AttendanceRegularizationService:216` and `LeaveRequestService:467` (all three inside exception messages). These are read by a person and never compared or keyed on. `NotificationService:77` is the borderline one, since the title is persisted, but it is persisted as prose rather than as an identity, so nothing looks it up or matches against it and a differently cased first letter changes nothing but how the sentence reads.

Three inlined copies of the attachment folder rule in `AttachmentControllerWeb` now call `AttachmentOwner.folder()` instead, which is the same expression with the same documentation. Qualifying four copies of one rule separately and hoping they stay in step is how they would drift apart again.

## Tests

`LocaleIndependentNormalizationTest` sets the default locale to `tr-TR` and asserts on the two pure functions; `LeavePolicyUniqueCodeCheckTest` gains a case for the service. Pushed as their own commit ahead of the fix, so the red is measured rather than argued: the run on `289671f` failed those five, the run on the fix commit passes.

The ten search-pattern calls are private helpers with no seam to test through, and the failure mode is a query that returns nothing rather than a wrong value, so they are fixed on the reasoning above and not covered by a test of their own.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/attendance/AttendanceService.java
+++ b/src/main/java/org/tornotron/echno_backend/attendance/AttendanceService.java
@@ -40,6 +40,7 @@ import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.stream.Collectors;
 
 /**
@@ -542,7 +543,7 @@ public class AttendanceService {
                                                                Pageable pageable) {
         String searchPattern = (search == null || search.isBlank())
                 ? null
-                : "%" + search.toLowerCase() + "%";
+                : "%" + search.toLowerCase(Locale.ROOT) + "%";
         return attendanceRepository
                 .findWithFilters(projectId, date, status, searchPattern, pageable)
                 .map(attendance -> attendanceMapper.toResponseDto(attendance)).getContent();
--- a/src/main/java/org/tornotron/echno_backend/category/CategoryNormalizer.java
+++ b/src/main/java/org/tornotron/echno_backend/category/CategoryNormalizer.java
@@ -1,11 +1,17 @@
 package org.tornotron.echno_backend.category;
 
+import java.util.Locale;
+
 public class CategoryNormalizer {
 
     public static String normalize(String input) {
         if (input == null) return null;
 
-        String normalized = input.toLowerCase().trim();
+        // Locale.ROOT, because this value is the uniqueness key and the strip below is
+        // ASCII-only. Under tr-TR the default lowering turns I into a dotless small i,
+        // which [^a-z0-9\s] then deletes rather than folds, so "MIX" would normalise to
+        // "mx" on one host and "mix" on another. See #719.
+        String normalized = input.toLowerCase(Locale.ROOT).trim();
 
         normalized = normalized.replace("&", "and");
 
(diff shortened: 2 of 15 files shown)


abhijithanandan:
@coderabbitai review

PR #721, opened 2026-09-06
Make the leave-balance recalculate endpoint recalculate

### LINKED ISSUE #708: The leave-balance recalculate endpoint does not recalculate
Found while fixing #689, and deliberately left out of that PR because it is a functional defect rather than an authorization one.

## What it does

`POST /api/v1/leave-balances/employee/{employeeId}/recalculate` (`LeaveBalanceController:111`) and its twin `POST /api/v1/leave-balances/web/recalculate` (`LeaveBalanceControllerWeb:107`) both call:

```java
return ResponseEntity.ok(balanceService.getAllBalancesForEmployee(employeeId, targetYear));
```

That is the same method `GET /leave-balances/employee/{employeeId}` calls. So the POST is the GET, wearing a different verb and a summary that promises a recomputation.

## Why it half works, which is the confusing part

`getAllBalancesForEmployee` delegates to `getOrCalculateBalance`, which does initialize a missing row and does call `leaveAccrualService.recalculate(balance)` when `needsRecalculation(balance, year)` is true. So the endpoint is not inert: it recomputes accrual when a month has passed since the row was last touched. What it does not do is force anything. The Operation description says "Recomputes every policy balance held by the employee for the given year", and a caller reaching for this endpoint is reaching for it precisely because they believe the stored figures are wrong and want them rebuilt, which is the one case `needsRecalculation` returns false for.

## There is no bulk recalculate to call

`LeaveBalanceService.recalculateBalance(employeeId, policyId, year)` (`:192`) exists and does force it, but it is per-policy and nothing calls it. There is no `recalculateBalances(employeeId, year)` in the service at all, so the controller could not have called the right thing.

## Consequence

Low, and worth saying so. The endpoint has no caller: `useRecalculateBalances` is defined in `echno-core` and in `echno-web`s own copy of the leave service, and a search across `echno-web` for any button, menu item or handler that invokes it finds none. So this is a route that would mislead whoever wired it up next rather than one misleading anybody now.

## Fix direction

Add `recalculateBalances(employeeId, year)` to `LeaveBalanceService`, iterating the applicable policies and calling the existing per-policy `recalculateBalance` for each, and point both twins at it. Keep the role gate: #689 left this route with the administrators on the grounds that it is written as a command, and forcing a rebuild of somebody's leave figures is an administrative act whether or not it currently does anything.

abhijithanandan:
Closes #708.

Build it rather than delete it. The reasoning is below, because "delete the route" was the other honest answer and the issue is right that nothing calls it today.

## Can a stored balance drift from its inputs

Yes, and by construction rather than by accident. `used` and `pending` are maintained two independent ways at once.

They are **derived** in `LeaveAccrualService.recalculate`, which rebuilds them from `sumApprovedDaysByEmployeePolicyYear` and `sumPendingDaysByEmployeePolicyYear`, and they are **incremented in place** by four helpers on the write paths: `LeaveRequestService.updatePendingBalance` and `restoreUsedBalance`, `LeaveApprovalService.transferPendingToUsed` and `restorePendingBalance`. Three things make the incremental half disagree with the derived half:

1. **All four are `.ifPresent(...)`.** A balance row is created lazily, on first read. Until somebody looks at an employee's balances, the row does not exist and every increment against it is a silent no-op, while the request rows the derived sum reads are written regardless.
2. **All four clamp with `Math.max(0, ...)`.** A decrement that lands on a row whose figure was never incremented is swallowed rather than going negative, so the row cannot even carry evidence that it is out of step.
3. **`needsRecalculation` stops reconciling a past year entirely.** It returns false as soon as `now.getYear() != year`, so once a year turns, the read path never rebuilds that year's balances again. A leave request for last December approved this January moves the incremental figures and nothing recomputes them, for the rest of the row's life.

So drift is not hypothetical: (3) in particular guarantees a class of it that no other code path repairs, and (1) and (2) mean the incremental figures are only as good as the order the row happened to be created in.

## What a forced recalculate does and does not repair

Worth being exact, because the endpoint's current description over-promises and this change should not repeat that.

It **does** rebuild `used` and `pending` from the request rows, fold the ADJUSTMENT ledger back in, and post any monthly ACCRUAL transaction the balance is missing.

It does **not** rewrite an ACCRUAL transaction already posted for a month, so a policy quota edited mid-year, or attendance backfilled after the fact, leaves the months already accrued as they were. It also does not recompute `openingBalance` or `carryForwardFromPrevious`, which are written once when the row is created. Both descriptions now say so.

## The change

`LeaveBalanceService.recalculateBalances(employeeId, year)` iterates the policies applicable to the employee and calls the existing per-policy `recalculateBalance` for each, which is the shape `getAllBalancesForEmployee` already uses over `getOrCalculateBalance`. Both twins point at it.

One constraint comes with forcing. `recalculateBalance` has no pre-joining-year guard, because the only caller until now passed a policy id and a year the caller had chosen together. Iterating every policy for an arbitrary year would have made a bulk recalculate of 2019 create rows for an employee who joined in 2024, which is exactly what the read path refuses to do. The bulk method carries the same `isBeforeJoiningYear` check the read path uses and answers with the transient zero balance.

The role gate is unchanged: #689 left this route with the administrators on the grounds that it is written as a command, and that reading is now true rather than aspirational.

## Contract

No route added or removed, so nothing to check in `echno-core` or `echno-web`. For the record, the callers are where the issue said: `useRecalculateBalances` is declared in `echno-core` (`src/hooks/leave/use-leave-mutations.ts:319`) and in `echno-web`'s own copy (`hooks/leave/use-leave-mutations.ts:258`), and `recalculateBalances` in both leave services, but no component, button or handler in `echno-web` invokes the hook. That is what made deleting the route thinkable; it is also why deleting it would have been the wrong trade, since the working implementation is a dozen lines and the drift above is real.

`docs/openapi.json` changes only in the two `description` strings.

## Tests

`LeaveBalanceRecalculationTest` builds a balance already recomputed this month, so the read path's `needsRecalculation` guard skips it, and asserts both twins rebuild it anyway. Pushed as its own commit ahead of the fix: the two forcing cases failed on that commit and pass on the fix.

Three further cases pass on both commits and are regression guards rather than measurements: one pins the read path's own skip as the baseline the endpoint has to differ from, and two pin the pre-joining-year rule that the naive version of this fix would have broken.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -63185,7 +63185,7 @@
     },
     "/api/v1/leave-balances/employee/{employeeId}/recalculate": {
       "post": {
-        "description": "Recomputes every policy balance held by the employee for the given year and returns the refreshed list, defaulting to the current year when none is given.",
+        "description": "Rebuilds every policy balance held by the employee for the given year and returns the result, defaulting to the current year when none is given. Unlike a read, which recomputes only when a month has turned since the balance was last touched, this recomputes unconditionally: used and pending days are rebuilt from the leave requests, manual adjustments are folded back in from the ledger, and any monthly accrual the balance is missing is posted. A monthly accrual already recorded is taken as it stands, and the opening balance carried forward from the previous year is not recomputed.",
         "operationId": "recalculateBalances_1",
         "parameters": [
           {
@@ -63785,7 +63785,7 @@
     },
     "/api/v1/leave-balances/web/recalculate": {
       "post": {
-        "description": "Recomputes every policy balance held by the employee for the given year and returns the refreshed list, defaulting to the current year when none is given.",
+        "description": "Rebuilds every policy balance held by the employee for the given year and returns the result, defaulting to the current year when none is given. Unlike a read, which recomputes only when a month has turned since the balance was last touched, this recomputes unconditionally: used and pending days are rebuilt from the leave requests, manual adjustments are folded back in from the ledger, and any monthly accrual the balance is missing is posted. A monthly accrual already recorded is taken as it stands, and the opening balance carried forward from the previous year is not recomputed.",
         "operationId": "recalculateBalances",
         "parameters": [
           {
--- a/src/main/java/org/tornotron/echno_backend/leave/LeaveBalanceController.java
+++ b/src/main/java/org/tornotron/echno_backend/leave/LeaveBalanceController.java
@@ -98,8 +98,14 @@ public class LeaveBalanceController {
     @PreAuthorize("@orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin')")
     @Operation(
             summary = "Recalculate an employee's leave balances",
-            description = "Recomputes every policy balance held by the employee for the given year and "
-                    + "returns the refreshed list, defaulting to the current year when none is given."
+            description = "Rebuilds every policy balance held by the employee for the given year and "
+                    + "returns the result, defaulting to the current year when none is given. Unlike a "
+                    + "read, which recomputes only when a month has turned since the balance was last "
+                    + "touched, this recomputes unconditionally: used and pending days are rebuilt from the "
+                    + "leave requests, manual adjustments are folded back in from the ledger, and any "
+                    + "monthly accrual the balance is missing is posted. A monthly accrual already recorded "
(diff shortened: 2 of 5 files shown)


abhijithanandan:
@coderabbitai review

Issue #707, opened 2026-09-06
The openapi-served artifact is unreachable in exactly the runs that need it


abhijithanandan:
Regenerating `docs/openapi.json` needs a container runtime, which the workstation does not have and the lab nodes are not always reachable for. The documented fallback is to take the document from the build's own `openapi-served` artifact, which `OpenApiSnapshotTest` writes from the identical canonical string the update flag writes.

That fallback does not work when you actually need it, for two reasons together:

1. The artifact uploads only on `if: failure()`.
2. The `Test` step runs **before** `openApiSnapshot`.

So a run that fails on `Test` never reaches the step that produces the artifact. And a run where the snapshot is stale is, by definition, a run where something changed, which is frequently a run where a test is also red. The artifact is available in the cases where it is not needed and missing in the cases where it is.

The workaround one of these changes needed was to push a throwaway probe PR with an annotation temporarily removed, so that `Test` would pass and `openApiSnapshot` would fail into an artifact, then restore the annotation and hand-patch the document. That worked and CI verified the result byte for byte, but it is not a procedure anyone should have to rediscover, and it puts a deliberately broken commit on a branch to get a build artifact out.

Two candidate fixes, either of which closes it:

- Upload the artifact with `if: always()` so a green run publishes the canonical document too. Cheapest, and it also gives anyone a current copy without running anything.
- Run `openApiSnapshot` before `Test`, or in its own job, so the two do not gate each other.

Worth doing the first regardless. The document is generated and CI verifies it byte for byte, so publishing it on every run costs an upload and removes a whole class of local-environment dependency.

Raised from the work on #694.

PR #722, opened 2026-09-06
Let the project attendance listing be narrowed by approval

### LINKED ISSUE #712: Let the attendance list be narrowed to an approval status
Split out of tornotron/echno-web#402, which is otherwise closed. This is the one point of it that never had a backend to build against.

`GET /attendance/project/{projectId}` (and its `/web` twin) takes a required `date` and offers no way to narrow by `approvalStatus`. The attendance list page wants "show me the pending ones on this site", and today it cannot ask for that.

Filtering client-side is not a substitute. The result is server-paged, so a client-side filter over the current page shows three rows under a footer reading "1-20 of 137", which is a worse answer than no filter at all.

What is wanted is an optional `approvalStatus` query parameter on the project listing, matching how the rest of the list params behave: absent means unfiltered, present narrows the page the server builds so the count and the paging stay honest.

Note this is unrelated to `GET /attendance/pending-approvals` from #701. That endpoint is the held days routed to the caller, a strict subset filtered on `requiresGeofenceApproval`, and it is a separate surface rather than a filter on the list. Every attendance record is created at `approvalStatus = pending` and stays there until somebody decides, so a plain pending filter on the project listing is a much wider set and is exactly what the list page is asking for.

Once this lands, the web side is a small change to `app/users/dashboard/attendance/page.tsx` and its filter chips.

abhijithanandan:
Closes #712. Comes out of tornotron/echno-web#402.

## What the useful parameter actually is

`approvalStatus` alone is not it, and shipping only that would have been a filter that looks like it works and filters almost nothing.

`checkIn` creates every attendance record at `approvalStatus = PENDING` and nothing moves it until somebody decides, so on an ordinary day the pending set on a site is the site's whole roll. Filtering the list on PENDING would return the page it was already showing, under a total that had barely moved. The web page shows this from the other side: it draws the approve and reject buttons on every row where `attendance.approvalStatus === 'pending'`, which today is every row.

The set a person means by "show me the pending ones on this site" is the days somebody actually has to look at, and those are the ones marked `requiresGeofenceApproval`: a punch fell outside the site and the day is held. So the listing gains two optional parameters:

- **`requiresApproval`** (boolean) is the narrow one, and the one the filter chip should use. On the seeded day in the test it selects 2 rows of 9 where `approvalStatus=PENDING` selects 8.
- **`approvalStatus`** (enum) is still worth having, because it is the only way to ask for the decided days. `APPROVED` and `REJECTED` are genuinely small sets and a reviewer wants "what did we decide here today". It is PENDING that is the near-no-op, and the operation description now says so rather than leaving the next person to find out.

They are two orthogonal columns, so they stay two orthogonal parameters rather than being folded into one four-valued enum, which would have made "held and already approved" unaskable. Asked together, `approvalStatus=PENDING&requiresApproval=true` is the site's outstanding work for the day.

## Not the queue

This does not overlap `GET /attendance/pending-approvals` from #701 and does not replace it. That endpoint is routed to the signed-in caller, excludes their own days, and is capped rather than paged. This is the project list: one site, one date, paged, and it shows every held day there whoever has to decide it. The two answer different questions and the distinction that made the queue correct is the distinction this change is built on.

## Shape of the change

The filters go into the `findWithFilters` JPQL beside the existing `status` and `search` predicates, in the same `:param IS NULL OR ...` form, so absent means unfiltered exactly as the neighbours do. The narrowing has to be in the query rather than over the returned page, or the total the client pages against stays the unfiltered one.

## Tests

`AttendanceProjectApprovalFilterIT`, a `@DataJpaTest` against the real CockroachDB the queue IT already uses. The seed is built so no assertion passes by accident: nine rows on the site that day, eight of them pending, three held, two both, one already decided, plus a same-shaped row on the next day and another site that must never appear. A filter that had stopped narrowing returns eight where the test wants two.

Two of the cases are the point rather than coverage. `filteringOnPendingIsAlmostTheWholeDay` pins the near-no-op as a measured fact, so nobody later mistakes it for a bug and "fixes" the pending filter into meaning the held set. `theNarrowingHappensInTheQuerySoTheTotalIsHonest` asks for a one-row page and reads `getTotalElements`, which is the assertion a filter applied after paging would fail.

**Measured red:** `docs/openapi.json` was pushed first, on its own commit, declaring the two parameters before the endpoint offered them. CI on `0b6b48e` passed `Test` and failed the `OpenAPI document up to date` step, which is the contract check reporting that the served document has no such parameters. The implementation commit turns it green. Worth noting because that step only produces its `openapi-served` artifact when the run reaches it, and a run failing on `Test` never does (#707).

## What is needed on the client side

Nothing is broken by this, because both parameters are optional and absent behaves exactly as today. To use them, `echno-core` needs two fields on `AttendanceListParams` and two lines in `attendanceListParamsToQuery` (`src/types/attendance/attendance-list-params.ts`), which today emits only `date`, `status`, `search`, `page` and `size`. Then the filter chips on `app/users/dashboard/attendance/page.tsx`, as the issue describes.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -25454,7 +25454,7 @@
     },
     "/api/v1/attendance/project/{projectId}": {
       "get": {
-        "description": "Returns a page of attendance records for a project on a given date, optionally filtered by status or a search term against the employee name, sorted by employee name.",
+        "description": "Returns a page of attendance records for a project on a given date, optionally filtered by status, by approval status, by whether the day is held for a geofence decision, or by a search term against the employee name, sorted by employee name. The two approval filters answer different questions: every record is created pending and stays there until somebody decides, so approvalStatus=PENDING is close to the whole day's roll, while requiresApproval=true is the small set of days actually waiting on a decision because a punch fell outside the site.",
         "operationId": "getByProject_1",
         "parameters": [
           {
@@ -25493,6 +25493,27 @@
                 "PENDING_REGULARIZATION"
               ],
               "type": "string"
+            }
+          },
+          {
+            "in": "query",
+            "name": "approvalStatus",
+            "required": false,
+            "schema": {
+              "enum": [
+                "PENDING",
+                "APPROVED",
+                "REJECTED"
+              ],
+              "type": "string"
+            }
+          },
+          {
+            "in": "query",
+            "name": "requiresApproval",
+            "required": false,
+            "schema": {
+              "type": "boolean"
             }
           },
           {
@@ -26488,7 +26509,7 @@
     },
     "/api/v1/attendance/web/project/{projectId}": {
       "get": {
-        "description": "Returns a page of attendance records for a project on a given date, optionally filtered by status or a search term against the employee name, sorted by employee name.",
+        "description": "Returns a page of attendance records for a project on a given date, optionally filtered by status, by approval status, by whether the day is held for a geofence decision, or by a search term against the employee name, sorted by employee name. The two approval filters answer different questions: every record is created pending and stays there until somebody decides, so approvalStatus=PENDING is close to the whole day's roll, while requiresApproval=true is the small set of days actually waiting on a decision because a punch fell outside the site.",
         "operationId": "getByProject",
         "parameters": [
           {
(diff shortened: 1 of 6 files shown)


abhijithanandan:
@coderabbitai review

PR #724, opened 2026-09-07
Sweep the HQL joins the tenant filter does not scope, and fix the duplicate-code check it hid

### LINKED ISSUE #718: The tenant filter does not scope an HQL join, and the leave-policy duplicate-code check is skipped because of it
Split out of #698, which repaired the guard. The guard was the smaller half. Two things were measured against a real database while fixing it, and neither is closed by that fix.

## The tenant filter does not scope an entity joined explicitly in HQL

The general fact, and the one worth carrying beyond this endpoint. `@Filter` narrows a query root and a filtered collection. It does **not** narrow `JOIN o.employees e`. `OrganizationRepository.findByIdAndUserEmail` therefore resolves any organization the caller holds an `Employee` row in, whatever tenant the request is scoped to.

The fail-closed load listener does not catch it either, and cannot: the joined `Employee` is only joined, never selected, so there is no post-load for the listener to judge. Neither mechanism scopes that query. `OrganizationLookupUnderTheOrgFilterIT` pins it, with an unfiltered control beside it showing the same query resolving either organization when the filter is off.

Anywhere a query joins a tenant-scoped entity without selecting it, and the result is not itself tenant-scoped, the filter is not doing the scoping and nothing else is either. A sweep for that shape has not been done and is the main work this issue is asking for.

> Worth recording because it cost three CI cycles. Two earlier readings of this measurement were wrong in opposite directions. The test asserted the lookup came back empty and was told a `TenantAccessDeniedException` came out of it, which reads as the listener refusing the row. It was not. `Organization` is a Lombok `@Data` entity whose generated `toString` walks its lazy `employees` collection, so AssertJ building the failure description for a *present* `Optional` initialized that collection, loaded the foreign organization's employees under the current tenant, and tripped the listener there. The assertion's own error message was raising the exception that was being read as the result. Splitting the chain one property per test exposed it: the first link, "anything is thrown at all", was already false. Any assertion whose failure message stringifies a `@Data` entity with a lazy collection has this hazard.

## The duplicate-code check is skipped on the path that needs it

`LeavePolicyService.duplicatePolicy` calls `existsByOrganizationIdAndLeaveTypeCode(targetOrganizationId, ...)` against `LeavePolicy`, which **does** carry `orgFilter` as a query root. So for any target other than the current tenant the filter's predicate and the argument name different organizations, the query can never match, and the uniqueness rule is silently skipped on exactly the path that needs it. A real collision reaches `uk_leave_policy_org_type` and surfaces as a 500 rather than the 409 the method means to raise.

This is live, not hypothetical: the write it guards succeeds, now that the target is known to resolve.

Naming the current tenant as the target instead makes that same check match the source policy's own leave-type code, so it answers 409 always. That is why the narrower option floated in #698, keeping `targetOrganizationId` but requiring it to be the current tenant, is not available: it would leave an endpoint that refuses every input.

## The decision this needs

Whether duplicating a policy into another organization is a supported feature.

- **If yes**, the duplicate-code check has to become a deliberate cross-tenant read rather than a query that quietly matches nothing, and the guard repaired in #698 is load-bearing.
- **If no**, drop `targetOrganizationId` and give the endpoint a purpose it can serve within one tenant, or remove it. Removing it is a change to the published contract, so `echno-core` and any caller in `echno-web` have to be checked first.

The working assumption should be that it is supported, since the path works today and somebody may be using it.

abhijithanandan:
Closes #718.

## The defect class

The `orgFilter` narrows a query **root** and a filtered collection. It does not narrow an entity joined explicitly in HQL. `TenantIsolationLoadListener` cannot cover the gap either, and not by oversight: it runs on a post-load, and an entity that is only joined is never loaded, so there is no event to judge. Where both hold at once nothing scopes that part of the query, and unlike the other gaps in this area it fails quietly, with a plausible answer about the wrong tenant rather than an empty result.

`OrganizationLookupUnderTheOrgFilterIT` measured that in #698. This is the sweep that was owed.

## The sweep

**All 151 `@Query` annotations read. 41 explicit join clauses across 25 of them**, counting `countQuery` beside `value`. Also checked and carrying no join at all: the five `Specification` classes, every use of the Criteria API, and the three `EntityManager.createQuery` calls in `VendorSummaryService`, which reach related rows through path expressions off a filtered root.

**Two were genuinely exposed**, both on `OrganizationRepository` and both already known from #698. Nothing new was found. Everything else is clear on one of three structural facts, and the reason for each is now written into the guide, which is the expensive half and the part a later sweep would otherwise re-derive:

- a **fetch join** loads the rows, so the listener refuses a foreign one (fail-closed);
- an **association join off a tenant-scoped root** walks a foreign key out of a row already in the tenant, so it can narrow the result but not widen it;
- an **entity join** (`JOIN MaterialLocationThreshold t ON ...`) inherits nothing and is safe only if the `ON` clause carries the tenant predicate itself, which `LowStockRepository` does and is the pattern to copy.

The count is deliberately not inflated. Neither of the two exposed sites is new, and one of them (`findAllByUserEmail`, the organization switcher) is cross-tenant on purpose.

## The structural fix

`TenantScopedJoinTest` scans every `@Query` and requires each explicit join to be clear by one of the three facts above, or registered in `CROSS_TENANT_JOINS` with a written reason. Any join in a native query is always registered, since nothing narrows a native query. It carries the staleness check `UnboundedRepositoryReadTest` has, so the registry can only shrink, and a third check refuses a bare entry with no reason.

Five entries today, matching the sweep exactly. This is worth more than five guards: the next join of this shape is answered for when it is written, by whoever writes it, in the place a reviewer looks.

## The second item: the duplicate-code check

The same blind spot from the other side. `LeavePolicyService.duplicatePolicy` asked `existsByOrganizationIdAndLeaveTypeCode(targetOrganizationId, code)` against `LeavePolicy`, whose root **does** carry the filter. Hibernate added `organization_id = <tenant>` beside the caller's `organization_id = <target>`, the two named different organizations, and the query matched nothing whatever the table held. The uniqueness rule was skipped on the one path that needs it, and a real collision reached `uk_leave_policy_org_type` as a 500.

Naming the current tenant instead is not the smaller repair it looks like: the check would then match the source policy's own leave-type code every time and the endpoint would answer 409 to every input.

The question genuinely is about another organization, so it is now asked with `countWithLeaveTypeCodeInOrganizationUnfiltered`: native so nothing narrows it, carrying `organization_id` itself rather than relying on a filter that will not be applied, returning a count and never a row, and reached only after `@orgSecurity.hasAnyOrgRole(#targetOrganizationId, ...)` has established the caller's role in the target. The compliance dispatcher's cross-tenant reads are written the same way and were the precedent.

The 409 is documented on both twins now that it is reachable, so `docs/openapi.json` moves with it.

## Tests

| Test | What it holds | Measured? |
|---|---|---|
| `LeavePolicyDuplicateAcrossOrganizationsIT.aCodeTheTargetAlreadyHoldsIsRefusedRatherThanLeftToTheUniqueConstraint` | a collision in the target is a `DuplicateResourceException`, not a constraint violation | **Measured red** on the first commit, run 34117399559, the only failure in the suite |
| `theFilteredExistenceCheckCannotSeeTheRowThatCollides` | the old check answers false while the row is in the table: the defect, asserted directly | Passes before and after; it is the measurement, not the fix |
| `withNoFilterEnabledTheSameCheckFindsIt` | the control: the collision is real and only the filter was hiding it | Passes |
| `theUnfilteredProbeAnswersForTheTargetWhileTheTenantIsSomewhereElse` | the new probe sees the target's row under a foreign tenant | Green after the fix |
| `theUnfilteredProbeStillDistinguishesOneOrganizationFromAnother` | unfiltered is not unscoped: the predicate is in the query | Green after the fix |
| `aCodeTheTargetDoesNotHoldStillDuplicates` | the endpoint is not made to refuse everything | Passes before and after |
| `TenantScopedJoinTest` (three rules) | no undeclared unscoped join; no stale registry entry; no reasonless entry | Green |

The first commit is the measurement, pushed ahead of the fix so the red was read rather than reasoned about. A mocked repository could not have shown any of it: the service calls the same method either way and a stub answers the question the filter would have refused.

## Guide

`docs/MULTI_TENANCY_FILTER_GUIDE.md` gains a section for the sweep: what each mechanism actually reaches, the three shapes and which are safe, the cleared list with the reason for each site, the registered list with why, the filtered-root-asked-about-another-organization case, and the AssertJ trap that cost #698 three CI cycles. Two claims already in that document were wider than the mechanism and are corrected: that a JPQL `@Query` is filtered, and that the filter reaches every query on a tenant-scoped entity.

### DIFF
--- a/docs/MULTI_TENANCY_FILTER_GUIDE.md
+++ b/docs/MULTI_TENANCY_FILTER_GUIDE.md
@@ -1,6 +1,8 @@
 # Multi-Tenancy via Hibernate @Filter — Implementation Guide
 
-Data-level organization isolation using Hibernate filters that automatically append `WHERE organization_id = :organizationId` to every query on tenant-scoped entities.
+Data-level organization isolation using Hibernate filters that append `WHERE organization_id = :organizationId` to a query whose **root** is a tenant-scoped entity, backed by a fail-closed check at the load boundary for the primary-key loads the filter never sees.
+
+Read [Explicit Joins](#explicit-joins-what-the-filter-does-not-reach) before assuming a query is covered. The filter reaches a query root and a filtered collection; an entity joined explicitly in HQL is scoped by neither mechanism, and that failure is silent.
 
 ---
 
@@ -23,7 +25,8 @@ Data-level organization isolation using Hibernate filters that automatically app
 15. [Entities Covered](#entities-covered)
 16. [Edge Cases](#edge-cases)
 17. [Caller-Supplied Organization Ids: Already Checked](#caller-supplied-organization-ids-already-checked)
-18. [Troubleshooting](#troubleshooting)
+18. [Explicit Joins: What the Filter Does Not Reach](#explicit-joins-what-the-filter-does-not-reach)
+19. [Troubleshooting](#troubleshooting)
 
 ---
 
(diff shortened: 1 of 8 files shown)


abhijithanandan:
@coderabbitai review

Issue #716, opened 2026-09-06
Six phantom guards left on EmployeeController refuse every caller


abhijithanandan:
Split out of #710, which repaired one of the seven guards on the mobile `EmployeeController`. The other six are the same phantom-authority shape as #641 and #684 and are all still live.

## Why the authority cannot be issued

`JwtAuthConverter` mints a bare `resource:scope` authority in exactly one place, `extractPermissions`, which reads the `authorization` claim of an RPT. `employee:create`, `employee:read`, `employee:update`, `employee:delete` and `employee:admin` therefore need a Keycloak Authorization Services resource named `employee` carrying the matching scope. The multi-tenancy audit of 2026-08-18 traced the identical `organization:admin` string end to end against the live staging Keycloak and found zero authorization scopes realm-wide, one scopeless `Default Resource` and one scopeless `Default Permission`. A scopeless permission yields no `resource:scope` authority at all. #641 confirmed the same for `billing:admin`, #684 for `leave:read` and `leave:admin`.

## The six

| method | current guard |
|---|---|
| `POST /` (create) | `employee:create` or `employee:admin` |
| `GET /` (list all) | `employee:read` or `employee:admin` |
| `GET /organization/{id}` | `(employee:read and @orgSecurity.isMember(#id))` or `employee:admin` |
| `PATCH /{id}` | `employee:update` or `employee:admin` |
| `PATCH /batch` | `employee:update` or `employee:admin` |
| `DELETE /{id}` | `employee:delete` or `employee:admin` |

`GET /{id}` was repaired in #710 and is not in this list.

## Why they were not repaired with it

Four of the six are writes. Repairing a phantom write guard moves it from "refuses everybody" to "allows the role", which is a widening even though nothing regresses, and #710 was scoped to a read asymmetry. `GET /organization/{id}` also has a second question underneath: it already pairs the phantom authority with a real `isMember(#id)` check, so repairing it is not a one-line substitution.

The obvious shape is to mirror the web twin, which gates create and delete on `system-admin, hr-admin`, the paginated and full listings on `system-admin, hr-admin, project-manager`, and the patch on `isSelfOrHasAnyOrgRole(#id, 'system-admin', 'hr-admin')`. Worth confirming that the mobile client still calls these before repairing rather than deleting: if it does not, deleting the endpoints is the smaller change.

abhijithanandan:
Review on #713 landed on the same two reads and proposed a concrete shape for one of them, worth recording here:

[...]

PR #726, opened 2026-09-07
Make org role holders members, and sweep the 36 guards whose role clause is dead

### LINKED ISSUE #717: The same dead or-clause stands on 36 more read guards across ten controllers
Split out of #709, which removed the clause from the seven read guards on `ProjectControllerWeb`. The identical expression stands unchanged on 36 more.

## The clause and why it decides nothing

```java
@PreAuthorize("@orgSecurity.isMemberOfCurrentTenant() or @orgSecurity.hasAnyOrgRoleForCurrentTenant(...)")
```

Both sides open on `TenantContext.getCurrentOrgId()` and return false when it is null. `TenantFilter` sets it in exactly two branches, and both confirm an `ORG_MEMBER_{orgId}` authority first: the `X-Organization-Id` branch refuses with 403 without it, and the inference branch reads the id out of the `ORG_MEMBER_` authorities themselves. A caller holding `ORG_7_ROLE_system-admin` and no `ORG_MEMBER_7` therefore reaches the controller with no organization in force and fails both clauses. The role clause cannot change the outcome on any request that reaches a controller, so these guards are membership guards wearing a two-part expression.

The global-admin branch is the same story from the other side: it sets the bypass flag and no organization id, so both clauses are false for a global admin too.

`TenantFilterTest.theRoleGuardCannotSucceedWhereTheMembershipGuardFails` pins the invariant against the real `OrganizationSecurityService` inside the filter chain.

## Where they are

| controller | occurrences |
|---|---|
| `AssetControllerWeb` | 7 |
| `IssueControllerWeb` | 6 |
| `TaskControllerWeb` | 4 |
| `StockAdjustmentControllerWeb` | 4 |
| `SubContractControllerWeb` | 3 |
| `ReceiptControllerWeb` | 3 |
| `IssueCommentControllerWeb` | 3 |
| `ExpenseControllerWeb` | 3 |
| `CategoryControllerWeb` | 2 |
| `SearchController` | 1 |

## What to do with them

Removing the clause is behaviour-preserving given the invariant above, and makes each guard say what it decides. It is a wide mechanical diff, which is why it is a tracked sweep rather than something folded into #709: the value is that the next reader of any of these does not believe a role is enforced where it is not.

Worth doing in one pass so the ten controllers stay consistent with `ProjectControllerWeb`, and worth doing only after deciding the question underneath, which is separate: should a caller holding an org role without the flat membership group be able to resolve a tenant at all? Today they cannot, so `eec6c37`'s intent, letting the bootstrap admin read what they may already write, was never achieved. That is a change to `TenantFilter` and to the security model, not to a controller, and if it is ever made these clauses become live and should come back rather than be removed now and rewritten later. Sequence the sweep after that decision.

abhijithanandan:
## What #717 turned out to be

#717 asked whether a caller holding an org role without the flat membership group should be able to resolve a tenant, and offered relaxing `TenantFilter` as the way to make its 36 dead `or` clauses live.

Tracing it first: the org id is baked into the role authority itself. `JwtAuthConverter` derives `ORG_5_ROLE_system-admin` from the Keycloak subgroup `/org-5/system-admin`, so a role always names its own organization and can never speak for another. The only state those 36 clauses can describe is a user in `/org-5/system-admin` who is not in `/org-5`.

That state is not a design requirement. It is a split the code produced in one place and tolerated in another:

- `assignOrgRole` joined the role subgroup only. It resolves the employee against the tenant first, so membership existed in practice, but nothing enforced it.
- `removeUserFromOrganization` left the parent group and stopped. Keycloak keeps membership of `/org-5` and of `/org-5/system-admin` as independent rows, so **an offboarded employee kept their role authority and lost only their membership authority**. Employee deletion is the reachable route into the split, and it is an offboarding path.

So relaxing tenant resolution would not mainly have enfranchised bootstrap admins. It would mainly have enfranchised removed employees, at their old role, in the organization they were removed from. On the per-client production model that is a client's former administrator keeping administrator access. It would also have changed what "the current tenant" means for all 354 places that read `TenantContext.getCurrentOrgId()`, to accommodate a state that should not exist.

## What this does instead

Makes the invariant hold, so the clauses are provably unreachable rather than merely unreachable today, then removes them.

- `assignOrgRole` joins the parent org group before the role subgroup. `joinGroup` is idempotent, and membership goes first so a failure between the two leaves the weaker state.
- `removeUserFromOrganization` clears every role subgroup under that organization before leaving the parent, found by path so a hand-made subgroup or one dropped from `OrgRole` is covered. Roles go first, for the same reason.
- `KeycloakInitializer` reconciles a realm that already holds the split on every startup. One-directional and minimal: it adds the membership a role implies and never grants, removes or alters a role, a credential or an account. It matters because the QA seed replays a captured export verbatim, so it can carry the split in.
- The 36 guards lose the unreachable role clause and keep the membership check that was deciding them. `hasAnyOrgRoleForCurrentTenant` now documents why the two must not be `or`ed, next to the code someone would read before restoring the clause.

**Tenant resolution itself is untouched.** A caller with no relationship to an organization is refused exactly as before, and a bare realm authority still resolves no tenant, so the phantom authorities of #684 and #716 stay dead.

## Evidence

Tests were pushed one commit ahead of the fix so the red is on the record. Run on the tests-only commit: 2209 tests, 6 failed.

| test | before the fix | after |
|---|---|---|
| `KeycloakOrgMembershipInvariantIT.assigningARoleToANonMemberAlsoMakesThemAMember` | FAILED (measured) | passes |
| `KeycloakOrgMembershipInvariantIT.removingMembershipAlsoRemovesTheRoleSubgroup` | FAILED (measured) | passes |
| `KeycloakOrgMembershipInvariantIT.removingMembershipClearsEveryRoleHeldInThatOrganization` | FAILED (measured) | passes |
| `KeycloakOrgMembershipInvariantIT.aRoleAuthorityIsNeverMintedWithoutItsMembershipAuthority` | FAILED (measured) | passes |
| `KeycloakOrgMembershipInvariantIT.anOffboardedUserKeepsNeitherAuthority` | FAILED (measured) | passes |
| `OrgRoleGuardRedundancyTest.noGuardOrsAnOrgRoleAgainstTenantMembership` | FAILED (measured), 36 occurrences | passes |
| `TenantIsolationIT` stranger cases (read, write, HQL join) | PASSED (measured) | passes |
| `ReadWriteGuardSymmetryTest.read_isOk_forARoleHolderThatIsNotRecordedAsAMember` | passed on a stub | removed, see below |

`KeycloakOrgMembershipInvariantIT` runs the real `KeycloakGroupService` against a real Keycloak container and then feeds the resulting group paths through the real `JwtAuthConverter`, so the invariant is asserted in the form the guards read rather than as a claim about group rows.

The stranger cases were green before the fix and after it, which is the point: they are the boundary the guard removal rests on, and this change does not move it. A caller scoped to one organization is refused another's row on a read, on a write, and through an HQL join that names the other organization, with the owning tenant still passing through the same join.

## The vacuous test underneath

`ReadWriteGuardSymmetryTest.read_isOk_forARoleHolderThatIsNotRecordedAsAMember` failed when the clause came off. It is the clause's own passing test, not a regression: it stubbed `isMemberOfCurrentTenant()` false and `hasAnyOrgRoleForCurrentTenant(...)` true, a pair the real beans cannot return, so it asserted an outcome for a request that cannot reach a controller. Same shape as `eec6c37`.

It is removed rather than rewritten, because the claim belongs in `TenantFilterTest.theRoleGuardCannotSucceedWhereTheMembershipGuardFails`, where it is made against the real beans in the real filter chain. In its place, `stubTenantGuards` refuses the impossible pair with the reason in the exception, and a test holds the helper to it. A scan of the whole test tree found exactly one method stubbing that combination, this one.

Closes #717

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/IssueComment/IssueCommentControllerWeb.java
+++ b/src/main/java/org/tornotron/echno_backend/IssueComment/IssueCommentControllerWeb.java
@@ -55,7 +55,7 @@ public class IssueCommentControllerWeb {
     }
 
     @GetMapping
-    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant() or @orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')")
+    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant()")
     @Operation(
             summary = "List issue comments",
             description = "Returns a single page of issue comments. The pageNo and pageSize parameters "
@@ -71,7 +71,7 @@ public class IssueCommentControllerWeb {
     }
 
     @GetMapping("{id}")
-    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant() or @orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')")
+    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant()")
     @Operation(
             summary = "Get an issue comment by id",
             description = "Returns a single issue comment."
@@ -86,7 +86,7 @@ public class IssueCommentControllerWeb {
     }
 
     @GetMapping("/issueId/{issueId}")
-    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant() or @orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')")
+    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant()")
     @Operation(
             summary = "List comments for an issue",
             description = "Returns every comment left on the given issue."
--- a/src/main/java/org/tornotron/echno_backend/asset/AssetControllerWeb.java
+++ b/src/main/java/org/tornotron/echno_backend/asset/AssetControllerWeb.java
@@ -45,7 +45,7 @@ public class AssetControllerWeb {
     }
 
     @GetMapping
-    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant() or @orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')")
+    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant()")
     @Operation(
             summary = "List all assets",
             description = "Returns at most 500 rows. X-Total-Count carries the true total and X-Result-Capped is set when rows were left out; use the paginated variant for a complete result."
@@ -59,7 +59,7 @@ public class AssetControllerWeb {
     }
 
     @GetMapping("/paginated")
-    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant() or @orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')")
+    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant()")
     @Operation(
             summary = "List assets, paginated",
             description = "Returns a page of assets for the caller's current tenant organization. The "
@@ -75,7 +75,7 @@ public class AssetControllerWeb {
     }
 
     @GetMapping("/{id}")
-    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant() or @orgSecurity.hasAnyOrgRoleForCurrentTenant('system-admin','project-manager')")
+    @PreAuthorize("@orgSecurity.isMemberOfCurrentTenant()")
     @Operation(
             summary = "Get an asset by id",
             description = "Returns a single asset, including its resolved vendor and storage location."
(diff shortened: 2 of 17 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

Issue #711, opened 2026-09-06
Decide what a project manager should see of a colleague's personal details


abhijithanandan:
`EmployeeDto` carries `salary`, `dateOfBirth`, `address` and `phoneNumber`. The manager directory returns it, so a `project-manager` reading their team reads all four.

This is a product decision, not a guard to tighten. Three things make it worth taking deliberately:

- **Salary is the one that stands out.** Phone number and address have an obvious operational use on a construction site: somebody needs to reach a person or arrange transport. Pay does not have that justification, and in most organizations it sits with HR and the reporting line rather than with every project manager.
- **The fix is not necessarily narrowing the guard.** A DTO that varies by caller role is the other shape, and it is the one that keeps the directory useful. Narrowing the endpoint would take the phone number away along with the salary.
- **It interacts with #650**, which asks whether there should be a role between plain membership and `system-admin` in Resources. If a new role appears, this question gets asked again for that role, so settling the principle is worth more than settling the one endpoint.

No urgency: nothing here crosses a tenant, and `Employee` carries the `orgFilter`. It is a question about what colleagues should see of each other.

Raised while tracing #690.

abhijithanandan:
Decided and built in #729.

## The cut

`salary` is written for a caller holding **system-admin** or **hr-admin** in the current tenant, and on the caller's own record. `dateOfBirth`, `address` and `phoneNumber` do not move.

Why that line and not another: `PATCH /employee/web/{id}` has always been gated on system-admin and hr-admin, while the directory reads add project-manager. That seam already exists in the module, so pay is now readable by exactly the roles that may set it, and nothing new was invented to hold it.

`dateOfBirth` stays, and it was the field that could have gone either way. It has ordinary uses (age-restricted work, birthdays), and this record already carries a blood group and an emergency contact, which are more sensitive on any honest ranking. Cutting the birth date while leaving those two would be a line drawn nowhere. There is also a concrete cost to cutting it: `echno-core` parses `dateOfBirth` as `parseLocalDateTime(raw.dateOfBirth) ?? new Date()`, so an absent value renders as **today's date** rather than as nothing. Omitting it would put a fabricated birth date on the screen.

A person always reads their own pay. Both single-record reads carry a self branch (the one added in #710) and `GET /user/web/employees` returns nothing but the caller's own records, so without that clause a person would be handed their own record with their own pay cut out of it.

## The mechanism, and why it cannot be forgotten

[...]

PR #729, opened 2026-09-07
Vary the employee record by what the caller may see of pay

abhijithanandan:
`EmployeeDto` carries salary, date of birth, address and phone number. The directory reads admit `project-manager`, so every project manager reads all four of every colleague.

Three of the four earn their place on a construction site: somebody has to reach a person, arrange transport, or know that a job has an age floor. Pay does not, and in this codebase the roles that may set it are `system-admin` and `hr-admin`, which is exactly what the PATCH beside these reads has always been gated on. So pay is now shown to the roles that may change it, plus the person whose pay it is.

Nothing is narrowed at the endpoint. Narrowing the directory would take the phone number away with the salary, and the whole point of the directory is that a manager can reach their team.

Refs #711

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -5223,7 +5223,7 @@
             "type": "string"
           },
           "salary": {
-            "description": "Salary of the employee.",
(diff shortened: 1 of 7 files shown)


abhijithanandan:
@coderabbitai review

Issue #680, opened 2026-09-06
A project manager reads procurement movements through the ledger while every procurement document stays closed to them


abhijithanandan:
Product decision, not a bug. Recorded so the contradiction #672 created is deliberate rather than accidental.

#672 moved ten reads on `InventoryTransactionControllerWeb` to `system-admin, project-manager`, so the role that approves a stock adjustment can read the `ADJUST` rows the approval wrote. The ledger does not separate movement types, so the same reads also return the `GRN`, `TRANSFER_IN`, `TRANSFER_OUT`, `USE` and `PRODUCTION_CONSUME` rows, and `InventoryTransactionDto` carries `referenceNumber`, `quantityChanged`, `remarks` and `unitCost`.

Meanwhile every controller for the documents that wrote those rows is `system-admin` on all of its methods, reads included:

| controller | methods, all `system-admin` |
|---|---|
| `GoodsReceivedNoteControllerWeb` | 7 |
| `PurchaseOrderControllerWeb` | 10 |
| `IndentControllerWeb` | 11 |
| `SiteTransferControllerWeb` | 11 |
| `MaterialConsumptionControllerWeb` | 8 |

So a project manager can now see what a goods receipt booked, at what unit cost, against which reference, without being able to open the goods receipt. The derived view is open and the document is shut.

Two coherent resolutions, and they point in opposite directions:

- **Accept it and open the documents.** A project manager already reads every project, task, employee, material, storage location, stock adjustment and the whole general ledger (the finance module is project-manager on reads and writes). On that evidence the five document controllers are the outlier and should carry `project-manager` reads, with writes staying where they are. Largest surface, but it makes one story.
- **Accept it and narrow the field.** Keep the documents shut and stop the ledger from carrying vendor pricing to non-admin callers, by omitting `unitCost` from `InventoryTransactionDto` unless the caller holds `system-admin`. Smaller change, but a role-varying response body is new in this codebase and would need a deliberate pattern.

Doing neither is also a choice and is the current state.

This is the same shape as #650, which asks what role the Resources domain should have between member and `system-admin`. Whatever answer #650 lands on probably settles this too, so it may be worth taking them together.

Related: #666, #672.

abhijithanandan:
Decided and built in #730: the documents open.

## What moved

Thirty-seven reads now carry `system-admin, project-manager`:

| controller | reads widened | endpoints untouched |
|---|---|---|
[...]

PR #730, opened 2026-09-07
Open the procurement document reads to the role that already reads the ledger

### LINKED ISSUE #695: A project-manager can raise the adjustment that closes a transfer variance but cannot read the transfer it names

abhijithanandan:
#672 moved ten reads on `InventoryTransactionControllerWeb` to `system-admin, project-manager`, so the role that approves a stock adjustment can read the `ADJUST` rows the approval wrote. The ledger does not separate movement types, so the same reads return the `GRN`, `TRANSFER_IN`, `TRANSFER_OUT`, `USE` and `PRODUCTION_CONSUME` rows too, each carrying `referenceNumber`, `quantityChanged`, `remarks` and `unitCost`. Meanwhile every controller for the documents that wrote those rows was `system-admin` on all of its methods, reads included. The derived view was open and the document was shut. That is #680.

It is settled by opening the documents, which is #680's first resolution. A project manager already reads every project, task, employee, material, storage location and stock adjustment, and the whole general ledger. On that evidence these five were the outlier.

**Closes #695** with the same change: a project manager could raise and approve the stock adjustment that settles a transfer receipt variance while being refused the site transfer the adjustment names.

**No write moves.** Reading a purchase order is not raising one, receiving a transfer posts stock into a project, and cancelling one takes it back out. `ProcurementDocumentGuardSplitTest` asserts that across all 56 endpoints by reading the annotation, which is the only way to cover the writes a slice cannot reach.

Refs #680
Closes #695

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -108962,7 +108962,7 @@
       "name": "Goods Received Notes"
     },
     {
-      "description": "Goods received notes (GRNs) record materials received against a vendor, capturing the delivery and its line items. This is the web-console API, gated by organization role for the current tenant rather than by flat authorities. Recording a receipt and reading one are open to the system-admin and store-keeper roles, because booking a delivery in is the store's own work and used to require administering the whole organization. Endpoints cover creating, browsing, filtering by vendor or date range, and updating GRNs.",
+      "description": "Goods received notes (GRNs) record materials received against a vendor, capturing the delivery and its line items. This is the web-console API, gated by organization role for the current tenant rather than by flat authorities. Recording a receipt and reading one are open to the system-admin and store-keeper roles, because booking a delivery in is the store's own work and used to require administering the whole organization. Reading is open to the project-manager role as well, which already sees the same receipt in the stock ledger. Endpoints cover creating, browsing, filtering by vendor or date range, and updating GRNs.",
       "name": "Goods Received Notes (Web)"
     },
     {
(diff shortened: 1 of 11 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

PR #732, opened 2026-09-07
Fix the three defects the nullability pass surfaced

### LINKED ISSUE #728: Three defects the nullability pass surfaced: project join columns, missing @Builder.Default, unwritten clock-event fields
Found while classifying response nullability for #645 (PR #725). None of them is a documentation problem, so none was fixed there: the contract now describes the behaviour as it actually is, and these are the underlying causes.

## 1. Three entities declare the project association non-optional over a nullable column

`Payable`, `PurchaseOrder` and `GoodsReceivedNote` each declare:

```java
@ManyToOne(fetch = FetchType.LAZY)
@JoinColumn(name = "project_id", nullable = false)
private Project project;
```

while the Liquibase column is nullable in all three (`v4.0/baseline-005-level4-tables.xml`, `baseline-003-level2-tables.xml`, `baseline-004-level3-tables.xml`). No `addNotNullConstraint` anywhere in the changelog names a `project_id`, so this is not an artefact of how the changelog was read.

`ddl-auto` is `validate`, and Hibernate's schema validation does not compare nullability, so the application starts cleanly and nothing reports the disagreement. The database wins at runtime, which is why `projectId` and `projectName` are published as nullable on all three schemas.

Needs a decision either way:

- if a project is genuinely mandatory on these three, the fix is a null sweep followed by `addNotNullConstraint` on each column, after which the three DTO properties move to the non-null half of `ReviewedResponseSchemas`;
- if it is not, the `nullable = false` on the three `@JoinColumn` annotations is simply wrong and should go.

Worth noting `project.project_name` is itself nullable, so a flattened `projectName` has two independent ways to be null.

## 2. Six `Attendance` fields carry an initialiser Lombok discards

`Attendance` is a `@Builder` class. These six carry an inline initialiser with no `@Builder.Default`, so the builder passes null through the all-args constructor and the initialiser never applies:

| field | initialiser | column |
|---|---|---|
| `totalWorkMinutes` | `= 0` | nullable |
| `morningSessionMinutes` | `= 0` | nullable |
| `afternoonSessionMinutes` | `= 0` | nullable |
| `overtimeMinutes` | `= 0` | nullable |
| `breakDurationMinutes` | `= 0` | nullable |
| `approvalStatus` | `= ApprovalStatus.PENDING` | `NOT NULL` |

For the five minute fields the null persists, because `AttendanceCalculationService.recalculate` is the only writer and it never runs on a record raised by `markAbsent` or `markLeave`, nor on any record without a shift. So those responses carry `"totalWorkMinutes": null` where the field declaration plainly intends `0`, and a client cannot tell "the day was not computed" from "no minutes were worked" without the description now added in #725.

`approvalStatus` is currently masked: all three builder sites set it explicitly, and its `NOT NULL` column means a fourth site that forgets would fail at insert rather than return null. It is in the same list because it is the same omission and the next creation path will hit it.

Lombok warns on exactly this, so the warning is either suppressed or being ignored in the build output.

The fix is one annotation each, but it changes what existing rows mean, so it needs a decision on whether to backfill the five columns for records that were never calculated.

## 3. `ClockEventDto` publishes three fields nothing ever writes

- `photoUrl`: `ClockEventMapper` declares `@Mapping(target = "photoUrl", ignore = true)`, and `ClockEvent` has no photo field of its own. Nothing calls `setPhotoUrl` anywhere in `src/main`. Photos live in `attachments`.
- `verifiedBy` and `verifiedAt`: the `clock_event` columns exist, but no builder site sets them and no `setVerifiedBy` / `setVerifiedAt` call anywhere in `src/main` targets a `ClockEvent`.

All three are null on every response this server produces. #725 changed their descriptions to say so, and removed the example URL on `photoUrl` that contradicted its own description, but the fields are still published.

Either they should come off the DTO, or `photoUrl` should be derived from the first attachment and the two verification fields wired to whatever was meant to write them. Removing them is a breaking contract change, so it needs a decision rather than a quiet edit.

## Related

- #645 for the nullability work these came out of, and the merged pass in #725.
- The `ReviewedResponseSchemas` javadoc records each of these next to the properties they affect, so a fix here should update the table in the same commit.

abhijithanandan:
Closes #728. Three defects the nullability pass in #725 surfaced but deliberately did not fix, since #725 was documenting behaviour rather than changing it.

Tests are pushed as their own commit ahead of the fix, so the red is on the record: run [34128370607](https://github.com/tornotron/echno-backend/actions/runs/34128370607) failed **12 tests, all of them new, with no collateral damage**.

## 1. The one serving wrong data: five minute fields null where the code means zero

Lombok drops an inline initialiser on a `@Builder` class unless the field carries `@Builder.Default`. The generated builder hands its own uninitialised slot to the all-args constructor, and that assignment overwrites whatever the initialiser wrote.

**What the null actually does, end to end.** Not an exception and not a corrupted sum, which is worth stating precisely because two of the three plausible answers turned out to be wrong:

- **Backend.** The five fields have exactly two readers outside the mapper, both in `AttendanceCalculationService.buildMonthlySummary`, and both are explicitly coalesced: `a.getTotalWorkMinutes() != null ? a.getTotalWorkMinutes() : 0`. Nothing unboxes to `int`, nothing uses `mapToInt`, and no JPQL or native query touches these columns. So there is no NPE and no arithmetic corruption anywhere in `src/main`. `morningSessionMinutes`, `afternoonSessionMinutes` and `breakDurationMinutes` have no backend reader at all beyond the mapper.
- **Contract.** `AttendanceMapper` copies `Integer` to `Integer`, so the null reaches the client as `"totalWorkMinutes": null`.
- **Client.** `echno-core`'s `parseAttendance` coalesces every one of the five with `?? 0` at a single boundary, into a `WorkDuration` whose members are all plain `number`. `echno-web` never reads the raw field names, only the normalised object.

So the user is shown **`0h 0m`, indistinguishable from a day someone clocked in and out at the same minute**. The distinction the field was carrying is destroyed at the parse boundary in core, and `WorkDuration` has no vocabulary to express it. Where it does bite is the two averaging reduces in `attendance-history.tsx` and `app/users/dashboard/attendance/page.tsx`: they divide by `filtered.length`, so absences fold in as genuine zero-hour days.

**It is a data bug, not a display bug.** The nulls are genuinely persisted, and they are permanent. Hibernate binds an explicit NULL, so the `defaultValueNumeric="0"` on all five columns never fires.

**Which rows.** Only `markAbsent` and `markLeave`, and only when no record existed for that employee, date and project. `checkIn` calls `recalculate` unconditionally with a shift that is guaranteed non-null, so every check-in row already stores explicit zeros. `markAbsent` is live at `POST /api/v1/attendance/mark-absent` on both controllers; `markLeave` currently has no caller. Both build the record **without a shift**, and the three later `recalculate` call sites are all guarded on the shift being non-null, so once such a row exists nothing can ever repair it.

**Why zero is the right answer, not a preserved null.** A day with a shift and no punches already gets an explicit `0` from `recalculate`, which walks its locals from `int morningMinutes = 0`. So "not computed" and "computed as zero" were never distinguishable to begin with, and the inconsistency was that two spellings of the same day gave different answers. The fix makes an absence agree with a no-punch day.

**Stored rows are not rewritten by this PR**, so the five properties stay `nullable: true` and the descriptions now say null means a row written before this change. The repair for the existing rows is proposed separately below rather than committed, because a changeset merged here would run on the next deploy without a second decision.

**The other six.** `Attendance.approvalStatus` and `attachments`, `ShiftTiming`'s four thresholds, and `AttendanceRegularization.status` carry the identical omission. Five of those are masked by a `NOT NULL` column, where the miss fails an insert rather than serving a wrong answer, and `ShiftTiming`'s four are additionally masked by `@NotNull` on the creation DTO. `Attendance.attachments` is not masked at all: `markAbsent` leaves it null, which makes `addAttachment` an NPE waiting on that row. All eleven are fixed, since it is one line each and the next creation path would hit them.

Worth noting for anyone reading the diff: `@Builder.Default` applies only when the builder method is never called. A site that passes an explicit null still writes null, so none of the masked six changes any current behaviour.

## 2. `@JoinColumn(nullable = false)` over a nullable column: the annotation is wrong

Derived from the changelog, not the entity. The include graph was checked to be complete first (199 includes, 200 files, no `includeAll`, no nested includes, no orphans), so a negative result over the tree means something.

| | `payable` | `purchase_order` | `goods_received_note` |
|---|---|---|---|
| Changeset | `v4-005-create-payable` | `v4-003-create-purchase-order` | `v4-004-create-goods-received-note` |
| Column | `<column name="project_id" type="BIGINT"/>` | same | same |
| `<constraints>` child | none | none | none |
| Later tightening | none anywhere | none | none |
| Create DTO | `@NotNull` | `@NotNull` | `@NotNull` |
| Can the app write a null? | no | no | no |

The omission is not a file-wide style: `vendor_id` in the same `createTable` does carry `<constraints nullable="false"/>`, and `task`, `wbs_element`, `current_stock` and `attendance` all have a `NOT NULL` `project_id`. These three are the exceptions. Same result on the v3 and the v1/v2 paths, where `080-add-project-id-to-goods-management.xml` added the column nullable to already-populated tables **with no backfill**.

**Decision, the same for all three: correct the annotation, do not tighten the column.**

- Tightening is the risky half and buys nothing today. `addNotNullConstraint` would pass on a fresh v4 install and **fail the migration on any database carrying a pre-`080` row**, which is the worst shape of migration bug: green on the lab staging, red on a client. That risk lands squarely on the per-client production instances.
- `nullable = false` on `@JoinColumn` is not an enforced check. Hibernate uses it for DDL generation, which is off under `ddl-auto: validate`. Removing it loses no protection.
- The real guard already sits where it belongs and is better than a constraint: `@NotNull` on all three creation DTOs with `@Valid` on all six endpoints, plus an unconditional `setProject` after a lookup that throws. That answers a bad request with a 400 or 404 rather than a 500.
- The published contract, the DTO descriptions and `ReviewedResponseSchemas` all already say nullable. Tightening would mean reversing all of it.

Tightening stays available once someone confirms no null rows exist across every deployed database. The sweep is in the report; it is a read, so it is safe to run anywhere.

`ProjectJoinColumnMatchesTheChangelogTest` now reads the expectation out of the changeset rather than writing it down twice, so if a later changeset does tighten one of these columns the test turns red and the entity can follow.

## 3. Three fields nothing has ever written: remove

**`photoUrl` is the leftover half of a migration that finished.** `63af5e41` replaced the photo URL with attachments across check-in and clock events; the entity's own field went in `cfb5322e`; `ClockEventDto.photoUrl` was left behind with the mapper carrying `@Mapping(target = "photoUrl", ignore = true)` to fill the hole.

The selfie path is live and lands one level more precisely than `photoUrl` did. `POST /check-in` and `POST /clock-event` take a `photo` part, gated on `photoRequiredOnCheckIn` / `photoRequiredOnCheckOut`, uploaded as `CLOCK_EVENT_CHECK_IN` or `CLOCK_EVENT_CHECK_OUT`, and linked **per punch** through `Attachment.clockEvent` (FK `fk_attachment_clock_event`). `AttachmentMapper` already signs each one with a one-hour download URL, and `ClockEventDto.attachments` is declared non-null. Deriving `photoUrl` would re-add a single-value view of a list the DTO already returns in full.

**`verifiedBy` and `verifiedAt` are the unstarted half.** Born in `072-create-clock-event` beside `MovementRecord`'s identically named columns; only the movement's were ever wired. That flow needed two more columns (`verified_by_id`, `is_verified`) and a service that takes the verifier from the session, locks the row and refuses to overwrite a stamp. A clock event has none of it. The live "somebody checked this" state for attendance is `Attendance.approvalStatus` / `approvedBy` / `approvedById` / `approvedAt`, decided at day level even though the geofence verdict is computed per punch. Anything that wants per-punch verification has to design it, and would not inherit these two.

**Removal does not break the client**, which matters because `echno-core` is hand-maintained with no codegen, so a break would show up at runtime rather than at compile time. Checked rather than assumed:

- `echno-core`'s `ClockEventResponseSchema` declares all three through `nullableString` / `backendDate`, both `z.string().nullish()`. `.nullish()` is `.nullable().optional()`, so an **absent key parses cleanly**; the object is non-strict besides.
- `parseClockEvent` does `verifiedAt: parseUTCDate(raw.verifiedAt) ?? undefined`, which yields `undefined` either way.
- `echno-web` has **zero** read sites for any of the three on a clock event. Every `verifiedBy` / `verifiedAt` hit in web belongs to `ConstructionPayment`, `MovementRecord` or `NcrReport`; the one `photoUrl` hit is an employee profile photo in a mock-data file.
- There is precedent: #631 removed `verifiedBy` and `verifiedAt` from the payment DTOs, and core absorbed it.

`ClockEvent`'s entity fields go with them. The `clock_event.photo_url`, `verified_by` and `verified_at` **columns are left in place**: dropping a column is irreversible and is its own decision, and `validate` does not object to a column with no mapped field.

## Follow-ups this turned up, not fixed here

1. **`echno-core` should drop the three from `ClockEvent`.** Nothing breaks while it keeps them, since they simply stop arriving, but the type says the server sends something it no longer does. `photoUrl: string` is already declared non-optional there while the server has been sending null, so it was wrong before this PR too.
2. **`AttachmentService.linkToEntity` mixes two ids.** `case "attendance"` does `attachment.setAttendance(attendanceRepository.findById(entityId)...)`, and the attendance photo path passes `entityId = clockEvent.getId()` with `folder = "attendance"`. A punch photo's `attendance_id` is therefore set from a clock-event id, pointing at an unrelated attendance row or at nothing. The `clock_event_id` link is correct, because `addAttachment` sets it afterwards, which is why nothing has noticed.
3. **Request-side `photoUrl` orphans**: `ClockEventCreationDto.photoUrl` and `AttendanceClockEventDto.photoUrl` are accepted on the wire and read by nothing. Also published in `docs/openapi.json` as a non-nullable string with an example URL. Request-side is out of scope here as it was in #670, #678 and #725.
4. **`InventoryEventListener.handleGrnCreated`** dereferences `grn.getProject().getId()` three times with no guard. It only ever runs on a GRN the service just created, so it cannot reach a legacy null row, but it is the one place where one would throw rather than degrade.
5. **`markLeave` has no caller** anywhere in `src/main` or `src/test`. It is fixed here alongside `markAbsent` rather than left as a trap, but it is dead code and should either be wired to the leave approval flow or removed.

---

## Summary by CodeRabbit

- **Attendance**
  - Clock-event responses now provide punch photos through attachments with signed download URLs.
  - Removed unused photo URL and verification timestamp fields from clock-event data.
  - Attendance and shift calculations now reliably default to zero or configured values when records are created.
  - Absence records store zero duration values when no work was performed.

- **Data Compatibility**
  - Project associations now support legacy records without a project while application-created records continue to require one.

- **Documentation**
  - Updated API descriptions to clarify attendance totals and legacy nullable project values.

### DIFF
--- a/docs/ATTENDANCE_MODULE.md
+++ b/docs/ATTENDANCE_MODULE.md
@@ -131,15 +131,14 @@ Multi-tenancy is enforced via a Hibernate `orgFilter` that automatically scopes 
 | `longitude` | Double | GPS longitude (optional) |
 | `gpsAccuracy` | Double | GPS accuracy in meters |
 | `altitude` | Double | GPS altitude |
-| `photoUrl` | String | Photo taken at clock event |
 | `devicePlatform` | String | "android" / "ios" / "web" |
 | `deviceId` | String | Device identifier |
 | `ipAddress` | String | Client IP address |
 | `isWithinGeofence` | Boolean | Whether location is within project geofence |
 | `distanceFromProject` | Double | Distance from project location in meters |
-| `verifiedBy` | String | Supervisor who verified this event |
 | `isRegularized` | Boolean | Whether this event was added via regularization |
 | `regularizationReason` | String | Reason for regularization |
+| `attachments` | List&lt;Attachment&gt; | Photos taken with the punch, each returned with a signed download URL |
 
 ---
 
@@ -556,7 +555,6 @@ Record any subsequent clock event (lunch out, lunch in, clock-out) on an existin
   "longitude": 77.5946,
   "gpsAccuracy": 5.0,
   "altitude": 920.0,
-  "photoUrl": null,
   "devicePlatform": "android",
   "deviceId": "device-uuid-123",
   "ipAddress": "192.168.1.10",
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -1357,7 +1357,7 @@
         "description": "A single day's attendance record for an employee, with its clock events, movements, regularizations and approval state.",
         "properties": {
           "afternoonSessionMinutes": {
-            "description": "Minutes worked in the afternoon session, after the lunch break. Null whenever the day's totals have not been computed.",
+            "description": "Minutes worked in the afternoon session, after the lunch break. Null on the same terms as totalWorkMinutes.",
             "example": 255,
             "format": "int32",
             "type": [
@@ -1408,7 +1408,7 @@
             "type": "string"
           },
           "breakDurationMinutes": {
-            "description": "Total minutes spent on breaks during the day. Null whenever the day's totals have not been computed; a computed zero means no break was punched.",
+            "description": "Total minutes spent on breaks during the day. Zero when no break was punched. Null on the same terms as totalWorkMinutes.",
             "example": 60,
             "format": "int32",
             "type": [
@@ -1491,7 +1491,7 @@
             ]
           },
           "morningSessionMinutes": {
-            "description": "Minutes worked in the morning session, before the lunch break. Null whenever the day's totals have not been computed.",
+            "description": "Minutes worked in the morning session, before the lunch break. Null on the same terms as totalWorkMinutes.",
             "example": 225,
             "format": "int32",
             "type": [
@@ -1507,7 +1507,7 @@
             "type": "array"
           },
           "overtimeMinutes": {
-            "description": "Minutes worked beyond the shift's overtime threshold. Null whenever the day's totals have not been computed, which is different from a computed zero meaning no overtime was earned.",
+            "description": "Minutes worked beyond the shift's overtime threshold. Zero when no overtime was earned. Null on the same terms as totalWorkMinutes.",
             "example": 30,
             "format": "int32",
             "type": [
@@ -1575,7 +1575,7 @@
             "type": "string"
           },
           "totalWorkMinutes": {
-            "description": "Total minutes worked across all sessions. Null until the day is computed, which never happens on a record with no shift or one raised by marking someone absent or on leave. Null means nothing was calculated and must not be read as zero minutes worked.",
+            "description": "Total minutes worked across all sessions. Zero on a day nobody worked, including one raised by marking someone absent or on leave. Null only on a record written before absences began storing a zero here; those rows were left as they are, so a client reading history still has to allow for it.",
             "example": 480,
             "format": "int32",
             "type": [
@@ -2999,13 +2999,6 @@
               "null"
             ]
           },
-          "photoUrl": {
-            "description": "Always null. The mapper ignores this field and the clock event has no photo of its own; photos taken with a punch are returned in attachments instead, each with its own signed download URL.",
-            "type": [
-              "string",
-              "null"
-            ]
-          },
           "projectId": {
             "description": "Id of the project the event was recorded against.",
             "example": 12,
@@ -3037,23 +3030,6 @@
           "remarks": {
             "description": "Optional remarks about the event. Null where none were given, and always null on an event written by a regularization.",
             "example": "Reported directly to the second floor slab pour",
-            "type": [
-              "string",
-              "null"
-            ]
-          },
-          "verifiedAt": {
-            "description": "When the event was verified. Null on the same terms as verifiedBy, and the two are written together if anything ever writes them.",
-            "example": "2026-01-16T11:00:00",
-            "format": "date-time",
-            "type": [
-              "string",
-              "null"
-            ]
-          },
-          "verifiedBy": {
-            "description": "Who verified the event. The column exists but no code path in the application writes it, so it is null on every event this server records.",
-            "example": "Anand Rajashekar",
             "type": [
               "string",
               "null"
@@ -6094,7 +6070,7 @@
             "type": "boolean"
           },
           "projectId": {
-            "description": "Project the goods were received for. The column permits null and rows exist without it, although the entity declares the association non-optional.",
+            "description": "Project the goods were received for. The column permits null. Every route the application offers requires a project, so a null here names a row that did not come through one, such as a database migrated from before the column existed.",
             "example": 42,
             "format": "int64",
             "type": [
@@ -13460,7 +13436,7 @@
             "type": "string"
           },
           "projectId": {
-            "description": "Id of the project the amount is charged to. The column permits null and rows exist without it, although the entity declares the association non-optional.",
+            "description": "Id of the project the amount is charged to. The column permits null. Every route the application offers requires a project, so a null here names a row that did not come through one, such as a database migrated from before the column existed.",
             "format": "int64",
             "type": [
               "integer",
@@ -14983,7 +14959,7 @@
             "type": "string"
           },
           "projectId": {
-            "description": "Id of the project the materials are for. The column permits null and rows exist without it, although the entity declares the association non-optional.",
+            "description": "Id of the project the materials are for. The column permits null. Every route the application offers requires a project, so a null here names a row that did not come through one, such as a database migrated from before the column existed.",
             "example": 3,
             "format": "int64",
             "type": [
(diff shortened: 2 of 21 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

abhijithanandan:
@coderabbitai review

Issue #735, opened 2026-09-07
The self clause on the employee PATCH lets a person set their own salary, status and reporting line


abhijithanandan:
`PATCH /api/v1/employee/web/{id}` is guarded by

```java
@PreAuthorize("@orgSecurity.isSelfOrHasAnyOrgRole(#id, 'system-admin', 'hr-admin')")
```

and #716 has just given the mobile twin the same expression, so the two agree. The self clause is right for what it was added for: a person maintaining their own contact details is ordinary, and it is what makes the read beside it symmetrical.

The problem is what "maintain" reaches. `EmployeeService.partialUpdateAnEmployee` accepts a map and switches on the key, and the accepted keys are:

`employeeId`, `status`, `employeeName`, `designation`, `joiningDate`, `phoneNumber`, `salary`, `shiftTimingId`, `department`, `emailAddress`, `dateOfBirth`, `managerId`

Six of those are not the subject's to set. `salary` is the obvious one: an employee can raise their own pay, and the roles that are supposed to decide pay are `system-admin` and `hr-admin`, which is exactly what this guard names for everybody else. `status` lets somebody reactivate a record that was closed. `designation`, `department`, `employeeId` and `managerId` are all organizational facts rather than personal ones, and `managerId` in particular rewrites the reporting line that other guards then read.

This is not a phantom guard and nothing about it is dead. It is a guard that answers "may this caller touch this record" where the question is "may this caller touch this field of this record".

## Shape of a fix

The field list is already a switch in one method, so the cheapest correct answer is probably to split the accepted keys into a self-editable set and an admin-only set, and refuse a self-caller who names a key outside theirs, rather than silently dropping it. The `default` branch already reports an unknown key through `PartialUpdateKeys.reportUnknown`, so there is a place for the refusal to live.

Worth deciding alongside: the same map reaches `PATCH /api/v1/employee/batch`, which #716 gated on `system-admin` and `hr-admin` with no self clause, so that route is unaffected and would stay unaffected.

Related: #711 is narrowing what `EmployeeDto` shows of pay by caller role. This is the write side of the same question.

PR #737, opened 2026-09-07
Scope what a person may change about their own employee record

abhijithanandan:
`PATCH /employee/{id}` and its `/web` twin are gated on
`isSelfOrHasAnyOrgRole(#id, 'system-admin', 'hr-admin')`. The self clause is right for what it was
added for: maintaining your own contact details from the phone is ordinary, and it is what keeps
the endpoint level with the read beside it. What the guard was never asked is which field. The map
the service accepts reaches pay, employment status, the organizational identifier, the joining
date, the shift and the reporting line, so the self clause covers all of those too.

The answer is the field list rather than the guard. Removing the self clause would take away
self-service to solve a question about eight fields.

Refs #735

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -5404,7 +5404,7 @@
         "type": "object"
       },
       "EmployeeUpdateFieldsDto": {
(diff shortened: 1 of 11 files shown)


abhijithanandan:
@coderabbitai review

PR #738, opened 2026-09-07
Tell the storekeeper when a material crosses its reorder level

### LINKED ISSUE #667: Decide whether a low-stock sweep should run, and what it raises, before building one
The query half of #652 is built and live: `GET /api/v1/materials/low-stock` (PR #663). Anyone who asks gets a correct answer at organization, project or storage-location scope. Nothing happens when nobody asks, and this is the decision that has to be made before anything can.

The mechanism is not the hard part. `ComplianceRuleSweep` (#553) is the shape to copy, down to being `@ConditionalOnProperty` and off by default. Three questions gate it and none of them can be answered from the codebase.

## 1. Where does it go

There is no delivery path that works today.

- **Infrastructure:** echno-deployment#66. Alertmanager's Discord webhook targets a forum channel; Discord refuses every execute against one without a `thread_name` or `thread_id`, and Alertmanager 0.27.0 sends neither and has no field for either. Eight of ten sends failed, and nine rules fired into a void for five weeks before anyone looked. This is not a channel to add traffic to.
- **Application:** the backend has no notification package, no `JavaMailSender`, and no in-app inbox. "Send an in-app notification" is a subsystem, not a field.

A job that fires per material per tenant per night into a path that drops everything reproduces exactly the failure in #66, with a bill attached.

## 2. Who receives it

There is no stores or storekeeper role. Most of Resources is `system-admin` only, so the only person who can act on a low-stock alert today is an organization administrator. That is the recipient most likely to mute it, and the role model is a prerequisite rather than a detail.

## 3. What it raises

- an in-app notification (needs the subsystem above)
- an email (needs a sender and an address policy)
- a drafted indent

The last is the one that changes what procurement means, and it is also the one this codebase is closest to being able to do: `indent` and `indentItem` already exist, and the endpoint already returns the minimum order quantity in force at the scope, which is the quantity such a draft would be raised for.

## Design, for when the answers exist

Follow `ComplianceRuleSweep` rather than inventing a second shape:

- `@ConditionalOnProperty(name = "inventory.reorder-sweep.enabled", havingValue = "true")`, false by default, so the bean does not exist until someone turns it on.
- `@Scheduled` elects no leader. On N replicas the pass runs N times, and a cap held in a local variable is really N caps. The compliance sweep counts its budget from a shared table for exactly this reason; anything periodic here inherits the problem and should use the same answer.
- **Fire on the transition across the level, not on every pass.** That needs somewhere to remember it already fired for this material at this scope, and somewhere to forget once stock comes back above. A nightly re-raise of the same forty materials is the version that gets muted in a week, and it is the version that gets built when nobody decides.
- **Scope has to be chosen, not defaulted.** Organization and project are different events. On staging the organization total says `TNT Steel` is healthy at 60 against a level of 30, while it sits at 1, 18, 1, 7 and 1 unit at five separate sites. The per-project answer is where the real shortfalls are, and it is also the noisier one.
- Bound the scan and the spend the way `ComplianceSweepProperties` does, and take candidates least-recently-attempted first so a capped pass still makes deterministic progress.

Until 1 and 2 have answers, the endpoint is the whole of what can be built correctly.

abhijithanandan:
Closes #667.

The query half of this was built (#652, PR #663) and nothing scheduled it, because the issue named three questions that had to be answered before a job could exist. Here are the answers and the job that follows from them.

## It notifies. It raises nothing.

No purchase order, no indent, no procurement document of any kind, and no approval workflow. That is the decision, not an omission. A nightly job that raised paperwork would be creating documents carrying approvals and money on nobody's instruction. Ignoring a notification is free; withdrawing an indent is not.

## Who is told

The storekeepers assigned to the project. If it has none, its project managers. Nobody otherwise.

`OrgRole.STORE_KEEPER` landed with #650, and it is exactly the person who books the goods receipt that answers a shortfall, which is the whole test for a recipient: can they act on it, in this application, today. An organization administrator is deliberately **not** the third rung. They cannot act on a site they do not run, and the issue names them correctly as the recipient most likely to mute the category, which loses every later crossing along with this one.

A project with nobody in either role is counted and logged, and no notification is sent **and no latch is written**, so the first crossing still reaches whoever is eventually assigned rather than having been silently spent on nobody.

Org roles are read from the `Employee` row where Keycloak's groups are mirrored. The admin client is user-id-in only with no reverse lookup from a subgroup to its members, so "who holds this role" is a question only the database can answer. `Project` has no manager field either, so the intersection of the project's assignment list with the role holders is a new query.

## Project scope, not organization and not location

The organization total hides precisely what this is for: on staging `TNT Steel` reads healthy at 60 against a level of 30 while holding 1, 18, 1, 7 and 1 unit at five separate sites. Location scope is one event per shelf and has no recipient of its own, because the person told about a location is the storekeeper on its project.

## What "low" means, including the unset case

Not restated by the job. It calls `LowStockRepository.findLowStockForProject`, the same read the endpoint answers from, so the two cannot drift apart. What that gives:

- **At or below**, not below. A reorder level is the level at which you reorder.
- **A null level is never reported, at any scope.** Nobody has said what low means for that material. A job that guessed would invent a threshold and then wake somebody about it.
- **A level of zero is a set level.** Somebody drew the line at nothing on hand. It reads as "tell me when it is gone" and is reported once the material holds nothing, never before. This is the state most easily conflated with the one above and it is a different state entirely.
- **Only materials the project carries.** Project scope inner-joins, so the catalogue is not reported as absent from every project that never held it.

## Repeat suppression: crossing, not being below

Being below is a standing fact and the endpoint answers it on demand. Re-sending it nightly is how the category gets muted. So `material_reorder_alert` latches (organization, project, material): a row means "reported, and not recovered since".

Two things end a latch, and the first is the answer to a material that hovers:

- **Recovery past a margin.** The latch clears once the project holds more than the level plus `rearm-margin` of it (10% by default), not the moment it ticks one unit above. Without the margin, a material oscillating on the boundary is re-reported every pass, which is the same nightly re-raise reached from the other side. A level of zero needs no special case: a fraction of zero is zero, so it rearms at anything on hand.
- **The level moving.** A material reported at 30 and now measured against 100 is below a different line than the one anybody was told about. Without this, raising a level latches a material permanently at the moment it becomes badly short.

A material that stays down for a month is reported once. That is intended: the notification says something changed, and the standing list of what is short is the endpoint's job.

The unique constraint on the latch is also the idempotency key. `@Scheduled` elects no leader, so on N replicas the pass runs N times, and the constraint is what makes the second replica's insert fail rather than the storekeeper be told twice. The notification subsystem has no idempotency key of its own: `(entity_type, entity_id)` is indexed but not unique, and the key here is a pair one id column cannot hold anyway.

## Safety

- **Off by default.** `inventory.reorder-sweep.enabled` is false and the bean is `@ConditionalOnProperty`, so nothing exists until somebody turns it on.
- **03:30 UTC, not 02:00.** `development` is deployed to both stagings and already carries the compliance re-assessment pass at 02:00. No reason to put both on the same minute.
- **Tenant scope.** The pass is `@WithoutTenant` with a reason and reads two ids at that level. Every read and write of a tenant's rows runs inside `TenantScopedJobRunner.runForTenant`. Both isolation mechanisms fail *open* on a missing organization id and `UnscopedAccessGuard` warns rather than denies, so a scheduled job that forgot its scope would read every tenant's rows and look healthy doing it.
- **Nothing throws out of the trigger.** Spring stops rescheduling a `@Scheduled` method that throws.
- **Bounded.** A scan cap on projects, a per-project cap on materials (most depleted first), and a per-run notification cap counted from the table rather than a local variable, so N replicas share one budget instead of having one each.

## The notification seam

`NotificationService` had five hardcoded senders, each with a `LeaveRequest` in its signature, so nothing outside that package could raise a notification without a sixth being written for it there. `NotificationDraft` splits the message from the recipient, which is also what role targeting needs: `Notification.recipient` is a single employee foreign key, so a role-targeted notification is a fan-out and the wording has to be composed once and delivered N times.

## Tests, and what each one actually proves

- **`LowStockSweepTest`** exercises the **logic**. It calls `runPass()` directly, so it says nothing about the schedule. It covers the recipient tiers including the never-fall-back-to-admin rule, the first crossing, the hovering material inside the margin, real recovery, a zero level rearming at one unit, a cleared level, a material holding nothing on the project, the level-moved re-notify, the lost latch race, the shared cap, and that the tenant context is established and restored. It uses the real `TenantScopedJobRunner`, subclassed only to record what it was scoped to, because stubbing it out would remove the one thing standing between the job and every tenant's rows.
- **`LowStockSweepScheduleTest`** exercises the **trigger declaration**. That the `@Scheduled` exists and reads from configuration, that the default cron parses, that its first firing does not coincide with the compliance sweep's, that `@ConditionalOnProperty` is off by default, that `@WithoutTenant` carries a reason, and that a failing pass does not throw out of `sweep()`. Nothing here waits for a cron to fire, and it does not claim to.

The tests are the commit before the implementation, so the red is on the record.

## Migration

`093-material-reorder-alert.xml`, one new table with a unique constraint, two indexes and cascading foreign keys. Nothing existing is altered.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -11327,7 +11327,8 @@
               "LEAVE_CANCELLED",
               "LEAVE_BALANCE_LOW",
               "LEAVE_REMINDER",
-              "APPROVAL_DELEGATED"
+              "APPROVAL_DELEGATED",
+              "MATERIAL_LOW_STOCK"
             ],
             "example": "LEAVE_PENDING_APPROVAL",
             "type": "string"
--- a/src/main/java/org/tornotron/echno_backend/employee/EmployeeRepository.java
+++ b/src/main/java/org/tornotron/echno_backend/employee/EmployeeRepository.java
@@ -109,6 +109,31 @@ public interface EmployeeRepository extends JpaRepository<Employee,Long> {
     List<Employee> findByOrganizationIdAndOrgRole(Long orgId, OrgRole role);
 
     /**
+     * The holders of one organization role who are assigned to one project.
+     *
+     * <p>Org roles are held on the employee and are organization-wide, so "the storekeeper of
+     * project 12" is not a field anywhere: it is the intersection of the project's assignment
+     * list with the holders of the role. Nothing computed that intersection before this, and
+     * every caller that needs it needs the same one.
+     *
+     * <p>Both joins are association joins out of a root the {@code orgFilter} has already
+     * narrowed, so neither can widen the result past the tenant, and the organization predicate
+     * is written out anyway because a background caller runs with the filter established
+     * explicitly rather than by a request.
+     *
+     * @param orgId The organization the employee and the project both belong to.
+     * @param projectId The project they must be assigned to.
+     * @param role The organization role they must hold.
+     * @return The employees holding that role on that project, in no particular order. Empty
+     *         when nobody on the project holds it, which is a normal state and not an error.
+     */
+    @Query("SELECT DISTINCT e FROM Employee e JOIN e.orgRoles r JOIN e.projects p "
+            + "WHERE e.organization.id = :orgId AND p.id = :projectId AND r = :role")
+    List<Employee> findByProjectAndOrgRole(@Param("orgId") Long orgId,
+                                           @Param("projectId") Long projectId,
+                                           @Param("role") OrgRole role);
+
+    /**
      * Whether the employee holds one of the given roles <em>in the given organization</em>.
      *
      * <p>The organization predicate is written out rather than left to the {@code orgFilter}.
--- a/src/main/java/org/tornotron/echno_backend/leave/NotificationDraft.java
+++ b/src/main/java/org/tornotron/echno_backend/leave/NotificationDraft.java
@@ -1,0 +1,36 @@
+package org.tornotron.echno_backend.leave;
+
+import org.tornotron.echno_backend.leave.enums.NotificationType;
+
+/**
+ * One notification, composed but not yet addressed.
+ *
+ * <p>Every sender on {@link NotificationService} before this one was a whole method per event,
+ * with the leave request it was about hard-wired into its signature. That is workable while every
+ * event is a leave event and stops being workable at the first one that is not: a caller outside
+ * the leave package cannot describe what it wants to say without a method being written for it
+ * inside a package it has nothing to do with.
+ *
+ * <p>Splitting the message from the recipient is the other half of it. A notification aimed at a
+ * role is a fan-out, because {@code Notification.recipient} is a single employee and there is no
+ * broadcast row, so the same draft has to be delivered several times over. Composing it once and
+ * addressing it N times is what {@link NotificationService#deliverToAll} does, and it keeps the
+ * wording of an event in one place rather than once per recipient.
+ *
+ * @param type What kind of event this is. Drives grouping and filtering in the inbox.
+ * @param title The one-line heading. At most 200 characters, which the column enforces.
+ * @param message The body. At most 1000 characters, which the column enforces.
+ * @param entityType What the notification is about, as a bare string: {@code "MATERIAL"},
+ *         {@code "LEAVE_REQUEST"}. Nullable, for a notification about nothing in particular.
+ * @param entityId The id of that thing, in the recipient's own organization. Nullable with
+ *         {@code entityType}.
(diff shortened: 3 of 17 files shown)


abhijithanandan:
@coderabbitai review

Issue #734, opened 2026-09-07
The phantom-authority guards left after #716, starting with the two this session walked into


abhijithanandan:
#641, #684, #710 and #716 each repaired one controller's worth of guards that name a bare `resource:scope` authority. `JwtAuthConverter.extractPermissions` mints those only from the `authorization` claim of an RPT, and the realm carries no authorization scopes, so every one of them refuses every caller. The mechanism has now been confirmed three times and is not in question.

What is still open is how much of it is left. Two surfaces turned up while settling #716 and are worth naming because each is a live endpoint that looks guarded and is not:

## 1. `EmployeeController.joinOrganization` is still dead

#697 bound its guard to the organization the route names, and the result reads:

```java
@PreAuthorize("@orgSecurity.isCurrentTenant(#orgId)"
        + " and (hasAuthority('employee:create') or hasAuthority('employee:admin'))")
```

The tenant half now works. The authority half cannot be satisfied, and it is ANDed, so the endpoint refuses everybody. Its web twin does the same job under `isCurrentTenant(#orgId) and hasAnyOrgRoleForCurrentTenant('system-admin','hr-admin')`, which is presumably what this should say. It was left alone in #716 because #716 enumerated six endpoints and this was not one of them, and because #697 shaped this expression recently enough that somebody may still own it.

Worth knowing before repairing it: nothing is blocked by it today. A person with no membership onboards by redeeming an invite code, which reaches `EmployeeService.joinOrganization` in process rather than through this route.

## 2. `IssueCommentController` is dead end to end

All three of its guards are phantom:

| endpoint | guard |
|---|---|
| `POST /api/v1/issues/comments` | `issue-comment:create` or `issue-comment:admin` |
| `GET /api/v1/issues/comments` | `issue-comment:read` or `issue-comment:admin` |
| `DELETE /api/v1/issues/comments/{id}` | `issue-comment:delete` or `issue-comment:admin` |

The web twin carries the working versions, so nothing is lost today, which is exactly why this went unnoticed. It came up because #676 added an edit to the web twin and deliberately did not add one here: a new endpoint on this controller would have arrived refusing every caller.

## The size of it

[...]

Issue #675, opened 2026-09-06
Decide whether a direct employee create endpoint should exist, or delete the commented-out one


abhijithanandan:
`EmployeeControllerWeb` has a plain `@PostMapping` sitting commented out (lines 78-88 on `development`), along with the `createEmployee` handler and its `employee:create` authority. So `POST /api/v1/employee/web` is a path nothing routes, and the only employee create the backend serves is

    POST /api/v1/employee/web/joinOrganization/userId/{userId}/organizationId/{orgId}

which provisions an employee record for a user who already exists in Keycloak.

This surfaced on tornotron/echno-core#57, the register of request fields and endpoints this client sends that no endpoint reads. echno-core published an `employeeService.create` addressing the absent path, plus a `useCreateEmployee` hook that threw rather than call it. Both have now been removed from the package, so nothing is calling the path any more and the client side is settled either way. What is not settled is the question this issue is for.

## The decision

Should a direct employee create exist?

**Keep it absent.** An employee is a person's membership of an organization, and a membership needs a user to hang off. Creating both in one call means the endpoint also owns Keycloak user provisioning, the invite mail and the password path, which is a different feature from what the commented-out handler was. On this reading the commented block should be deleted rather than left as a suggestion, since it has misled at least one client already.

**Or bring it back.** An admin adding a site engineer who has never signed in has no route today: somebody has to create the user first. If that is the intended flow, the endpoint needs a fresh design rather than uncommenting: it needs to decide what it does about the Keycloak user, and the old `EmployeeCreationDto` field names were never checked against anything because nothing ever called them.

Whichever way it goes, the commented-out block should not survive it.

## What to check before deciding

- Whether the mobile app (`echno`, Flutter) calls a direct employee create. It is stale, last pushed 2025-11, but it is the only other client.
- Whether `EmployeeController` (the non-web prefix) publishes a plain POST. If it does, the question is narrower: it is about the web variant only.

abhijithanandan:
Checked the two things this issue said to check, and the first one changes the shape of the decision.

**`EmployeeController` does publish a plain POST.** `POST /api/v1/employee` is live on the non-web prefix (`EmployeeController` line 84), gated on `hasAuthority('employee:create') or hasAuthority('employee:admin')`, taking `EmployeeCreationDto`. So a direct employee create is not absent from the product. It is absent from the web prefix only.

That makes the question narrower and sharper: **should the web console call the endpoint that already exists, or should the web variant be uncommented, or neither?**

[...]

PR #739, opened 2026-09-07
Repair the issue guards and the mobile join route, and inventory what is left

abhijithanandan:
`JwtAuthConverter.extractPermissions` mints a `resource:scope` authority only from the `authorization` claim of an RPT, and the realm defines no authorization scopes, so `hasAuthority('thing:action')` is satisfied by nobody. Established by #641, #684, #710 and #716, inventoried in #734.

## The true count

#734 reported 168 as an upper bound. Reading the annotations rather than grepping lines, on `development` at `d6cbff54`, the live figure was **136 guards on 17 controllers**. The 168 counted commented-out annotations, non-guard uses of the same string in `TenantFilter` and `OrganizationSecurityService`, and the whole leave module, which was repaired already. 313 is the raw occurrence count, because most guards name two authorities on one line.

Of the 136, 133 are phantom-only and refuse everybody outright. Two of the three mixed ones are dead as well: the mobile `joinOrganization` ANDs the phantom onto the tenant check, and `OrganizationController.readOrganization` has a phantom in both branches of its `or`. So **135 endpoints refuse every caller**, and one, `deleteOrganization`, survives on an `@orgSecurity.hasOrgRole` branch.

## What this repairs

The two surfaces #734 names, plus the sibling controller of one of them, taken as the Issues module:

| endpoint | now |
| --- | --- |
| `GET /api/v1/issues/{id}`, `POST /api/v1/issues` | `isMemberOfCurrentTenant()` |
| `PATCH /api/v1/issues/{id}`, `DELETE /api/v1/issues/{id}` | `system-admin, project-manager` |
| `POST /api/v1/issues/comments`, `GET /api/v1/issues/comments` | `isMemberOfCurrentTenant()` |
| `DELETE /api/v1/issues/comments/{id}` | `system-admin, project-manager` |
| `POST /api/v1/employee/joinOrganization/{userId}/{orgId}` | `isCurrentTenant(#orgId)` and `system-admin, hr-admin` |

Every target is the `/web` twin's answer, endpoint for endpoint, so **nobody's effective reach changes**: the same caller already does each of these things through the twin. Raising a defect from the site with a photograph is the workflow the mobile issue controller exists for, which is why it is repaired rather than proposed for removal.

No route is added. The twin's author-only comment edit has no counterpart here and none is introduced, because adding a route is a contract change rather than a guard repair.

**`EmployeeController.createEmployee` keeps its refusal.** It has no caller, cannot succeed against a non-null `user` column, and resolves its tenant from `organizationName` in the body, so repairing its guard alone would turn an unreachable route into a cross-tenant write. It is proposed for removal under #675 and #716 and is deliberately not repaired by analogy with the guards beside it.

## What is left, and why it is a decision rather than a task

128 guards on 15 controllers, all mobile twins of controllers whose `/web` side works, so nothing a user does today is blocked by one. `PhantomAuthorityCensusTest` records them exactly, by scanning every `@RestController` and reading the annotations, so the number can only come down and a new phantom guard fails the build. Removing an entry means that controller was repaired or removed.

The remainder is not one decision. A dead read whose twin is alive is usually a mobile surface that should work; a dead write may still be required, as the mobile leave endpoints turned out to be; and an endpoint with no caller anywhere is a removal candidate, which is a change to a published contract. The Resources cluster in particular (Material, Vendor, GRN, Indent, IndentItem, SiteTransfer, MaterialConsumption, InventoryTransaction, StorageLocation: 98 of the 128) has just had its `/web` role model settled by #727 and #730, and its mobile twins are also candidates for the twin collapse, so repairing them now would be deciding the collapse by accident.

## Tests

The tests are the first commit, ahead of the fix, but the red is **reasoned rather than measured**: `pull-request.yml` runs on `pull_request` only and the PR was opened after all three commits were pushed, so no CI run exists at `ed80ca04`. The reasoning is deterministic rather than a guess. Every allowed-case assertion sends `.with(jwt())` carrying no authority at all, against a pre-fix guard that requires one, so each asserts 200 or 201 where the old code answers 403; and `PhantomAuthorityCensusTest` expects no entry for two controllers that carried four and three phantom guards before this branch. `EmployeeMobileJoinGuardTest` needed rewriting for a reason worth naming: its passing case granted `employee:create` on the token directly, so it went green against an endpoint that refused every real caller. That is the shape of vacuous test this whole family hides behind. Every role list is stubbed as its exact varargs, with the `any(String[].class)` stub kept answering false, so a guard that widened to every member fails rather than falls through to a Mockito default.

The slice tests exercise a mock of `@orgSecurity`, so `PhantomAuthorityCensusTest` reads the real annotations as well.

`docs/org-scoped-roles.md` taught the layer-1 idiom without saying it cannot be satisfied, which is how these guards kept being written. It now says so.

Refs #734

## Summary by CodeRabbit

- **Authorization**
  - Updated issue and issue-comment access to use tenant membership and role-based permissions.
  - Tenant members can view and create issues and comments.
  - Issue updates/deletions and comment deletions are restricted to system administrators or project managers.
  - Employee organization joining is restricted to system administrators or HR administrators in the current tenant.

- **Documentation**
  - Updated API descriptions and authorization guidance to reflect tenant-role access controls.

- **Tests**
  - Added coverage for tenant membership and role-based authorization across issue, comment, and employee workflows.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -57080,7 +57080,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant, or has no employee record in it, so the record would name nobody"
+            "description": "Caller is not a member of the current tenant, or has no employee record in it, so the record would name nobody"
           },
           "404": {
             "content": {
@@ -57214,7 +57214,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller is not a member of the current tenant"
           },
           "404": {
             "content": {
@@ -57324,7 +57324,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant, or has no employee record in it, so the record would name nobody"
+            "description": "Caller is not a member of the current tenant, or has no employee record in it, so the record would name nobody"
           },
           "404": {
             "content": {
@@ -59503,7 +59503,7 @@
                 }
               }
             },
-            "description": "Caller lacks the required role in the current tenant"
+            "description": "Caller is not a member of the current tenant"
           },
           "404": {
             "content": {
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -34239,7 +34239,7 @@
                 }
               }
             },
-            "description": "Caller lacks the employee create or admin authority, or named an organization that is not the current tenant"
+            "description": "Caller lacks the required role in the current tenant, or named an organization that is not it"
           },
           "404": {
             "content": {
@@ -57080,7 +57080,7 @@
                 }
               }
             },
-            "description": "Caller lacks the issue create or admin authority, or has no employee record in the current tenant, so the record would name nobody"
+            "description": "Caller lacks the required role in the current tenant, or has no employee record in it, so the record would name nobody"
           },
           "404": {
             "content": {
@@ -57214,7 +57214,7 @@
                 }
               }
             },
-            "description": "Caller lacks the issue-comment read or admin authority"
+            "description": "Caller lacks the required role in the current tenant"
           },
           "404": {
             "content": {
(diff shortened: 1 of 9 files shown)


abhijithanandan:
@coderabbitai review

PR #740, opened 2026-09-07
Say which finance response fields can be null

abhijithanandan:
The response side of #645, fifth module: **finance**. Same shape as the vendor pass (#678) and the procurement, receipts and attendance pass (#725). What is left is specified at the bottom rather than implied.

## What this pass did

Twenty schemas, 218 properties, 85 nullable. Picked on the criterion that chose the last three: this is the largest untouched block and it is money, so a client writing `?? 0` on an unmarked field turns "no figure was agreed" into a figure someone will act on.

| schema | nullable | of |
|---|---|---|
| `ConstructionInvoiceDto` | 21 | 36 |
| `ConstructionPaymentDto` | 22 | 31 |
| `ConstructionInvoiceLineDto` | 5 | 16 |
| `InvoiceDto` | 3 | 16 |
| `InvoiceLineDto` | 0 | 10 |
| `PaymentDto` | 3 | 13 |
| `AllocationDto` | 0 | 4 |
| `JournalEntryDto` | 6 | 13 |
| `JournalEntryLineDto` | 1 | 8 |
| `AccountDto` | 2 | 7 |
| `AccountTreeDto` | 1 | 8 |
| `CustomerDto` | 7 | 11 |
| `AddressDto` | 7 | 7 |
| `CompanyBankAccountDto` | 2 | 11 |
| `CostCategoryDto` | 3 | 6 |
| `BudgetAllocationDto` | 0 | 5 |
| `ProjectCostControlDto` | 0 | 3 |
| `ProjectCostControlLineDto` | 1 | 7 |
| `PostingAccountMappingDto` | 0 | 5 |
| `FinanceSettingsDto` | 1 | 1 |

`ReviewedResponseSchemas` records both halves and `ResponseNullabilityRatchetTest` holds the table, the annotation and the published document to each other, so a property added to any of these fails the build until someone classifies it.

## The warning the last pass left did not fire here

The procurement pass found three entities declaring `@JoinColumn(name = "project_id", nullable = false)` over a nullable column, which `ddl-auto: validate` cannot catch. Finance has no such disagreement: every `@Column` and `@JoinColumn` in the module matches its changeset, checked rather than assumed. Nothing filed as a defect from this pass.

## Where an answer did not come from a column

The line the earlier passes drew holds: a computed value with no column is non-null, and a nullable column stays nullable even where the application writes it on every path.

- **`InvoiceDto.balanceDue`** is the third instance of the `VendorSummaryDto` / `PayableDto.amountDue` shape. `Invoice.balanceDue()` subtracts one `NOT NULL` column from another and has no column of its own, so nothing in it can produce a null.
- **`ProjectCostControlDto` and its lines** are computed end to end. Every figure passes through `MoneyUtils.normalize`, which turns a null into zero; the head name falls back to `(unknown)`, and the totals row is labelled. Only `costCategoryId` admits null, and only on that totals row, which is what its null means.
- **`PostingAccountMappingDto`** is non-null throughout because `PostingAccountResolver.resolveWithSource` either returns an account or throws. A role with no mapping falls back to the configured default code, and a default code that resolves to nothing is an error rather than a null account.
- **`CustomerDto.billingAddress`** is nullable for a reason that is neither a column nor a mapper. `Address` is an `@Embedded` value whose seven columns are all nullable, and Hibernate leaves the embedded object null when every one of them is, whatever the `= new Address()` initialiser on `Customer` says. That makes `AddressDto` the one schema here with nothing at all in its non-null half.

## Two things that look inconsistent and are not

**Construction invoice money is non-null where purchase order money was not.** `subtotal`, `taxAmount`, `discountAmount`, `totalAmount`, `paidAmount` and `balanceAmount` sit on `NOT NULL` columns, so the constraint carries the claim; `PurchaseOrderDto.totalAmount` had a nullable column behind the same computation.

**`JournalEntryDto.createdBy` is nullable where `createdAt` on the same schema is not.** The two auditing columns disagree in the changelog: `created_at` is `NOT NULL` and `created_by` is not.

## Document diff

Audited semantically rather than by line count: **73 properties changed, 0 added, 0 removed**, `paths` byte-identical, and nothing outside the thirteen schemas that gained an annotation. The other twelve nullable properties were already marked and already published as unions. Non-null properties keep the descriptions they had, so the diff is the real change rather than a rewrite.

## Remaining

- **About 1587 response-only properties across 134 schemas**, of which roughly 435 are Spring's own `Page*` wrappers and are not ours to annotate. The reviewed set is now 34 schemas and 409 properties. The largest untouched blocks are `leave` (129), `inspection` (119), the rest of `attendance` (86) and `asset` (72).
- **The bean-bound request properties**, about 1000, unchanged in priority from the #670 comment.
- **The `echno-core` response-side check** now has 34 schemas to check against, which is past the two or three modules the vendor comment said would make it worth filing.

Refs #645

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -15,9 +15,12 @@
             "type": "string"
           },
           "description": {
-            "description": "Optional description of the account.",
+            "description": "Optional description of the account. Null where none was given.",
             "example": "Petty cash and bank balances",
-            "type": "string"
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "id": {
             "description": "Unique account id.",
@@ -96,9 +99,12 @@
             "type": "string"
           },
           "description": {
-            "description": "Optional description of the account.",
+            "description": "Optional description of the account. Null where none was given.",
             "example": "Petty cash and bank balances",
-            "type": "string"
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "id": {
             "description": "Unique account id.",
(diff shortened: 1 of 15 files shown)


abhijithanandan:
@coderabbitai review

Issue #741, opened 2026-09-07
Leave and inspection defects found by the #645 nullability pass


abhijithanandan:
Found while classifying the leave and inspection response schemas for #645. None of these is a documentation gap: the contract now describes each field correctly, and what it describes is wrong behaviour. Filed apart from the classification PR so the ratchet diff stays reviewable, the same way #728 was split out of #725.

## Leave

**1. `LeaveApprovalDto.delegatedFromName` is structurally always null.** `LeaveRequestMapper` ignores it and nothing else sets it. The `delegatedFromName` local in `NotificationService` builds message text and never reaches the DTO. A delegated approval therefore shows the id of the delegating approver and no name.

**2. `LeaveTransactionDto.createdByName` is structurally always null.** Same shape, in `LeaveTransactionMapper`. `createdById` is populated, only the name is missing, so a manual balance adjustment cannot be attributed in the UI without a second lookup.

**3. `LeaveRequestDto.handoverToName` behaves two ways for one schema.** Only `LeaveRequestService.getRequest` resolves it; every list response and every write response leaves it null. The same client reading the same field gets a name from one endpoint and null from another. That lookup also calls `employeeRepository.findById` rather than the tenant scoped `findByIdAndOrganizationId` used everywhere else in the service, so it reads across the organization boundary.

**4. `LeavePolicyService.createPolicy` discards the entity's declared defaults.** It calls every setter unconditionally with the payload value, so an explicit null in the request overwrites the field initialiser and Hibernate then writes NULL, bypassing the column default as well. Eight nullable columns are affected. `multiLevelApprovalEnabled` is the one field with a null guard. The consequence is silent because `LeaveRequestValidator` is null safe: a null `allowHalfDay` makes `Boolean.TRUE.equals(...)` false, so half day requests are refused on a policy that was meant to allow them, and null `minDaysPerRequest` or `advanceNoticeDays` make the `!= null` guards skip those checks entirely. `duplicatePolicy` copies the nulls onward.

## Inspection

**5. `inspection_defects.status` defaults to a value the enum cannot read back.** `027-create-inspection-defects.xml` sets `DEFAULT 'open'`. `'open'` is the `@JsonValue` form; `@Enumerated(STRING)` persists and reads the constant name `OPEN`. Migration `054-05` uppercased the existing rows but left the DEFAULT clause alone, so any insert that omits the column writes a row that throws `IllegalArgumentException` when the parent inspection is loaded.

[...]

PR #742, opened 2026-09-07
Classify the leave and inspection response schemas for nullability

abhijithanandan:
Sixth module pass on #645, and by the reasoning at the bottom the last one worth taking on its own. Seventeen schemas, 248 published properties, 107 nullable. Leave and inspection were the two largest untouched blocks and both shipped work this week.

| schema | nullable | of |
|---|---|---|
| `LeaveApprovalDto` | 5 | 12 |
| `LeaveBalanceDto` | 3 | 14 |
| `LeaveBalanceSummaryDto` | 0 | 7 |
| `LeaveCalendarDto` | 1 | 10 |
| `LeavePolicyDto` | 14 | 24 |
| `LeavePolicySimpleDto` | 2 | 6 |
| `LeaveRequestDto` | 13 | 27 |
| `LeaveTransactionDto` | 7 | 18 |
| `NotificationDto` | 4 | 11 |
| `InspectionDto` | 21 | 37 |
| `InspectionCheckItemDto` | 9 | 15 |
| `InspectionDefectDto` | 7 | 11 |
| `NcrDto` | 12 | 21 |
| `DefectPhotoAnnotationDto` | 2 | 11 |
| `ChecklistTemplateDto` | 1 | 9 |
| `ChecklistTemplateItemDto` | 5 | 10 |
| `StarterChecklistTemplateDto` | 1 | 5 |

Leave is 49 of 129, inspection 58 of 119. Request bodies and the `Page*` wrappers are out of scope, as before. `LeavePolicySimpleDto` appears in no path and is in scope anyway: it is nested inside `LeaveBalanceDto` and `LeaveRequestDto`.

### The entity-versus-changelog warning

Leave has no disagreement at all. Its tables are created across `v1.2` and `v1.3` and consolidated in the `v4.0` baseline, and both paths were read rather than one, because a column that is nullable on one install and not on the other would make the answer depend on how the database was built. They agree, and every `@Column` and `@JoinColumn` matches its changeset.

Inspection has three, all in the safe direction (the entity permits null where the column does not) and all on the tenant association: `Ncr`, `ChecklistTemplate` and `DefectPhotoAnnotation` each use a bare `@JoinColumn` over a `NOT NULL` `organization_id`. Nothing breaks today. Recorded on #741 because `validate` catches neither direction.

Neither module contains a Lombok `@Builder`, an `@Embedded` or an `@Embeddable`, so the attendance trap and the `CustomerDto.billingAddress` one have no analogue here. Leave does have a setter-shaped equivalent, on #741.

### Where an answer did not come from a column

- **`035-relax-inspection-not-null-for-ai-rows`** is the finding of this pass. `scheduledDate` and `inspectorId` are nullable even though `CreateInspectionRequest` and `UpdateInspectionRequest` both declare them `@NotNull`: the migration dropped the constraint so `ComplianceGenerationService` could write compliance rows without either. The constraint lives on the payload rather than on the row, so no API caller can omit them and rows exist that do. An entity read on its own would have got this backwards in both directions.
- **The five compliance fields** (`compliancePhase`, `riskLevel`, `resolutionOptions`, `complianceRuleRef`, `aiRationale`) are written only by that same service, so they are null on every manually created inspection. Two of them fall back to the rule's own default, which can itself be null.
- **`LeaveBalanceDto.available` and `bookable`** are the `VendorSummaryDto` shape a fourth time: `@Transient` getters over `LeaveDays.round` arithmetic on `NOT NULL` columns, with no column of their own.
- **`InspectionCheckItemDto.deviation`** is a computation that returns null. `MeasurementDeviation.of` yields nothing unless the measurement and the expected value both parse as numbers in the same unit.
- **`currentEmployeeId()`** returns null when the signed-in user has no employee row in the current organization, which is what `raisedById`, `verifiedById`, `closedById` and `createdById` mean by null. This is the `PostingAccountMappingDto` resolver shape with the opposite ending: it returns null where the posting resolver throws.
- **`InspectionDefectDto.severity`** is nullable because of a migration rather than a design. `054-03` cleared blank values to null rather than guessing a bucket.
- **Three names are null because nothing fills them in**: `delegatedFromName` and `createdByName` are structurally always null, and `handoverToName` is resolved on the single-request read alone. Classified as nullable because that is what the server sends, and filed on #741 because it is not what the code means.
- **The standing line held again.** `carryForwardFromPrevious`, `LeaveTransactionDto.description`, `priority` on both check-item schemas and `InspectionDefectDto.status` all sit on nullable columns the application writes on every path, so they stay nullable with the writing named in the description.

There is no `COALESCE` and no normalizer anywhere in either module.

### Document diff

Patched property by property rather than regenerated, then audited semantically: **107 properties changed, 0 added, 0 removed**, `paths` byte-identical, sixteen schemas touched and nothing outside them (`LeaveBalanceSummaryDto` has no nullable property, so it does not appear in the diff). Every changed property is one of the 107. Non-null properties keep the descriptions they had.

A checker held all three representations together before the commit: the `@Schema` description on each field, the manifest, and the description and type union in `docs/openapi.json`. All 107 agree, and no property listed as non-null admits null in the document. `openapi-served` from CI is the check that means anything on a hand patch, per #723.

### Remaining after this

About **1339 response-only properties over 117 schemas**, of which roughly 435 are Spring's own `Page*` wrappers. The reviewed set is now **51 schemas and 657 properties**. Largest untouched: the rest of `attendance` (86), `asset` (72).

Refs #645. Defects on #741.

## Summary by CodeRabbit

- **Documentation**
  - Updated API documentation across inspection, checklist, defect, non-conformance, leave, approval, and notification endpoints to accurately identify nullable fields.
  - Added descriptions explaining when values may be absent, defaulted, cleared, or populated by specific workflows.
  - Improved Swagger/OpenAPI field guidance without changing endpoint behavior or data structures.

- **Tests**
  - Expanded schema review coverage for nullable fields across inspection and leave-related responses.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -9122,7 +9122,6 @@
           },
           "delegatedFromName": {
             "description": "Name of the approver this action was delegated from. Always null on this response: the mapper leaves the field unset and nothing else fills it in, so resolve the name from delegatedFromId.",
-            "example": "Suresh Pillai",
             "type": [
               "string",
               "null"
@@ -10362,7 +10361,6 @@
           },
           "createdByName": {
             "description": "Name of the employee who created the transaction. Always null on this response: the mapper leaves the field unset and nothing else fills it in, so resolve the name from createdById.",
-            "example": "Anand Rajan",
             "type": [
               "string",
               "null"
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -2575,7 +2575,11 @@
             "type": "string"
           },
           "description": {
-            "type": "string"
+            "description": "What the template covers. Null where none was recorded, including on a template adopted from a starter checklist that carried none.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "id": {
             "format": "uuid",
@@ -2626,7 +2630,11 @@
         "description": "A check point within a checklist template, as returned by the API.",
         "properties": {
           "acceptanceCriterion": {
-            "type": "string"
+            "description": "What makes the check point pass. Null where none was recorded.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "category": {
             "type": "string"
@@ -2635,7 +2643,11 @@
             "type": "string"
           },
           "expectedValue": {
-            "type": "string"
+            "description": "Target value the check point is measured against. Null on qualitative check points.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "id": {
             "format": "uuid",
@@ -2649,13 +2661,25 @@
             "type": "boolean"
           },
           "priority": {
-            "type": "string"
+            "description": "Priority of the check point. The service substitutes \"medium\" wherever a payload or a starter checklist omits it, but the column permits null, so a row loaded outside the application can carry none.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "specification": {
-            "type": "string"
+            "description": "Reference specification for the check point. Null where none was recorded.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "tolerance": {
-            "type": "string"
+            "description": "Permitted band around the expected value. Null where none was recorded.",
+            "type": [
+              "string",
+              "null"
+            ]
           }
         },
         "type": "object"
@@ -4979,8 +5003,12 @@
         "description": "A mark drawn over one defect photo. Coordinates are fractions of the image width and height, so the mark holds its place at any rendered size.",
         "properties": {
           "createdById": {
+            "description": "Employee who drew the mark, resolved from the signed-in user's employee record in the current organization. Null where that user has no employee record.",
             "format": "int64",
-            "type": "integer"
+            "type": [
+              "integer",
+              "null"
+            ]
           },
           "id": {
             "format": "uuid",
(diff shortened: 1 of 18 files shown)


abhijithanandan:
@coderabbitai review

abhijithanandan:
Two of the three points taken, the other two declined for the same reason.

**Examples on `delegatedFromName` and `createdByName`: fixed** in a3fdcf3a. The description says the field is always null on this response and an example of a name next to that sentence contradicts it. Both examples are gone; neither field has a value to illustrate until #741 is fixed.

**`contactDuringLeave` example is a number: pre-existing, and not fixable from the annotation.** The DTO already carries `example = "9847012345"` as a quoted string, on `development` as well as here, and springdoc still emits it as a JSON number because it parses the example and infers the type. The property was rewritten by this patch only because it gained `"null"`; the example is byte-identical to what `development` publishes. Changing it needs the generator configured, not this file edited, since `docs/openapi.json` is CI-verified against the served document and a hand edit that springdoc will not reproduce fails the build.

**`null` missing from nullable enums: correct reading of the spec, wrong file to fix it in.** Same reason. Every one of these is springdoc output: `PayableDto.contractType` on `development` today is `"enum": [...]` with `"type": ["string", "null"]` and no `null` in the enum, from the finance pass that merged before this one. The document is generated and snapshot-checked, so adding `null` by hand would diverge from the served document and fail `OpenApiSnapshotTest` on the next regeneration. It is a real gap and it belongs in a separate issue against the generator configuration, applied to every nullable enum in the document at once rather than to the seven this PR happens to touch.

PR #743, opened 2026-09-07
TEMPORARY: measure the #741 reds

abhijithanandan:
Do not merge or review. Opened only to run the #741 tests against unrepaired `development` so the failures can be measured rather than asserted. Closed as soon as the run reports.


abhijithanandan:
Reds measured: all seven fail on unrepaired development, exactly the set expected. Recorded on #744; closing.

PR #744, opened 2026-09-07
Repair the leave and inspection nullability defects

### LINKED ISSUE #741: Leave and inspection defects found by the #645 nullability pass

abhijithanandan:
Closes #741.

Eight items, six repaired and two left as contract decisions. Every claim below about live data was checked against the compose-monolith staging on `.116` with read-only SQL; nothing was written there.

## The two that produce rows the application cannot handle

**`inspection_defects.status` DEFAULT `'open'`.** `@Enumerated(STRING)` persists and reads the constant name, so any insert omitting the column wrote a row that throws `IllegalArgumentException` when its parent inspection loads. `054-05` uppercased the rows that existed and left the DEFAULT alone, so the clause has kept writing unreadable rows since.

*Bad rows on staging: none.* All 2 defect rows hold `OPEN`. The default is still wrong, so this is a latent break rather than a live one, and `054-05` is why: the only rows that could have been written through the default predate it. `094-01` re-runs the promotion anyway, because a row written between `054` and this change would have taken the broken default.

**`inspections.organization_id` nullable.** `orgFilter` cannot satisfy its predicate against a null row and `TenantIsolationLoadListener` denies one outright, so such an inspection is unreachable and undeletable through the tenant-scoped API while its NCRs and annotations still carry a real org id.

*Bad rows on staging: none.* 0 of 15 inspections have a null organization.

`094` sequences both: normalise, then correct the default, then constrain; and backfill `organization_id` from the inspection's own project, then constrain. Each tightening carries a `sqlCheck` precondition that HALTs rather than proceeding, matching how `054-04` and `054-05` guard theirs. Nothing is deleted: an inspection with no recoverable organization can only be placed by someone who knows which tenant it belonged to, and removing it would destroy the evidence needed to answer that.

## The behaviour bug

**`LeavePolicyService.createPolicy` overwrote entity defaults with payload nulls.** Nine fields: `minDaysPerRequest`, `advanceNoticeDays`, `requiresAttachment`, `applicableGenders`, `minServiceMonths`, `allowHalfDay`, `isPaid`, `displayOrder`, and `multiLevelApprovalEnabled` (which already had the guard, spelled as an `if`). The column default never applied either, because a default fills an absent column and the insert named it.

The trigger is narrower than it first reads: `LeavePolicyCreationDto` carries the same initialisers, so a body that omits the field never reaches this. It takes an explicit JSON null, which is what a client sends when it serialises a form with the field cleared.

*Policies affected on staging: none.* All 4 rows are clean in all eight columns.

`duplicatePolicy` takes the same guard, so a policy written before this cannot copy its nulls into another organization.

## The isolation half of the naming defects

`handoverToName` was resolved with `employeeRepository.findById`, the one access `orgFilter` never narrows, while every other employee lookup in the service is scoped. It is **not** a live leak: `TenantIsolationLoadListener` denies the foreign row at the load boundary. The cost of leaning on that is that its denial is an exception, so the whole request-detail read fails where one optional name could not be resolved. Now scoped, and the read carries on with the name unresolved.

## Left alone, deliberately

The three names that are structurally always null (`delegatedFromName`, `createdByName`, `handoverToName`) stay unpopulated. Something else does carry the data: `delegatedFromId`, `createdById` and `handoverToId` are all published beside them, and each name's own `@Schema` description already directs the client to resolve from the id. Populating them means enriching fourteen read paths plus a nested list (`LeaveRequestDto.approvals` carries `LeaveApprovalDto`), which is a feature and an N+1 design decision; removing them from a hand-maintained contract breaks web at runtime rather than at compile time. Both are decisions for the contract owners, so `ReviewedResponseSchemas` records the outcome instead.

The five entities whose `@JoinColumn` permitted a null the column refuses now agree with the schema. Nothing depended on the disagreement and the build catches neither direction of it.

`035-relax-inspection-not-null-for-ai-rows` is untouched. Its constraints sit on the request DTOs, not the rows.

## Tests

Pushed as their own commit ahead of the repair, and run against unrepaired `development` in #743 to measure the failures rather than assert them.

- `InspectionNullabilityRepairIT` inserts a defect naming every column except `status`, so the database answers rather than the field initialiser. Asserts the stored value, that the row then loads through the entity, and that both columns are `NOT NULL` in `information_schema`.
- `LeavePolicyDeclaredDefaultsTest` asserts the entity handed to the repository, for a payload of explicit nulls and for a duplicate of a policy that already holds them. A third case sends `false` and `0`, so a repair that fell back to the default on any falsy value would be caught.
- `LeaveHandoverNameIsReadInTenantTest` asserts the scoped lookup answers, and that the unscoped one is not consulted even when it would answer.

`InspectionTaxonomyMigrationIT` now seeds an organization, since its inspections were written without one.

The OpenAPI document is unchanged: no DTO, controller or `@Schema` was touched.

### DIFF
--- a/src/main/java/org/tornotron/echno_backend/inspection/domain/ChecklistTemplate.java
+++ b/src/main/java/org/tornotron/echno_backend/inspection/domain/ChecklistTemplate.java
@@ -70,7 +70,7 @@ public class ChecklistTemplate implements TenantScopedEntity {
     private int version = 1;
 
     @ManyToOne(fetch = FetchType.LAZY)
-    @JoinColumn(name = "organization_id")
+    @JoinColumn(name = "organization_id", nullable = false)
     private Organization organization;
 
     @OneToMany(mappedBy = "template", cascade = CascadeType.ALL, orphanRemoval = true,
--- a/src/main/java/org/tornotron/echno_backend/inspection/domain/DefectPhotoAnnotation.java
+++ b/src/main/java/org/tornotron/echno_backend/inspection/domain/DefectPhotoAnnotation.java
@@ -128,7 +128,7 @@ public class DefectPhotoAnnotation implements TenantScopedEntity {
     private Long createdById;
 
     @ManyToOne(fetch = FetchType.LAZY)
-    @JoinColumn(name = "organization_id")
+    @JoinColumn(name = "organization_id", nullable = false)
     private Organization organization;
 
     @CreationTimestamp
--- a/src/main/java/org/tornotron/echno_backend/inspection/domain/Inspection.java
+++ b/src/main/java/org/tornotron/echno_backend/inspection/domain/Inspection.java
@@ -186,7 +186,7 @@ public class Inspection implements TenantScopedEntity {
     private String aiRationale;
 
     @ManyToOne(fetch = FetchType.LAZY)
-    @JoinColumn(name = "organization_id")
+    @JoinColumn(name = "organization_id", nullable = false)
     private Organization organization;
 
     @OneToMany(mappedBy = "inspection", cascade = CascadeType.ALL, orphanRemoval = true,
--- a/src/main/java/org/tornotron/echno_backend/inspection/domain/InspectionDefect.java
+++ b/src/main/java/org/tornotron/echno_backend/inspection/domain/InspectionDefect.java
@@ -64,7 +64,7 @@ public class InspectionDefect {
     private LocalDate targetDate;
 
     @Enumerated(EnumType.STRING)
-    @Column(name = "status", length = 20)
+    @Column(name = "status", length = 20, nullable = false)
     private DefectStatus status = DefectStatus.OPEN;
 
     @Column(name = "resolved_date")
--- a/src/main/java/org/tornotron/echno_backend/inspection/domain/Ncr.java
+++ b/src/main/java/org/tornotron/echno_backend/inspection/domain/Ncr.java
@@ -133,7 +133,7 @@ public class Ncr implements TenantScopedEntity {
     private LocalDateTime closedAt;
 
     @ManyToOne(fetch = FetchType.LAZY)
-    @JoinColumn(name = "organization_id")
+    @JoinColumn(name = "organization_id", nullable = false)
     private Organization organization;
 
     @CreationTimestamp
--- a/src/main/java/org/tornotron/echno_backend/leave/LeavePolicyService.java
+++ b/src/main/java/org/tornotron/echno_backend/leave/LeavePolicyService.java
@@ -23,6 +23,7 @@ import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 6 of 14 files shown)


abhijithanandan:
@coderabbitai review

